PR #8, opened 2017-01-04
Add README

MariusBluem:
I have done it without this "Why is it so awesome"-stuff, because - we should talk about `richdocuments`-only - and less possible about Collabora Online. Otherwise - the users will also misuse the issue tracker 😜 

@LukasReschke


MariusBluem:
cc @jospoortvliet

MariusBluem:
Anybody? 🏜🐪 cc @schiessle @jancborchardt @nickvergessen

PR #32, opened 2017-03-22
Try opening readonly documents too

pranavk:
Also simplify the logic during file action registration. Basically, we
should try to open all known mimetypes; file permissions are taken care
of later in the WOPI protocol.

Without this commit, it is not possible to open a document in a
directory for which user has no write access.


pranavk:
cc @nickvergessen @LukasReschke @icewind1991

Issue #51, opened 2017-05-12
Support opening encrypted files


pranavk:
Currently the app is not able to open any files when the server-side encryption in enabled. The reason is that the API call made by Collabora Online server doesn't have a real user context => no private key available => no decryption of documents.

pranavk:
cc @LukasReschke @schiessle

mmeeks:
Interesting; so - given that the users are pleading for this feature =) what information can we transfer from the user-context outside the iframe, to add to the WOPI credentials we pass to make this fly ? I guess this is further complicated by multi-factor authentication (?) anyhow - worth talking through. Ultimately, if we have the right WOPI token, and the user is still logged into OC the server has all the information we need - if only it would share it with us (?)

LukasReschke:
@schiessle is on holidays right now but I already talked to him about this and I can share some more insights on this.

So basically the encryption in Nextcloud works the way that every user has a private and a public key. The private key is stored encrypted by a derivative of the users' password and only stored unencrypted in memory when the user accesses Nextcloud logged-in. So we cannot access the users' file and modify it.

To work with publicly shared files we basically have a system public and private key which is accessible to the system obviously. What we should do here is have a possibility to generate such a key on-demand for single files and destroy it once used. So the workflow would be like:

1. User presses "open file in Collabora"
2. `DocumentController::index` gets called and would create a public sharing key for the file
3. When Collabora accesses the file via WOPI the public sharing key is used
4. When Collabora is closed the public sharing key is deleted

@schiessle Did I recap that correctly? If you would find some time to look into item 2 and 4 basically by just showing us a snippet creating and deleting the keys we could probably go a long way here. :)

pranavk:
Spent some time reading code around it. There doesn't seem to be any provision of adding extra system public/private key pair that can be used in this case as you described above. The default encryption module and internal encryption library handles publicShareKey, recovery keys, master keys and user keys. It might be possible to create a random $uid everytime and then add it in KeyManager so it is used as a user key instead of a system key (?)

Thoughts ?

ColdSphinX:
The nextcloud-drawio app seems to get the decrypted document as payload and sends that to the draw.io backend. Which is questionable if that backend is hosted by a third party...but I actually can't find a solution without temporary decrypting the document at some point. (More reasons to host everything yourself.)

https://github.com/pawelrojek/nextcloud-drawio/blob/master/drawio/js/editor.js#L43

PR #52, opened 2017-05-17
Support opening encrypted files

pranavk:
For #51

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -116,6 +116,13 @@ class DocumentController extends Controller {
 				'instanceId' => $this->settings->getSystemValue('instanceid'),
 			];
 
+			// Update the current file to be accessible with system public shared key
+			$owner = $item->getOwner()->getUID();
+			$absPath = '/' . $owner . '/' .  $item->getInternalPath();
+			$accessList = \OC::$server->getEncryptionFilesHelper()->getAccessList($absPath);
+			$accessList['public'] = true;
+			\OC::$server->getEncryptionManager()->getEncryptionModule()->update($absPath, $owner, $accessList);
+
 			$response = new TemplateResponse('richdocuments', 'documents', $params, 'empty');
 			$policy = new ContentSecurityPolicy();
 			$policy->addAllowedFrameDomain($this->appConfig->getAppValue('wopi_url'));
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -120,7 +120,7 @@ class WopiController extends Controller {
 			/** @var File $file */
 			$userFolder = $this->rootFolder->getUserFolder($res['owner']);
 			$file = $userFolder->getById($fileId)[0];
-
+			\OC_User::setIncognitoMode(true);
 			if ($version !== '0')
 			{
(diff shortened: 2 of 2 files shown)


pranavk:
First cut in making this work. I used system public shared key because that seemed to be the only possible way to make it work without changing/touching the server part of encryption module. Let me know your views.

I am adding the public shared key to encrypt the file key in the index() method and then setting the incognito mode in Wopi::GetFile which does the file decryption for us down the stack in the deep.

lib/Controller/DocumentController.php
```diff
@@ -116,6 +116,13 @@ public function index($fileId) {
 				'instanceId' => $this->settings->getSystemValue('instanceid'),
 			];
 
+			// Update the current file to be accessible with system public shared key
+			$owner = $item->getOwner()->getUID();
+			$absPath = '/' . $owner . '/' .  $item->getInternalPath();
+			$accessList = \OC::$server->getEncryptionFilesHelper()->getAccessList($absPath);
+			$accessList['public'] = true;
+			\OC::$server->getEncryptionManager()->getEncryptionModule()->update($absPath, $owner, $accessList);
```
schiessle: In general your approach works. I see only one problem. With `\OC::$server->getEncryptionManager()->getEncryptionModule()` you always get the default encryption module, so in a scenario where multiple encryption modules exists it could happen that you use the wrong module.

I confess that this is a theoretical problem because as of now only one encryption module exists. But it could become a problem in the future. Unfortunately I don't have a easy suggestion how to do it different with the current API.

timar: When you have a second encryption module, you should change the API, it is not possible to solve this problem at the app level. We need to keep this in mind. But for now, I think this patch is good to be merged. Thanks for your input.

schiessle: Well, I'm not sure that it is that easy. Because this way you work around the default process how to encrypt/decrypt the file.

Think about following scenario:

The document is already shares with multiple people. Now you want to work with someone else on the document using colabora online and execute the code path in this patch. You start editing and while you work on it someone else with access to the file decides to re-share it. This will trigger a re-calculation of the access list and since the file isn't shared as public link it will remove the additional key again and your edit session will break because access was removed.

I think we need something more integrated in the normal workflow. Either create a real public share for the edit session and remove the share afterwards again or, probably the better solution, implement a small share provider as part of the documents app which introduces it's additional (internal only) share type.

pranavk: While the document is being edited, the file has already been downloaded by Collabora Online. Any further attempts to open the same file in the nextcloud by any user will not ensue a download request again to the nextcloud server as long as the document is opened in Collabora Online; other users just join that same document. So, even if the access list is recalculated for any reason depriving public access to the file, editing session should keep working.

The window between when this patch of code is executed and when Collabora Online makes a request to download the file is ephemeral. If public access is stripped off in this short window, yeah, we would be unable to get the encrypted file but chances of this happening are less.

Having said that, I am convinced that implementing a small private share provider as part of documents app might be a better approach but for the time being, I also don't see any major repercussions with the current approach.

Do you know if there's already some app that implements a private share provider ?

schiessle: @pranavk sounds reasonable

PR #62, opened 2017-05-19
allowfullscreen

### LINKED ISSUE #13: View -- Full screen does not work.
### Steps to reproduce
1. Select View -- Full screen from Collabora Online's menu
2.
3.

### Expected behaviour
iframe goes full screen. It's convenient on mobile devices. 

### Actual behaviour
Nothing happens.
Firefox says: Request for fullscreen was denied because at least one of the document’s containing elements is not an iframe or does not have an “allowfullscreen” attribute.  loleaflet.htm
It works with ownCloud, that's why I think that it's a bug in Nextcloud's richdocuments or in Nextcloud.

pranavk:
This fixes #13

### DIFF
--- a/js/viewer/viewer.js
+++ b/js/viewer/viewer.js
@@ -90,7 +90,7 @@ var odfViewer = {
 			FileList.setViewerMode(true);
 		}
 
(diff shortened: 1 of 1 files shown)


pranavk:
It also fixes the slideshow presentations, the bug for which I remember is filed somewhere but don't know where.

pranavk:
#59 also

Issue #76, opened 2017-05-25
1.12.30 No default encryption module defined, 1.12.28 works


JSoko:
1.12.30 only shows Errorpage in Nextcloud. Have to revert to 1.12.28

Error:
OC\Encryption\Exceptions\ModuleDoesNotExistsException: No default encryption module defined

    /var/www/nextcloud/lib/private/Encryption/Manager.php - line 201: OC\Encryption\Manager->getDefaultEncryptionModule()
    /var/www/nextcloud/apps/richdocuments/lib/Controller/DocumentController.php - line 124: OC\Encryption\Manager->getEncryptionModule()
    [internal function] OCA\Richdocuments\Controller\DocumentController->index('864564')
    /var/www/nextcloud/lib/private/AppFramework/Http/Dispatcher.php - line 160: call_user_func_array(Array, Array)
    /var/www/nextcloud/lib/private/AppFramework/Http/Dispatcher.php - line 90: OC\AppFramework\Http\Dispatcher->executeController(Object(OCA\Richdocuments\Controller\DocumentController), 'index')
    /var/www/nextcloud/lib/private/AppFramework/App.php - line 114: OC\AppFramework\Http\Dispatcher->dispatch(Object(OCA\Richdocuments\Controller\DocumentController), 'index')
    /var/www/nextcloud/lib/private/AppFramework/Routing/RouteActionHandler.php - line 47: OC\AppFramework\App main('OCA\\Richdocumen...', 'index', Object(OC\AppFramework\DependencyInjection\DIContainer), Array)
    [internal function] OC\AppFramework\Routing\RouteActionHandler->__invoke(Array)
    /var/www/nextcloud/lib/private/Route/Router.php - line 299: call_user_func(Object(OC\AppFramework\Routing\RouteActionHandler), Array)
    /var/www/nextcloud/lib/base.php - line 1000: OC\Route\Router->match('/apps/richdocum...')
    /var/www/nextcloud/index.php - line 40: OC handleRequest()
    {main}

### Steps to reproduce
1. update to 1.11.30
2. Error (see above)
3. revert to 1.11.28, it works

### Server configuration
**Operating system**: Linux nc-vm 4.8.0-46-generic #49~16.04.1-Ubuntu SMP Fri Mar 31 14:51:03 UTC 2017 x86_64

**Web server:** Apache/2.4.18 (Ubuntu) (apache2handler)

**Database:** mysql 5.7.18

[...]

PR #78, opened 2017-05-25
Guard encryption support

pranavk:


pranavk:
#76 
should fix the issue. Sorry for the missing check there.

Issue #103, opened 2017-06-13
OCP\Files\NotFoundException if Shared File is in a Subdir


wscheicher:
### Steps to reproduce
1. Share a directory by link, allow upload and editing
2. Put a office document in a subdir
3. Access that share in a browser without authentification using the link
4. Try to open that document

### Expected behaviour
Document opens and can be viewed and edited

### Actual behaviour
Internal Server Error

The Nextcloud Log contains OCP\Files\NotFoundException and the filenames and paths in the log suggest that the information about the subfolder got lost.

Note: If the toplevel dir of the share contains a file with identical name, this one is opened instead (despite url and everything suggests the file in the subfolder would be edited)

### Server configuration
**Operating system**: Linux puck 4.9.0-0.bpo.3-amd64 #1 SMP Debian 4.9.25-1~bpo8+1 (2017-05-19) x86_64

**Web server:** Apache/2.4.10 (Debian) (apache2handler)

**Database:** pgsql PostgreSQL 9.6.3 on x86_64-pc-linux-gnu, compiled by gcc (Debian 4.9.2-10) 4.9.2, 64-bit

**PHP version:** 5.6.30-0+deb8u1
Modules loaded: Core, date, ereg, libxml, openssl, pcre, zlib, bcmath, bz2, calendar, ctype, dba, dom, hash, fileinfo, filter, ftp, gettext, SPL, iconv, mbstring, session, posix, Reflection, standard, shmop, SimpleXML, soap, sockets, Phar, exif, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xml, xmlreader, xmlwriter, zip, apache2handler, PDO, apcu, curl, gd, imagick, intl, json, ldap, mcrypt, pdo_pgsql, pgsql, readline, mhash, apc, Zend OPcache

**Nextcloud version:** 12.0.0 - 12.0.0.29

**List of activated apps:**
<details>
	<summary>App list</summary>

```
Enabled:
 - activity: 2.5.2
 - audioplayer: 2.0.2
 - bookmarks: 0.10.0
 - bruteforcesettings: 1.0.2
 - calendar: 1.5.3
 - circles: 0.10.0
 - comments: 1.2.0
 - contacts: 1.5.3
 - dav: 1.3.0
 - deck: 0.1.4
 - federatedfilesharing: 1.2.0
 - federation: 1.2.0
 - files: 1.7.2
 - files_external: 1.3.0
 - files_markdown: 1.0.1
[...]

PR #109, opened 2017-06-28
Fix editing publicly shared documents

andyboeh:
This fixes editing of documents in publicly shared subfolders (#12 and #103).

### DIFF
--- a/js/viewer/viewer.js
+++ b/js/viewer/viewer.js
@@ -66,7 +66,7 @@ var odfViewer = {
 		var viewer;
 		if($('#isPublic').val() === '1') {
 			viewer = OC.generateUrl(
-				'apps/richdocuments/public?shareToken={shareToken}&fileName={fileName}&requesttoken={requesttoken}',
+				'apps/richdocuments/public?shareToken={shareToken}&fileName={fileName}&requesttoken={requesttoken}&dir={dir}',
 				{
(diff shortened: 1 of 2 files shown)


pranavk:
It confuses me why we are passing fileName at all in case of public pages. The fileId is available in case of publicPages too and we can easily get the node from the fileId like this -- $node->getById($fileId)[0].

andyboeh:
@pranavk It confused me as well, given the fact that the file ID can be retrieved in the JavaScript code easily as well. I wanted a quick solution, so I came up with this.

pranavk:
Unless @LukasReschke sees anything fishy there, I would use the fileId there instead of fileNames. I tried the fileId approach on local and it works fine.

pranavk:
Can you update the PR to use fileIDs instead of fileNames, please ? Let's keep it consistent.

andyboeh:
Sorry, currently I'm too busy with other work. Maybe I'll have the time to do that in a few weeks.

wscheicher:
For now i use this fix, looks good so far.
Thanks a lot andyboeh!

pranavk:
Follow-up 

https://github.com/nextcloud/richdocuments/commit/e9f9711421ed1eba9cdb3b787b1a49eaed7373d1

pranavk:
Thanks andyboeh for working on it  !

PR #121, opened 2017-09-21
@NoCSRFRequired added

kapilnadiyapara:
@NoCSRFRequired annotation added


pranavk:
Hi,

As discussed, this doesn't sound like a good idea from security pov. I've opened #124 for your use-case.

PR #130, opened 2017-10-11
Add avatar to UserExtraInfo

LukasReschke:
Will require master of Nextcloud (aka 13) and https://github.com/nextcloud/server/pull/6813 merged.

Signed-off-by: Lukas Reschke <lukas@statuscode.ch>

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -12,7 +12,7 @@
 	<category>office</category>
 	<category>integration</category>
 	<dependencies>
-		<nextcloud min-version="12" max-version="12" />
+		<nextcloud min-version="12" max-version="13" />
 	</dependencies>
 	<documentation>
 		<admin>https://nextcloud.com/collaboraonline/</admin>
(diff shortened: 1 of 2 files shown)


LukasReschke:
screen shot 2017-10-11 at 15 26 45
screen shot 2017-10-11 at 15 28 50

timar:
Thanks for this @LukasReschke ! Does that mean that we need to make a branch for nc12 before we merge this patch?

LukasReschke:
In addition, I'll work on the following change in the server to always return a proper avatar: https://github.com/nextcloud/server/issues/5750

LukasReschke:
> Thanks for this @LukasReschke ! Does that mean that we need to make a branch for nc12 before we merge this patch?

Sadly, this feature will only properly work with Nextcloud 13 due to https://github.com/nextcloud/server/issues/5750 and https://github.com/nextcloud/server/pull/6813 being required.

I can add a if block around it to include the user extra info only if we're on >= 13. That should work around the problems here.

Objections? :-)

timar:
`

> I can add a if block around it to include the user extra info only if we're on >= 13. That should work around the problems here.

That would be great, thanks.

rullzer:
https://github.com/nextcloud/server/pull/6813 is merged

LukasReschke:
@timar Done.

Issue #69, opened 2017-05-24
feature request: let external users without account set their display name in the user list


DanScharon:
When a file is shared by link and made editable, an external user currently connected is only listed as an empty string within the users list.

Please make it possible for such external users to set their own display name.

nithom:
We (@JonasBlaesche and me) started working on this. You can find the current state
[here](https://github.com/nextcloud/richdocuments/compare/master...JonasBlaesche:development#files_bucket) (until now, we only tested it with CODE 2.5 and Nextcloud 12).

If a document is shared via link and the recipient is also a logged in member
of this NC instance, nothing changes. If the recipient is not logged in or does
not have an account, a textbox is displayed.

When the user submits his/her name, a cookie is set and the name is stored
in an additional column in the WOPI token relation.

![nextcloud-user-nickname](https://user-images.githubusercontent.com/13665755/35358151-f4a51272-0155-11e8-9b18-968801aa2730.png)


As long as the cookie exists, the form is not shown again. This allows to
join multiple documents via share link with the correct guest name everywhere.

The guest user names get a prefix ("Guest: ") which is localized in the TokenManager.

Regarding localization: There already exists a `"Guest %s"` string for the guest ID, is
there a way to duplicate this to get the additional colon? Manually doing this
would get overwritten by Transifex, would it?

We'd appreciate some feedback on this. After further testing, we'll open a PR.

errotu:
I really like the idea! Thanks for your work!

Nevertheless, I'm not sure if it's the best idea to create a "gate" before accessing the document. I think I'd prefer a possibility to set the name while being in the editing mode of the document. Another option is using the same form but only display it when someone tries to edit the document. Therefore, a only-reader wouldn't necessarily need to enter a nickname.

JonasBlaesche:
We tried to implement it they way you prefer but failed on the collabora side.

We also have some code for the read only case but from our point this code is not very nice at the moment. Our current code makes a xmlHttp request to apps/richdocuments/public?... and parses the resulting HTML for the read/write permission.
Does anyone know if there is a nice way to get information about read/write permissions in the viewer.js.

pranavk:
@JonasBlaesche Maybe you can move your 'enter nickname' logic to `documents.js` where you can already know the permission of the file from the exported variable `richdocuments_permissions` ?

nithom:
[...]

PR #179, opened 2018-03-03
Allow guest user to set their own display names (Issue #69)

JonasBlaesche:
Let external users without an account choose their own display name.


JonasBlaesche:
We found that the read only user solution is missing and therefor closed the PR. Once we are done with the fixes we will open a PR.

PR #192, opened 2018-03-27
Move over to Entity/EntityMapper

rullzer:
* Use the AppFrameworks entities
* Use the query builder
* Pass Entities (with types) around
* Remove old classes

It works properly for me. But it would be good if you all could give it a good test.


lib/Db/WopiMapper.php
```diff
@@ -0,0 +1,124 @@
...
+		$qb->select('*')
+			->from('richdocuments_wopi')
+			->where(
+				$qb->expr()->eq('token', $qb->createNamedParameter($token))
+			);
+		$result = $qb->execute();
+		$row = $result->fetch();
+		$result->closeCursor();
+
+		$this->logger->debug('Loaded WOPI Token record: {row}.', [ 'row' => $row ]);
```
ChristophWurst: wouldn't hurt to add the `app` name to associate the log entry with this app.

rullzer: Done

PR #200, opened 2018-04-16
Let external users set their display name when editing

### LINKED ISSUE #69: feature request: let external users without account set their display name in the user list

nithom:
Fixes #69.


pranavk:
This looks good to me. Also gave a quick test to it and it works as expected. Thanks for working on it.

PR #212, opened 2018-07-06
No need for allow inline script since we use a nonce

rullzer:
Function call had been deprecated in NC10 already.
We add a nonce to the inline javascript already. So no need to weaken
the CSP by allowing possible other inline scripts.


rullzer:
This might need to be reverted since it could do :boom: on some exotic browsers...

PR #221, opened 2018-07-10
Move spaces&tab mix to tabs in JS

rullzer:


rullzer:
Only reformat so merging :)

PR #222, opened 2018-07-10
Start with endpoint for mobile clients

rullzer:
This will allow mobile client to edit using collabora as well.

### Overview
Via capabilities the client can see that the richdocuments app is installed.
Via a call to the collabora app it can see which mimetypes are supported.

This also introduces a new OCS endpoint that `<server>/ocs/v2.php/apps/richdocuments/api/v1/document` A post to this endpoint with `fileId=<fileId>` will do the required verification and preparations. An url is returned in the response. All the client has to do is open this url in a webview.

### Todo

- [x] Add capability
  * Shows mimetypes
- [x] Add OCS endpoint to request URL
- [x] Cleanup
- [ ] Improve JS
- [x] Add lifetime to token
- [ ] Add job to cleanup tokens
- [ ] Extract common code with the normal endpoint
- [ ] Check if we need to disable some functionality on mobile

### Future work

- [ ] Support user-key server-side encryption?
- [ ] When closing the collabora view the process consumes 100% cpu
  * I think we can ignore this when we can just relay the close event to the mobile clients. Should still fix it at some point but not that urgent.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -22,6 +22,7 @@
 namespace OCA\Richdocuments\AppInfo;
 
 use OC\AppFramework\Utility\TimeFactory;
+use OCA\Richdocuments\Capabilities;
 use OCA\Richdocuments\WOPI\DiscoveryManager;
 use OCP\AppFramework\App;
 use OCP\AppFramework\IAppContainer;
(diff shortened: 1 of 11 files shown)


lib/AppInfo/Application.php
```diff
@@ -32,5 +33,8 @@ class Application extends App  {
 
 	public function __construct (array $urlParams = array()) {
 		parent::__construct(self::APPNAME, $urlParams);
+
+		// Regsiter capabilities
```
tobiasKaminsky: Typo ;-)

PR #223, opened 2018-07-11
Add sharing button to the header and open files details sidebar

juliushaertl:
### Todo

- [x] .icon-shared-white missing

### DIFF
--- a/js/viewer/viewer.js
+++ b/js/viewer/viewer.js
@@ -90,17 +90,7 @@ var odfViewer = {
 		}
 
 		OC.addStyle('richdocuments', 'mobile');
-		var $button = $('<div class="richdocuments-sharing"><a class="icon-shared icon-white"></a></div>');
-		$('.header-right').prepend($button);
-		$button.on('click', function() {
-			if (!$('#app-sidebar').hasClass('disappear') || $('#app-content').hasClass('with-app-sidebar')) {
-				OC.Apps.hideAppSidebar();
-				return;
-			}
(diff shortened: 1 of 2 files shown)


juliushaertl:
@rullzer I've tested on 14 with a real collabora instance, on 13 without, but it seems to work fine there with an empty collabra frame.

js/viewer/viewer.js
```diff
@@ -90,7 +90,18 @@ var odfViewer = {
 		}
 
 		OC.addStyle('richdocuments', 'mobile');
-		var $iframe = $('<iframe id="richdocumentsframe" scrolling="no" allowfullscreen style="width:100%;height:100%;display:block;position:absolute;top:0;z-index:60;" src="'+viewer+'" />');
+		var $button = $('<div class="richdocuments-sharing"><a class="icon-shared icon-white"></a></div>');
```
rullzer: Ah maybe we need a check here to see the header is actually present etc?
Because in the mobile UI this won't be there for example.

juliushaertl: I'd guess the header would just get hidden then on mobile so it should still be fine. Or would there be a different route that renders a template without the header?

rullzer: It is a different route that does not load the header at all.

juliushaertl: Ok, I'll make it optional then, when a header is present.

juliushaertl:
@rullzer Fixed.

PR #226, opened 2018-07-16
Fix not showing mobile UI for Safari on iOS

kevin147147:


kevin147147:
Before

bf41aeba-ed6e-485a-91bf-f8573cac1a15

After
523157ad-8668-4c1b-b9e8-77f1591a86b4

PR #249, opened 2018-09-04
Enable remote image insert and sharing

rullzer:


rullzer:
Note for now it is just enabled. This will maybe change in the future depending on the platform.

PR #255, opened 2018-09-06
Start of template API

rullzer:

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -1,14 +1,14 @@
 # Templates
 
 Often users will use the same templates. This documents describes the flow and
-API's that will eventually come into place.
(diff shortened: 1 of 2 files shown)


docs/templates.md
```diff
@@ -0,0 +1,84 @@
...
+* presentation
+
+This returns a list of templates with
+
+* id
+* name
+* link to preview
+* extension
+
+### Template previews
```
tobiasKaminsky: I am not sure if we need this.
In my current idea I will retrieve the preview by the link above.

tobiasKaminsky:
In future, this should also have eTag mechanism (or similar), so that clients do not need to download entire template listing.

PR #261, opened 2018-09-14
Enable remote images & share

tobiasKaminsky:
Then there will also be a share button in File menu

@skjnldsv can you please merge & update our server?


kendy:
I don't think this is correct, such a thing should be in the in WopiController.php, not in DirectViewController.php; and actually it is already there:

commit ab977712fc181fd081942bccd6882ec4334a21e4
Author: Roeland Jago Douma <roeland@famdouma.nl>
Date:   Tue Sep 4 12:48:39 2018 +0200

    Enable remote image insert and sharing

PR #263, opened 2018-09-18
Template management

skjnldsv:
- [x] System templates ui
- [x] Ocs api preview
- [x] Ocs api list
- [x] Ocs api create from template
- [x] Controller
- [x] finish js to actually fire off the get the new url
- [x] implement default endpoint to create a template token
---
- [x] Use absolute preview url

---
- [x] Expose empty templates by default in template endpoint (not sure how exactly yet)
- [x] Make 13 compatible (no scalar typehints)


lib/Controller/OCSController.php
```diff
@@ -45,32 +49,52 @@ class OCSController extends \OCP\AppFramework\OCSController {
...
 	 * @NoAdminRequired
 	 *
+	 * Init an editing session
+	 *
 	 * @param int $fileId
+	 * @return DataResponse
+	 * @throws OCSNotFoundException|OCSBadRequestException
 	 */
-	public function create($fileId) {
+	public function create(int $fileId): DataResponse {
```
rullzer: yeah we can't do that yet if we support NC12/13 still

skjnldsv: :cry:

juliushaertl:
@rullzer Pushed the personal settings implementation.

rullzer:
@juliushaertl with https://github.com/nextcloud/richdocuments/pull/334 merged. Could you make it so that we fall back to the old behavior for when templates are not available?

appinfo/database.xml
```diff
@@ -124,6 +124,12 @@
 				<unsigned>true</unsigned>
 				<length>4</length>
 			</field>
+			<field>
+				<name>template_destination</name>
+				<type>integer</type>
+				<notnull>false</notnull>
+				<length>4</length>
+			</field>
```
skjnldsv: Do we still use database xml?
I thought we killed the feature in profit or migrations?

rullzer: Yes we do for now. As 13 is still supported

PR #270, opened 2018-10-08
documents-overlay does not cover toolbar

tzolnai:
Need to use 0 px margin here otherwise this document-overlay does not cover the toolbar / menubar.
This document-overlay is expected to be displayed during loading the document and it should hide the building toolbar / menubar.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -224,7 +224,7 @@
 
 #documents-overlay,  #documents-overlay-below{
     position: fixed;
-    top: 45px;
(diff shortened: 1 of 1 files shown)


tzolnai:
By now this documents-overlay is not shown because an issue of the URL. If you change onStartup() method in documents.js to call documentsMain.overlay.documentOverlay('show'); unconditionally you can check what is the difference with the margin and without it.

PR #276, opened 2018-10-15
Move to compiled handlebars

rullzer:
The CSP in NC15 will be stricter so compiling at runtime won't work
anymore.

Generate the precompiled tempaltes by installing handlebars and running:
handlebars -n OCA.RichDocuments.Templates js/templates/ -f js/templates.js


juliushaertl:
@rullzer Can you add a build target to the Makefile with the handlebars command, so we have it at hand when changing something?

rullzer:
I did... just did not add it to git it seems... done now :P

PR #282, opened 2018-10-17
Properly remove view entries from the array

### LINKED ISSUE #279: Multiple of the same editor shown
Have 2 users: userA and userB

1. userA shares a document with userB
2. userB opens the document
3. userA opens the document
4. userA closes the document
5. userA opens the document again

Now the avatar userA shows up 2 times for userB.

juliushaertl:
documentsMain.UI.views is an array, therefore we cannot use indexOf and directly access the objects by ViewId. 

Fixes #279

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -543,12 +543,17 @@ var documentsMain = {
 							documentsMain.$deferredVersionRestoreAck.resolve();
 						}
 					} else if (msgId === 'View_Added') {
-						documentsMain.UI.views[args.ViewId] = args;
+						documentsMain.UI.views.push(args);
 						documentsMain.UI.renderAvatars();
 					} else if (msgId === 'View_Removed') {
-						index = documentsMain.UI.views.indexOf(args.ViewId);
-						if (index > -1) {
-							documentsMain.UI.views.splice(index, 1);
+						var view = $.grep(documentsMain.UI.views, function(element, index) {
+							return element.ViewId === args.ViewId;
+						});
+						if (view.length === 1) {
+							index = documentsMain.UI.views.indexOf(view[0]);
+							if (index > -1) {
+								documentsMain.UI.views.splice(index, 1);
+							}
 						}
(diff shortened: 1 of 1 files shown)


gohrner:
Richdocuments 3.0.3 still shows duplicate avatars - is this issue really fixed?

image

rullzer:
> Richdocuments 3.0.3 still shows duplicate avatars - is this issue really fixed?
> 
> image
> 
> Those "I"s and "A"s really are the same users.

@juliushaertl could you have a look?

gohrner:
Unfortunately, I don't know yet how to exactly reproduce it - I just know it happens pretty often. (I'd nearly say, more often than not.)

I'll try to find out how exactly it happens and report back.

juliushaertl:
I probably got your report wrong. So A is a user and I is a user, right? Then this is fixed with #304 :wink:

gohrner:
Yes, both "A"s are one user and both "I"s are another user. As far as we were able to check, both also only had a single session open, but maybe there were some stale sessions left - how can this be checked?

Great if #304 fixes it, then I'll patiently wait for the next release. :)

gohrner:
> @gohrner Can you post some steps how you got to that result?

I just tried, this phenomenon also happens if the users really have only one (active) open session.

I opened a spreadsheet with one user account in one browser and then logged in as another user using another browser and parallel and also opened this document.

The second user saw the first user's avatar twice while the first user saw the second user's avatar only once.

I then closed the spreadsheet as the first user, and only one of the avatar's disappeared.

Interestingly, the "User"-session menu in the status bar was correct for the second user and only showed the first user once, who also disappeared after closing the spreadsheet.

juliushaertl:
@gohrner Did you try that with v3.0.5 released earlier today?

gohrner:
Thanks, seems to work - at least with 3.0.5 there's no immediate Avatar duplication. :)

PR #288, opened 2018-10-26
Parse the link to AssertURL trought he generateURL

rullzer:
This is to make sure that index.php etc is properly added. Else this
might fail on some setups.

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -527,7 +527,7 @@ var documentsMain = {
 								var filename = path.substring(path.lastIndexOf('/') + 1);
 								$.ajax({
 									type: 'POST',
-									url: OC.linkTo('richdocuments', 'assets'),
+									url: OC.generateUrl('apps/richdocuments/assets'),
 									data: {
 										path: path
 									}


js/documents.js
```diff
@@ -527,7 +527,7 @@ var documentsMain = {
 								var filename = path.substring(path.lastIndexOf('/') + 1);
 								$.ajax({
 									type: 'POST',
-									url: OC.linkTo('richdocuments', 'assets'),
+									url: OC.generateUrl('apps/richdocuments/assets'),
```
MorrisJobke: What if it is installed in `apps2`? 🤔

rullzer: Ah shit. see that is why we used the LinkTo...

but LinkTo doesn't take index.php into account

rullzer: Mmmm ok lets fix that in a different PR
https://github.com/nextcloud/richdocuments/blob/4a30187fc11655a5f3de15f91e20f1c90d8a90e6/js/documents.js#L138

We do this in many places

PR #299, opened 2018-11-07
Fix Travis' config

ChristophWurst:
* Remove duplicate `matrix` entry
* Move `sudo` to the top
* Separate passing and failing code checker steps
* Simplify build matrix
* Fix indentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
+sudo: true
 language: php
-
 php:
   - 7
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,5 @@
+sudo: true
 language: php
-
 php:
   - 7
```
rullzer: Ah we'd also need 5.6, 7.1 and 7.2 then :wink:

ChristophWurst: Done. Obviously the build matrix grew but the test scripts are fast, so no big deal IMO

rullzer:
- [ ] Seems to still miss a 5.6 linter
- [ ] Should not test against stable12 (not supported)
- [x] Should test against stable13
- [ ] Should test against stable14

ChristophWurst:
Fixed: https://travis-ci.org/nextcloud/richdocuments/builds/452262372

rullzer:
Nice thnx a lot!

PR #304, opened 2018-11-09
Avatars fixes

### LINKED ISSUE #300: Editors in top bar should show readonly status
It would be nice if we somehow indicate which users can edit and which can only read a document.

Currently all users are created equally in the editor list (in the header).

CC: @juliushaertl

### LINKED ISSUE #302: Clicking on avatar opens dropdown at wrong location
wrong_loc

I clicked on the R avatar. And as you can see the dropdown with 'no actions available' opens all the way to the left.

juliushaertl:
- Use object to store views instead of array
- Show user avatars just once (even if one user has multiple sessions open)
- Indicate read only users (fixes #300)
image
- Fix popover menu position (fixes #302)
image

### DIFF
--- a/css/viewer/odfviewer.css
+++ b/css/viewer/odfviewer.css
@@ -52,6 +52,9 @@
   margin-left: -20px;
   box-shadow: 0px 0px 3px rgba(0, 0, 0, 0.5);
 }
+.richdocuments-avatar.read-only {
+  opacity: 0.8;
+}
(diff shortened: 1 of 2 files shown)


rullzer:
@juliushaertl would it be possible to make sure that `(read only)` is a single line?

juliushaertl:
Fixed. 

image

PR #323, opened 2018-12-07
Fetch capabilities from collabora and store them in appdata

### LINKED ISSUE #318: Expose endpoint for clients
Only the upcoming COOL version 4.0 should be used with clients, as the current one does not support hdpi, no decent mobile/response design.

So RichDocuments should expose "direct_editing" via Capabilities, which checks against COOL version via their capability endpoint:
- no capability / version <4.0 => "direct_editing": false
- version >= 4.0 => "direct_editing": true

With this clients can right now add and check this.

@rullzer as discussed on irc
@juliushaertl maybe you can help with the implementation as Roeland is not available tomorrow
@marinofaggiana please implement the check also on iOS

juliushaertl:
- [x] Fetch collabora capabilities and store them in app data
- [x] Add collabora capabilities to the richdocument capabilities
- [ ] Wait for version capability to be available and make direct_editing depend on that

```
        "richdocuments": {
          "mimetypes": [
            "application/vnd.oasis.opendocument.text",
            "application/vnd.oasis.opendocument.spreadsheet",
            "application/vnd.oasis.opendocument.presentation",
            "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
            "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
            "application/vnd.openxmlformats-officedocument.presentationml.presentation"
          ],
          "collabora": {
            "convert-to": {
              "available": false
            }
          },
          "direct_editing": false
        },
```

Fixes #318

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -23,19 +23,48 @@
 
 namespace OCA\Richdocuments;
 
+use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\Capabilities\ICapability;
+use OCP\Files\IAppData;
+use OCP\Files\NotFoundException;
+use OCP\Files\SimpleFS\ISimpleFolder;
+use OCP\Http\Client\IClientService;
+use OCP\IConfig;
 use OCP\IURLGenerator;
 
 class Capabilities implements ICapability {
 
-	/** @var IURLGenerator */
-	private $urlGenerator;
+	/** @var IConfig */
+	private $config;
+	/** @var IClientService */
+	private $clientService;
+	/** @var ITimeFactory */
+	private $timeFactory;
+	/** @var ISimpleFolder */
+	private $appData;
 
-	public function __construct(IURLGenerator $urlGenerator) {
-		$this->urlGenerator = $urlGenerator;
+	/**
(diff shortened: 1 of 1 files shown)


lib/Capabilities.php
```diff
@@ -46,8 +75,44 @@ public function getCapabilities() {
 					'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
 					'application/vnd.openxmlformats-officedocument.presentationml.presentation',
 				],
+				'collabora' => $collaboraCapabilities,
+				'direct_editing' => false, //TODO: fix once proper capability is available
+				'templates' => false, //TODO: fix once proper capability is available
```
rullzer: @tobiasKaminsky @marinofaggiana please check this value before you show the 'create new file' - collabora thingy. As it is not available by default.

tobiasKaminsky: So you can have direct editing (cool webview) without templates?

rullzer: yes

tobiasKaminsky: @rullzer can you update our test server, so we can test it?

PR #339, opened 2018-12-31
Improve current editor listing in Nextcloud header

### LINKED ISSUE #320: Improve listing of current editors
As discussed:

image

- [ ] Add option to remove view
- [ ] Allow to follow editing users

juliushaertl:
- [x] Limit avatar list in header to 3
- [x] Show popover with a full list of editing users
- [x] Follow user activities by clicking the name
- [x] Allow to remove the user from collabora by clicking the x icon
- [x] Indicate current following user
- [x] Add option to follow the current editor
- [x] Add color indicator for editors
- [x] Update on changes by FollowUser_Changed post message

image

Fixes #320

### DIFF
--- a/css/viewer/odfviewer.scss
+++ b/css/viewer/odfviewer.scss
@@ -56,7 +56,7 @@
       border-left: 3px solid var(--color-primary, $color-primary);
     }
 
-    div.label {
+    .label {
       padding: 6px;
       flex-grow: 1;
 
@@ -66,6 +66,10 @@
         padding-left: 10px;
         vertical-align: top;
       }
+    }
+    label.label::before {
+      margin-right: 16px;
+      margin-left: 7px;
     }
     .icon-close {
       padding: 8px;
--- a/css/viewer/odfviewer.scss
+++ b/css/viewer/odfviewer.scss
@@ -44,7 +44,31 @@
   z-index:110;
 }
 
-.richdocuments-avatar {
+#editors-menu {
+  min-width: 200px;
+}
+
+#editors-menu ul li {
+  display: flex;
+  padding: 3px;
+
+  div.label {
+    padding: 6px;
+    flex-grow: 1;
+
+    .icon-play-next {
+      opacity: .5;
+      display: inline-block;
+      padding-left: 10px;
+      vertical-align: top;
+    }
+  }
+  .icon-close {
+    padding: 8px;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


jancborchardt:
> Option to "always follow the current editor" is missing

Just a checkbox on the bottom with short text "Follow editor"

> How do we indicate the current following user

Just put a line "Editing …" below their name? Or a pencil (icon-rename) overlaying the bottom right of their name. Or bolding their name. (First option is best.)

Also, we need the colored ring around the avatars since the colors differ from the placeholder colors.

- Ideally they should be the same as the placeholder colors.
- But we do need the ring anyway cause you need to see the color when people have an image set. So a circular 5px border around the image is good.

jancborchardt:
Also, very nice @juliushaertl! :+1:  :slightly_smiling_face:

juliushaertl:
Ready for review.

> Just put a line "Editing …" below their name? Or a pencil (icon-rename) overlaying the bottom right of their name. Or bolding their name. (First option is best.)

I actually use a 3px left border like we use in the app navigation for highlighting the current editor. (See screenshot in op)

> Ideally they should be the same as the placeholder colors.

Probably not that easy to achieve. Collabora would need to have an option to hand over a color or they need to take our mechanism to generate the color from the username.

> But we do need the ring anyway cause you need to see the color when people have an image set. So a circular 5px border around the image is good.

Done. However I used a 2px radius, which looks a bit nicer:

image
image

eszkadev:
> Done. However I used a 2px radius, which looks a bit nicer:
> 
> image
> image

Nice, it looks coherent with our mobile widget:
users

What about adding separator after users list and before "Follow current editor" entry?

jancborchardt:
> I actually use a 3px left border like we use in the app navigation for highlighting the current editor. (See screenshot in op)

:+1: looks good!


> However I used a 2px radius, which looks a bit nicer

Yes, that’s much better. @eszkadev note that on your screenshot the border looks bigger. It should be like in @juliushaertl’s top screenshot with the 2px radius. :)

> What about adding separator after users list and before "Follow current editor" entry?

No need for a separator, the whitespace is enough. @juliushaertl the checkbox seems to be missing? Also, the text of "Follow current editor" should be moved to the right a bit so it’s left-aligned with the usernames.

juliushaertl:
@jancborchardt Added the checkbox as well: 
image

PR #343, opened 2019-01-05
convert OC locale to correct BCP 47 Tags for Identifying Languages

leejoneshane:


leejoneshane:
sorry, still some bug!

PR #344, opened 2019-01-05
convert nextcloud locale to BCP47 languages tag correctly

leejoneshane:
tested out!


juliushaertl:
Thanks a lot for your pull request, can you describe a bit more in detail where there was an issue with the code?

juliushaertl:
cc @eszkadev Maybe that is something to be fixed on your side as well?

eszkadev:
Thanks for notification, modification indicator needed some tweak

rullzer:
Lets do this then

PR #363, opened 2019-01-24
remove language setting of Open Document Format templates

timar:


juliushaertl:
@timar Seems to partly do the trick, however the placeholder texts in the presentation template still show english placeholders. When I click them and click outside of the placeholder again, they switch to the users language.

rullzer:
> @timar Seems to partly do the trick, however the placeholder texts in the presentation template still show english placeholders. When I click them and click outside of the placeholder again, they switch to the users language.

Maybe we also just kill that ;)

juliushaertl:
Lets get this in, as we have proper spell checking language set then.

PR #368, opened 2019-01-25
Add new file dialog on UI_CreateFile

juliushaertl:
This PR implements creating a new document from withing collabora:

peek 2019-01-25 14-30

This is how it works:

- Click the button in the top left
- The user enters a filename
- We open the files app in a new tab with the filetype and filename as parameters
- viewer.js parses those parameters and asks the user to choose a template
- We create a document with the existing method from the template

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -555,6 +555,9 @@ var documentsMain = {
 							$buttons.eq(0).text(t('richdocuments', 'Cancel'));
 							$buttons.eq(1).text(t('richdocuments', 'Save'));
 						});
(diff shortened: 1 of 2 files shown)


skjnldsv:
UX is good from me! :+1: 
Is it deployed on the test server?

tobiasKaminsky:
> This PR implements creating a new document from withing collabora

This is disabled for mobile, or?

rullzer:
> > This PR implements creating a new document from withing collabora
> 
> This is disabled for mobile, or?

I think so yes

PR #369, opened 2019-01-25
Add file actions menu to the header bar

juliushaertl:
peek 2019-01-25 16-47


jancborchardt:
Looks good! Just 2 details:
- The top left and right corners should also have border-radius
- Centering the menu below the 3-dot menu instead of the left align would look better?

rullzer:
@juliushaertl can you tackle @jancborchardt comments in a followup?

PR #379, opened 2019-02-05
Move fetching the capabilities to a backgroundjob

rullzer:


juliushaertl:
@rullzer Should we maybe also trigger fetching when the admin sets the url of the collabora server? Otherwise you'll need to wait up to an hour until the capabilities are fetched the first time.

rullzer:
@juliushaertl makes sense. Feel free to take over as I do not have time today or tomorrow.

Probably good to move the fetching code to a service then. So we can call it when the url is set. indeed.

rullzer:
@juliushaertl please have another look

PR #384, opened 2019-02-06
Fix scrolling behaviour on Nextcloud >= 14

### LINKED ISSUE #217: Calc: In some time scroll the code app showing nextcloud files app.
### Steps to reproduce
1. Launch a calc document
2. Some random scroll

### Expected behaviour
Scrool document..

### Actual behaviour
In some time scroll the code app showing nextcloud.

imagen

### Server configuration

**Operating system**: Fedora 28 but others too..

**Web server:**: Fedora 28, Apache 

**Database:**: Mariadb/mysql

**PHP version:**: 7.1

**Nextcloud version:**: Now  13.0.4, but others too..

**Updated from an older Nextcloud/ownCloud or fresh install:**: Updated.

**Where did you install Nextcloud from:** Source..

juliushaertl:
Fixes #217

@rullzer This is actually the first one where I cannot think of an easy solution to keep up 13 support, since the scroll container changed in Nextcloud 14. Also the height of the header bar is different, so we cannot just calculate depending on the window height.

### DIFF
--- a/css/viewer/odfviewer.scss
+++ b/css/viewer/odfviewer.scss
@@ -37,7 +37,8 @@
 #richdocumentsframe {
   background-color: #fff;
   width:100%;
-  height:100%;
+  /** TODO: This will probably look odd on 13 */
+  height: calc(100vh - 50px);
   display:block;
   position:absolute;
   top:0;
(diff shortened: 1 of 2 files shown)


juliushaertl:
@rullzer Pushed a fix to apply some styling depending on the Nextcloud version, so it seems to work fine with 13-15 now.

PR #400, opened 2019-02-13
Move to single viewer css file

juliushaertl:
Saves one request

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -39,16 +39,14 @@ $eventDispatcher->addListener(
 	'OCA\Files::loadAdditionalScripts',
 	function() {
 		\OCP\Util::addScript('richdocuments', 'viewer/viewer');
-		\OCP\Util::addStyle('richdocuments', 'viewer/odfviewer');
-		\OCP\Util::addStyle('richdocuments', 'templatePicker');
+		\OCP\Util::addStyle('richdocuments', 'viewer');
 	}
 );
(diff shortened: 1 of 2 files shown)


thebearon:
Please note that this change caused a regression, when clicking the "+" in the file list to create a new file, the document/spreadsheet/presentation icons aren't shown anymore.

juliushaertl:
> Please note that this change caused a regression, when clicking the "+" in the file list to create a new file, the document/spreadsheet/presentation icons aren't shown anymore.

I cannot reproduce that, maybe it is an issue with your browser caching the old file, because we didn't do the app version bumb yet on master.

thebearon:
> I cannot reproduce that, maybe it is an issue with your browser caching the old file, because we didn't do the app version bumb yet on master.

I checked in more detail, the problem is that the CSS file was moved a level up, but the following URLs inside weren't updated:
- background-image: url('../../../../core/img/filetypes/x-office-document.svg')
- background-image: url('../../../../core/img/filetypes/x-office-spreadsheet.svg');
- background-image: url('../../../../core/img/filetypes/x-office-presentation.svg');

And yes, the bug still occurs with a freshly pulled master.

juliushaertl:
Ah I guess you have installed Nextcloud in a subdirectory, that explains why I was not seeing any issues there, as with Nextcloud in the webserver root, the path resolves just fine. Fix incoming.

thebearon:
Thank you!

PR #402, opened 2019-02-13
fix 'Collabora Online' product name

timar:


l10n/cs.js
```diff
@@ -71,7 +71,7 @@ OC.L10N.register(
     "template preview" : "náhled šablony",
     "Select a template directory" : "Vyberte složku se šablonou",
     "Remove personal template folder" : "Odebrat složku s osobní šablonou",
-    "Templates inside of this directory will be added to the template selector of Collabora online." : "Šablony v této složce budou přidány do výběru šablon v Collabora online.",
```
juliushaertl: Unfortunately changes in the translation files will get overwritten by the strings from transifex. Please remove those from the commit. The actual translations should get fixed by our translators once the main string update. has been pushed there.

timar: OK, I created a new PR without the l10n files.

PR #407, opened 2019-02-16
Fixes #381: use language instead of locale

### LINKED ISSUE #381: Collabora doesnt respect NextCloud language setting
**Describe the bug**
I have Language set to "English (US)". But Locale is set to "Swedish (Sweden)". I do this, because I want my UI in english, but dates and sorting to work like swedish. It works fine for everything in NextCloud, the UI is in english and the sorting and dates are according to swedish rules. But for Collabora, the UI becomes Swedish. Which I find very annoying :)

**To Reproduce**
1. Change Locale
2. Open a collabora document, language is set after locale not, Language setting.

**Expected behavior**
Collabora to use Language setting to control language

**Screenshots**

**Client details:**
 - OS: Linux
 - Browser Firefox
 - Version 64.0.2
 - Device: Laptop

## Server details

**Operating system**: Fedora 29 Server

**Web server:** Apache

**Database:** Maria DB

**PHP version:** 7.2.14

**Nextcloud version:** 15.0.2

**Version of the richdocuments app** 3.1.1

**Version of Collabora online** 6.0-21

totpet:
* Resolves: #381
* Target version: master 

### Summary

Use the [`getLanguage`](https://github.com/nextcloud/server/blob/1ceddc1cc2b9a3757118830c3f92d80932fd77ef/core/js/js.js#L763) method instead of `getLocale`. This is the expected behaviour and  consistent with the rest of NextCloud UI.

### TODO

- [ ] Testing

### Checklist

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -176,6 +176,10 @@ var documentsMain = {
 			return parent.OCA.Files.FileList;
 		}
 		return null;
+    },
+
+	getLanguage: function() {
+		return OC.getLanguage ? OC.getLanguage() : OC.getLocale();
 	},
 
 	UI : {
@@ -388,7 +392,7 @@ var documentsMain = {
 			var urlsrc = documentsMain.urlsrc +
 				"WOPISrc=" + wopisrc +
 				"&title=" + encodeURIComponent(title) +
-				"&lang=" + OC.getLocale().replace('_', '-') + // loleaflet expects a BCP47 language tag syntax
+				"&lang=" + documentsMain.getLanguage().replace('_', '-') + // loleaflet expects a BCP47 language tag syntax
 				"&permission=readonly";
 
 			// access_token - must be passed via a form post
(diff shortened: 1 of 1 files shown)


juliushaertl:
@thebearon Is there a way we can hand over the locale as well or is collabora currently limited to one language parameter?

thebearon:
> @thebearon Is there a way we can hand over the locale as well or is collabora currently limited to one language parameter?

Currently only a single language parameter is supported.

totpet:
This change is not compatible with NC 13 (which is EOL as of this month), because the `getLangugage` method was added in 14.0. If older version should be supported, we could use a wrapper like this:
`getLang = oc.getLanguage ? oc.getLanguage : oc.getLocale`

juliushaertl:
@totpet Yes, we try to keep 13 compatibility for now, so please add a fallback to getLocale :+1:

totpet:
> @totpet Yes, we try to keep 13 compatibility for now, so please add a fallback to getLocale +1

Sure, okay. Should I use a wrapper function (and where should I put it?), or simply use the ternary inline?

juliushaertl:
> Sure, okay. Should I use a wrapper function (and where should I put it?), or simply use the ternary inline?

Lets add a wrapper function to `documentsMain`

totpet:
Ok, seems to work on both 13.0.11 and 15.0.4

juliushaertl:
Thanks a lot @totpet

Issue #264, opened 2018-09-18
Self-signed certificate on Collabora does not result in proper error message


matthijskooijman:
I suspect this issue belongs in this repo. If not, I'll happily move it elsewhere.

TLDR; if you run Collabora with a self-signed certificate, you get a internal server error and a (seemingly unrelated) PHP fatal error. Refusing to run seems acceptable, but it would be good to show a better error message.

I realize that mixing HTTP and HTTPS and using self-signed certificates  is not recommended for production, but I was just intending to test. It is totally acceptable for this setup to not work, but it looks like there is a problem in the error handling, that might be worth fixing, so anyone else trying something like this will not be as confused as I was :-)

### Steps to reproduce
1. Install Next cloud & Collabora using their docker images
2. Modify the Collabora container to expose its port directly, rather than setting up a reverse proxy
3. Set up the richdocuments app in Nextcloud, pointing to the Collabora port directly (e.g. https://1.2.3.4:9980/)
4. Create and open an odt file.

### Expected behaviour
I would have hoped this setup to work. Failing that, I would have liked to see a proper error message directly.

### Actual behaviour
Opening up a odt file from nextcloud gives a mostly empty page. The url inside the iframe for the richtdocuments app (http://1.2.3.4:8080/apps/richdocuments/index?fileId=154&requesttoken=removed)  gives a 500 error. Looking in the docker logs, I see a fatal error:

```
[Tue Sep 18 10:05:04.189698 2018] [php7:error] [pid 416] [client 83.80.204.85:46282] PHP Fatal error:  Cannot declare class GuzzleHttp\\Handler\\CurlFactory, because the name is already in use in /var/www/html/3rdparty/guzzlehttp/guzzle/src/Handler/CurlFactory.php on line 16
83.80.204.85 - - [18/Sep/2018:10:05:04 +0000] "GET /apps/richdocuments/index?fileId=154&requesttoken=removed HTTP/1.1" 500 870 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:52.0) Gecko/20100101 Firefox/52.0"
```

Later, I noticed another error in the logging section of the Nextcloud section, preceding each fatal error: GuzzleHttp\Exception\RequestException: cURL error 60: SSL certificate problem: self signed certificate in certificate chain (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)

[...]

PR #414, opened 2019-02-24
Add option to disable certificate validation

kesselb:
* Resolves: #264
* Target version: master 

### Summary

image

Add option to disable certificate validation for internal nextcloud <-> collabora connections. Useful if you use self signed certificates or your collabora certificate is not trusted by your nextcloud server.

### TODO

- [ ] Test with Nextcloud 13 (don't have a running test setup)

### Checklist

### DIFF
--- a/js/admin.js
+++ b/js/admin.js
@@ -27,10 +27,11 @@ var documentsSettings = {
 	},
 
 	save : function() {
-		$('#wopi_apply').attr('disabled', true);
+		$('#wopi_apply, #disable_certificate_verification').attr('disabled', true);
 		var data = {
-			wopi_url  : $('#wopi_url').val().replace(/\/$/, '')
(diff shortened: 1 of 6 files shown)


juliushaertl:
Fine by me :+1:

rullzer:
I do not fully get the reasoning here... if your collabora has an invalid certificate then also your browser will complain...

kesselb:
> if your collabora has an invalid certificate then also your browser will complain...

Of course but you have to choice to ignore the warning. Before this pull request there is no way to accept a self signed certificate.

rullzer:
ok fine by me then

kevdogg:
This fix does not fix the guzzlehttp name already in use error. I wish people would quit referencing this as a fix for this problem

PR #422, opened 2019-02-28
Fix some issues reported by sentry

juliushaertl:
- Fix parent access (NEXTCLOUD-P9 NEXTCLOUD-N8 NEXTCLOUD-P1)
- Fix call to missing file model NEXTCLOUD-PK
- Use proper fileList object when fetching the file model 
- Cache json errors when parsing messages NEXTCLOUD-J9
- Fix sidebar event removal NEXTCLOUD-JX

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -923,16 +923,18 @@ var documentsMain = {
 	},
 
 	getFileModel: function() {
-		parent.OCA.Files.App.fileList._updateDetailsView(documentsMain.fileName, false);
-		var fileModel = parent.OCA.Files.App.fileList.getModelForFile(documentsMain.fileName);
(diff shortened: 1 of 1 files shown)


js/documents.js
```diff
@@ -607,7 +606,10 @@ var documentsMain = {
 			// Listen for App_LoadingStatus as soon as possible
 			$('#loleafletframe').ready(function() {
 				var editorInitListener = function(e) {
-					var msg = JSON.parse(e.data);
+					var msg = {};
+					try {
+						msg = JSON.parse(e.data);
+					} catch (e) {}
```
rullzer: Should we return here?

PR #424, opened 2019-03-01
Fetch proper file info for files that are out of view after reload

juliushaertl:
Reloading the file list will cause the file model to be missing, therefore we need to call scrollTo to load the full file list and have it accessible.


juliushaertl:
@rullzer Another one, I found when testing.

PR #446, opened 2019-03-11
Change timeout from 5 to 10 seconds

### LINKED ISSUE #445: increase timeout for capabilities refresh
Change this:
https://github.com/nextcloud/richdocuments/blob/v3.2.3/lib/Service/CapabilitiesService.php#L70
from 5 to 10 for example.

5 seconds is the default timeout for dns. 
If dns response failes on first request, request takes more than 5 seconds. As result, capabilities are empty.

@MorrisJobke

MorrisJobke:
Fixes #445


ljo:
I would suggest to add this timeout change to fetchFromRemote() in DiscoveryManager.php too. The narowing down to only refresh in #445 is a bit unlucky since discovery also fails with a 500 on curl timeout errror 28.

PR #460, opened 2019-03-22
Add list tof supported mimetypes to capabilities

juliushaertl:
This PR adds the list of supported mimetypes to the capabilities endpoint, as well as a list of mimetypes that are supported but should not be used as a default handler.

So for PDF we open the regular pdf viewer by default but allow to open it in collabora via the context menu. Same for SVG.

cc @tobiasKaminsky 

This is currently 15+ only, since capabilities are not loaded at the time the actions are registered on older Nextcloud versions.

https://github.com/nextcloud/android/issues/3650

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -34,6 +34,44 @@ use OCP\IURLGenerator;
 
 class Capabilities implements ICapability {
 
+	const MIMETYPES = [
+		'application/vnd.oasis.opendocument.text',
+		'application/vnd.oasis.opendocument.spreadsheet',
+		'application/vnd.oasis.opendocument.graphics',
+		'application/vnd.oasis.opendocument.presentation',
+		'application/vnd.lotus-wordpro',
+		'application/vnd.visio',
+		'application/vnd.wordperfect',
+		'application/msonenote',
+		'application/msword',
+		'application/rtf',
+		'text/rtf',
+		'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
+		'application/vnd.openxmlformats-officedocument.wordprocessingml.template',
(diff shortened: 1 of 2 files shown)


kesselb:
```
curl -X GET \
  https://admin:admin@nextcloud.test/ocs/v1.php/cloud/capabilities \
  -H 'OCS-APIRequest: true' \
  -H 'Postman-Token: 69c970f6-e762-4178-a1d7-ec58171561ea' \
  -H 'cache-control: no-cache'
```

image

thebearon:
After this commit the documents are automatically downloaded instead of opening.

Additonally, I see this in the code "Edit with Collabora online", please capitalize Online. Thank you!

lib/Capabilities.php
```diff
@@ -34,6 +34,44 @@
...
+		'application/vnd.openxmlformats-officedocument.presentationml.slideshow',
+		'application/vnd.ms-powerpoint.addin.macroEnabled.12',
+		'application/vnd.ms-powerpoint.presentation.macroEnabled.12',
+		'application/vnd.ms-powerpoint.template.macroEnabled.12',
+		'application/vnd.ms-powerpoint.slideshow.macroEnabled.12'
+	];
+
+	const MIMETYPES_OPTIONAL = [
+		'image/svg+xml',
+		'application/pdf'
```
tobiasKaminsky: Can we also add text files here?

tobiasKaminsky:
Library: https://github.com/nextcloud/android-library/pull/292
Files app: https://github.com/nextcloud/android/pull/3974

tobiasKaminsky:
@marinofaggiana to stay consistenst across mobile

PR #470, opened 2019-04-10
Nextcloud version bump and code cleanup

### LINKED ISSUE #431: 16 compatibility
We need to do some adjustments to make richdocuments compatible with 16:

- personal settings: file upload seems broken
- admin settings: group select 
```
admin.js:101 Uncaught TypeError: Cannot read property 'setupGroupsSelect' of undefined
```

juliushaertl:
Fixes #431 

This also removes some legacy code we kept for 13/14 compatibility

### DIFF
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -4,61 +4,6 @@
 	<create>true</create>
 	<overwrite>false</overwrite>
 	<charset>utf8</charset>
-	<table>
-		<name>*dbprefix*richdocuments_member</name>
-		<declaration>
-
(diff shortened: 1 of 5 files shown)


kesselb:
https://github.com/nextcloud/richdocuments/blob/90f72c0a44f940e8840115ee03cffe71bd0bce2e/lib/Preview/Office.php#L70-L74 here is some more legacy code ;)

juliushaertl:
@kesselb Thanks for the hint, removed that as well.

Issue #457, opened 2019-03-18
Enhance template creation with file extension


juliushaertl:
When creating a new file from the mobile integration, we need to properly set the file extension depending on the ooxml setting.

PR #472, opened 2019-04-12
Provide ooxml file extensions when set in templates

juliushaertl:
Fixes https://github.com/nextcloud/android/issues/3703
Fixes https://github.com/nextcloud/richdocuments/issues/457

@tobiasKaminsky When fetching the list of templates you now get the proper file extension that should be used for the new file:

image

### DIFF
--- a/lib/TemplateManager.php
+++ b/lib/TemplateManager.php
@@ -73,6 +73,18 @@ class TemplateManager {
 		'presentation' => self::MIMES_PRESENTATIONS
 	];
 
+	public const TYPE_EXTENTION = [
+		'document'     => 'odt',
+		'spreadsheet'  => 'ods',
+		'presentation' => 'odp'
+	];
+
+	public const TYPE_EXTENSION_OOXML = [
+		'document'     => 'docx',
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
When I now create a file with the proper file extension, will then also the right template be used? E.g. when enabling "ooxml" in settings, do I then also get an ooxml file? Or only a file with ooxml suffix?

tobiasKaminsky:
As Android already uses extension this should work out of the box and I do not have to adjust anything on Android?

juliushaertl:
> When I now create a file with the proper file extension, will then also the right template be used? E.g. when enabling "ooxml" in settings, do I then also get an ooxml file? Or only a file with ooxml suffix?

The suffix seems to be sufficient. From what I can tell only opendocument templates are supported right now, but when creating a new file with the ooxml extension the resulting files are proper ooxml format.

juliushaertl:
CI should be fine, once we have dropped testing for PHP 5.6 support with https://github.com/nextcloud/richdocuments/pull/474

juliushaertl:
CI happy, ready to review.

PR #477, opened 2019-04-16
Use product name from collabora capabilities

juliushaertl:


juliushaertl:
cc @tobiasKaminsky @marinofaggiana On android the 3-dots menu shows "Open with collabora" which should probably also use the product name from the capabilities. Not sure about iOS, but everywhere you use "Collabora Online" you should respect the capability value `productName` if present.

tobiasKaminsky:
Do you have an idea how to do this on client side without checking for version or have an ever-lasting backward compatiblity check in it?

juliushaertl:
Hm, no idea unfortunately. I the app provides a proper fallback value of "Collabora Online" but that only helps for richdocuments >= 3.3.0 

I think just supporting the most recent app version should be fine, since we do not support older app versions in general, right @rullzer ?

tobiasKaminsky:
Library: https://github.com/nextcloud/android-library/pull/291
App: https://github.com/nextcloud/android/pull/3971

tobiasKaminsky:
@marinofaggiana to stay consistenst across mobile

PR #480, opened 2019-04-25
Set timeout in case of editor inactivity

eszkadev:
Added 15s timeout what is enough even for a slow mobile
client. If richdocuments doesn't receive Frame_Ready
message then app will be closed.


eszkadev:
This fix prevents an endless spinner (frequently reported issue).

juliushaertl:
Let me add another commit to show a message to the user.

PR #483, opened 2019-04-29
Add new config item public_wopi_url.

rainerjung:
It will be used in addAllowedFrameDomain() and automatically
determined from the Capabilities URL in the discovery response.

* Resolves: #484
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -13,6 +13,7 @@ namespace OCA\Richdocuments\Controller;
 
 use OCA\Richdocuments\Service\CapabilitiesService;
 use OCA\Richdocuments\WOPI\DiscoveryManager;
+use OCA\Richdocuments\WOPI\Parser;
 use \OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
@@ -32,6 +33,8 @@ class SettingsController extends Controller{
 	private $config;
 	/** @var DiscoveryManager  */
 	private $discoveryManager;
+	/** @var Parser */
+	private $wopiParser;
 	/** @var string */
 	private $userId;
 	/** @var CapabilitiesService */
@@ -44,7 +47,9 @@ class SettingsController extends Controller{
 	 * @param AppConfig $appConfig
 	 * @param IConfig $config
 	 * @param DiscoveryManager $discoveryManager
+	 * @param Parser $wopiParser
 	 * @param string $userId
+	 * @param CapabilitiesService $capabilitiesService
 	 */
 	public function __construct($appName,
 								IRequest $request,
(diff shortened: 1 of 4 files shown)


juliushaertl:
Sorry for the delay on this one, I'll have a look at it next week.

rainerjung:
It would be nice if one of the reviewers could have a look. I think it's a rellay useful and simple improvement.

lib/Controller/SettingsController.php
```diff
@@ -152,6 +155,22 @@ public function setSettings($wopi_url,
 		}
 
 		$this->discoveryManager->refretch();
+		try {
+			$capaUrlSrc = $this->wopiParser->getUrlSrc('Capabilities');
+			if (is_array($capaUrlSrc) && $capaUrlSrc['action'] === 'getinfo') {
+				// XXX Works, but probably better to strip trailing '/hosting/capabilities'
+				$public_wopi_url = $capaUrlSrc['urlsrc'];
```
juliushaertl: Since collabora is always using the hostname it was called with this will still be the private hostname that Nextcloud uses to call the discovery endpoint.

rainerjung:
Sorry for being late: The host name that collabora uses in the discovery responses in my setup could get controlled by the following entry in /etc/loolwsd/loolwsd.xml:

<server_name default="" desc="Hostname:port of the server running loolwsd. If empty, it's derived from the request." type="string">**testdocs.kippdata.de**</server_name>

From the comment in that config file, deriving the host name form the request is only the default. That is what is not sufficient when running in an enterprise setup, because the richdocumnts app might need to conact the collabora server via a non-public name.

rainerjung:
It would be nice if you could reopen this PR. I think I can't do it myself. Thanks!

juliushaertl:
I still fail to get a proper use case where you want to have separate urls for user accessed endpoints and the capabilities/discovery endpoint that is called by richdocuments. Any chance you could describe the setup a bit more in detail?

rainerjung:
Nextcloud URL from the Internet:
https://cloud.myorg.com
Collabora URL from the internet:
https://docs.myorg.com

Nextcloud server sits behind a proxy in the core company network. It can not call out to docs.myorg.com. Therefore it must use an internal address:
Collabora URL from Nextcloud:
https://docs.internal.myorg.com

Without the patch, if I configure richdocuments to use the Collabora server under https://docs.myorg.com, the app cannot connect to it due to network separation. If I configure Nextcloud to user https://docs.internal.myorg.com, the Internet Browser will also try to use this address, which will not work. So we need a separation of config aspects, the internet- URL and the internal URL of the Collabora server.
For more details see my initial post.
Thanks.
Rainer

juliushaertl:
OK, thanks for explaining. That makes sense to have then :+1: 

Could you rebase your branch against current master?

rainerjung:
Rebase done.

PR #490, opened 2019-05-08
add txt mimetype to optional

tobiasKaminsky:
Fixes https://github.com/nextcloud/richdocuments/pull/460#discussion_r279279272

Or do we want to add more?
https://wiki.selfhtml.org/wiki/MIME-Type/%C3%9Cbersicht#T

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -70,7 +70,10 @@ class Capabilities implements ICapability {
 
 	const MIMETYPES_OPTIONAL = [
 		'image/svg+xml',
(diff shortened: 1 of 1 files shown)


juliushaertl:
> Or do we want to add more?

I just checked what collabora returns on their discovery endpoint, so I'd say we can also add
- text/rtf 
- text/csv
- text/spreadsheet

tobiasKaminsky:
> I just checked what collabora returns on their discovery endpoint,

Clever :-)

PR #524, opened 2019-05-29
Implement proper file renaming

juliushaertl:
This PR adds support for the X-WOPI-Override: RENAME_FILE header.


juliushaertl:
@tobiasKaminsky @marinofaggiana When renaming a document though collabora, the clients are not aware of the changed filename yet, so sharing after the rename won't work as the file with the old name cannot be found anymore.

What do you think about adding a new method renamed(newFilename) to the RichDocumentsMobileInterface so that the webview can inform the apps about the changed filename? Would that work for you?

juliushaertl:
Let's get this in for now, we can look into the mobile integration in a separate PR.

PR #526, opened 2019-05-30
Allow clients to send a Grab_Focus post message

juliushaertl:
Waiting for client implementation to test with:
- [ ] iOS
- [x] Android

@marinofaggiana @tobiasKaminsky Since this might not be available on older richdocuments versions, the integration should call something like this:

```
webview.evaluateJavascript(
  "if (typeof OCA.RichDocuments.documentsMain.postGrabFocus !== 'undefined') {" + 
  "    OCA.RichDocuments.documentsMain.postGrabFocus();" + 
  "}"
);
```

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -1036,6 +1036,10 @@ var documentsMain = {
 			url: url
 		});
 	},
+
+	postGrabFocus: function() {
+		documentsMain.WOPIPostMessage($('#loleafletframe')[0], 'Grab_Focus');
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Android: https://github.com/nextcloud/android/pull/4106

marinofaggiana:
@juliushaertl  Is available on collabora-test.nextcloud.com ?

PR #530, opened 2019-06-04
Only show session close icons for the file owner

juliushaertl:


juliushaertl:
> However it would also be good to have lool check this of course...

Yes, I'll discuss this with Collabora.

PR #532, opened 2019-06-13
Hide fullscreen buttons when using the app integration

juliushaertl:
Note this is not tested yet, as the current collabora build doesn't seem to have this shipped yet.


tobiasKaminsky:
I can test it, once a new Collabora version is available.
As I have Collabora also locally, I can even checkout any docker git stuff.

juliushaertl:
Tested and works

PR #514, opened 2019-05-24
Start with new template creation mechanism

juliushaertl:
This pull request implements a new template creation mechanism, that will hand over an URL in the TemplateSource entry of CheckFileInfo. If present the file will be downloaded by Collabora during the document creation.

- [X] WOPI endpoint to fetch the template file
- [X] Check for hasTemplateSource capability
- [x] Provide TemplateSource in CheckFileInfo
- [x] Create token with TemplateSource when creating new documents from the web UI
- [x] Create token with TemplateSource when creating new documents from the direct editing API

PR #539, opened 2019-07-04
Federated document editing

juliushaertl:
## Requirements

Collabora currently sets a frame-ancestors CSP so in order to work with all Nextcloud instances the following needs to be set in loolwsd.xml:

      <frame_ancestors>https://*</frame_ancestors>

## TODO

- [x] Get rid of `parent.*` calls in documents.js and use proper postMessages
- [x] Reload files list to get the proper CSP set
- [x] Direct editing
- [x] Set proper WOPI token parameters for federated editors
- [x] Use proper addressbook information for remote editors
- [x] Access file though share token instead of the owners root
- [x] Use dependency injection in various places where it isn't used

## Limitations
- Creating on federated folders will still open newly created documents on the local collabora instance since there is no way we can obtain the template otherwise (This can be fixed once https://github.com/nextcloud/richdocuments/pull/514 is ready)
- Editing remote documents requires a page reload to set a proper CSP. We could set allowed frame domains for some domains per default for GS but I would do that in a later step maybe.

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -435,7 +435,7 @@ class DocumentController extends Controller {
 				$remoteWopi = $this->federationService->getRemoteFileDetails($remoteServer, $remoteServerToken);
 
 				$uid = $remoteWopi['editorUid'] . '@' . $remoteServer;
-				$wopi->setEditorUid($remoteWopi['editorUid']);
+				$wopi->setEditorUid($shareToken);
 				$wopi->setCanwrite($wopi->getCanwrite() && $remoteWopi['canwrite']);
 				$wopi->setRemoteServer($remoteServer);
 				$wopi->setRemoteServerToken($remoteServerToken);
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -170,6 +170,47 @@ class DocumentController extends Controller {
 	}
 
 	/**
+	 * Redirect to the files app with proper CSP headers set for federated editing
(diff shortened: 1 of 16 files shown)


lib/Controller/DocumentController.php
```diff
@@ -160,11 +169,53 @@ private function domainOnly($url) {
...
+			if (!($item instanceof Node)) {
+				throw new \Exception();
+			}
+
+			if ($item->getStorage()->instanceOfStorage(\OCA\Files_Sharing\External\Storage::class)) {
+				$remote = $item->getStorage()->getRemote();
+				$remoteCollabora = $this->federationService->getRemoteCollaboraURL($remote);
+				if ($remoteCollabora !== '') {
+					$fullPath = explode('/', $item->getPath());
+					$dir = implode('/', array_slice($fullPath, 3, -1));
```
juliushaertl: @rullzer Is there a better way to get the path of the node inside of the user folder? getInternalPath() returns the path starting from inside of the storage only

rullzer: On a file, get the absolute path. And then do a getrelative path with that on the userfolder

juliushaertl: Thanks, works like a charm.

juliushaertl:
Github doesn't seem to update the web view, but the recent state is pushed and ready for review/testing. Lets see if it is just a github caching issue.

juliushaertl:
@rullzer rebased and ready for testing and review :wink:

PR #547, opened 2019-07-15
Switch to native font stack

jancborchardt:
As per core change: https://github.com/nextcloud/server/pull/16055, please check @nextcloud/office

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -169,7 +169,7 @@ body {
 }
 
 .loleaflet-font{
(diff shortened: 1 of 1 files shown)


juliushaertl:
@jancborchardt Did you test this somewhere? As far as I can see the .loleaflet-font class is only applied in the footer menu bar and the css file is also not loaded in the actual collabora frame. :/

jancborchardt:
Right, I think actually it loads an example document which uses some default font. I think I set it to Open Sans back then.

I’d still say let’s merge this to fix the footer menu, the document is then fine with Open Sans.

PR #552, opened 2019-07-19
Pass though paste postmessage to mobile apps

juliushaertl:


juliushaertl:
Added to #553 for master

PR #598, opened 2019-09-05
PHP 7.0 compatibility

juliushaertl:


juliushaertl:
Ah `: void` is valid syntax but does cause a runtime error :disappointed: Another reason to have some integration tests.

Issue #594, opened 2019-09-04
App disabled after upgrade on PHP7.0


ralcazar-oeg:
**Describe the bug**
After upgrade to collabora 3.4.0 collabora is disabled. If I click enable nothing happens, and log show error:
"TypeError: Return value of OCA\Richdocuments\AppInfo\Application::updateCSP() must be an instance of OCA\Richdocuments\AppInfo\void, none returned"

Other nextcloud instance use the same collabora server and is working like a charm, so error follows to upgrade v3.4.0 in this test instance.

Each time I go to apps collabora is disabled and I can enabled it, but it does nothing

**Client details:**
Tested in multiple clients and browsers

## Server details

**Operating system**: Ubuntu 16.04

**Web server: Apache 2.4

**Database: mysql

**PHP version: 7.0.33

**Nextcloud version: 15.0.11

**Version of the richdocuments app**

**Version of Collabora Online**

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
[no app in context] Error: TypeError: Return value of OCA\Richdocuments\AppInfo\Application::updateCSP() must be an instance of OCA\Richdocuments\AppInfo\void, none returned at <<closure>>

0. /var/www/alcazar/apps/richdocuments/appinfo/app.php line 66
   updateCSP("*** sensitive parameters replaced ***")
1. /var/www/alcazar/lib/private/legacy/app.php line 261
   require_once("/var/www/alcaza ... p")
2. /var/www/alcazar/lib/private/legacy/app.php line 154
   requireAppFile(OCA\Richdocuments\AppInfo\Application {})
3. /var/www/alcazar/lib/private/legacy/app.php line 127
   loadApp("richdocuments")
4. /var/www/alcazar/apps/dav/appinfo/app.php line 30
   loadApps(["dav"])
5. /var/www/alcazar/lib/private/legacy/app.php line 261
   require_once("/var/www/alcaza ... p")
6. /var/www/alcazar/lib/private/legacy/app.php line 154
   requireAppFile("dav")
7. /var/www/alcazar/lib/private/legacy/app.php line 127
   loadApp("dav")
8. /var/www/alcazar/lib/base.php line 972
   loadApps()
9. /var/www/alcazar/index.php line 42
   handleRequest()

GET /alcazar/index.php/apps/files
from ******* by ralcazar at 2019-09-04T16:23:47+00:00
#### Browser log
```

</details>

Maybe this issue is related to this
[...]

PR #603, opened 2019-09-06
Bugfix: with PHP > 7 the app cannot be enabled anymore

pitchum:
"TypeError: Return value of OCA\Richdocuments\AppInfo\Application::updateCSP() must be an instance of OCA\Richdocuments\AppInfo\void, none returned"

* Resolves: #594

Tested on our production platform.


pitchum:
Just noticed that this fix is already included in PR #598
Sorry for the noise.

PR #621, opened 2019-09-16
Fix missing check for shared files

juliushaertl:


LinneyS:
Thanks for correction
I try to fix too

PR #652, opened 2019-10-07
Fix undefined index warning

juliushaertl:


lib/Controller/WopiController.php
```diff
@@ -606,7 +606,10 @@ private function getFileForWopiToken(Wopi $wopi) {
 			}
 
 			$userFolder = $this->rootFolder->getUserFolder($editor);
-			$file = $userFolder->getById($wopi->getFileid())[0];
+			$files = $userFolder->getById($wopi->getFileid());
+			if (isset($files[0]) && $files[0] instanceof File) {
+				$file = $files[0];
+			}
```
rullzer: Should it not throw a NotFoundException btw?

juliushaertl: Yeah, but there was this sentry error where I'm not sure how it happened: https://sentry.io/share/issue/c9b243560cea45f6bad487f3f7ee6d02/

rullzer: Yes. But in this case that would mean we return null right. If there are no files retruned in getById we should throw a NotFoundException right?

juliushaertl: Ah right, will fix that.

rullzer: added a fixup

juliushaertl: Oh thanks, totally forgot about that. Rebased it

PR #694, opened 2019-11-13
Retry putContent operation if locked

juliushaertl:


lib/Controller/WopiController.php
```diff
@@ -437,7 +439,14 @@ public function putFile($fileId,
...
 			}
 
-			$file->putContent($content);
+			try {
+				$this->retryOperation(function () use ($file, $content){
+					return $file->putContent($content);
+				});
+			} catch (LockedException $e) {
+				$this->logger->logException($e);
+				return new JSONResponse(['message' => 'File locked'], Http::STATUS_INTERNAL_SERVER_ERROR);
```
juliushaertl: We should probably use STATUS_CONFLICT but right now Collabora behaves as if the file was saved successfully if that is returned, so let's keep the 500 status for now

juliushaertl:
Yeah, I'm open for other ideas, but for now this should solve most of the cases when saving fails due files being locked

rullzer:
Yeah let's do it now

PR #712, opened 2019-11-25
Include locale in the loleaflet lang parameter

juliushaertl:
Make sure to apply different locale settings in Collabora


juliushaertl:
OK also seems to work fine with the de_DE language we have for formal german.

PR #796, opened 2020-01-23
Add demo servers

juliushaertl:
Add list of demo servers:
image

Properly show status when connection fails:
image

Migrated advanced settings to vue:
image

## ToDo
- [x] Check if all strings are properly translated
- [ ] Migrate template settings (follow up PR)
- [x] Add provider link
- [x] Cache demo server response
- [ ] Check which server is the fastest responding


rullzer:
> Check which server is the fastest responding

In general sure. But this is more of a nice to have. Most users will know which one to pick there.

juliushaertl:
> On this branch the whole admin settings for richdocuments disappears for me...

:see_no_evil: Argh, indeed I forgot to add the new DemoService file. Fixed.

> In general sure. But this is more of a nice to have. Most users will know which one to pick there.

Yes, was the proposal by Collabora, but since we switched from a default selected demo server to opt-in I'd also leave this out for now.

rullzer:
* [ ] On a clean install (so neither own nor demo server selected it shows the green checkmark).
* [ ] Seems the choice (if you pick demo server) is not rememebed (reload the page and see nothing selected)

juliushaertl:
> On a clean install (so neither own nor demo server selected it shows the green checkmark).
Fixed.

> Seems the choice (if you pick demo server) is not rememebed (reload the page and see nothing selected)

I don't see this happening. The demo servers should actually be stored in a file cache and then read from when reloading the page. Do you see any errors in the console?

PR #805, opened 2020-01-27
Properly check for a set guest name when obtaining the file

juliushaertl:
To reproduce, put a file on external storage and share it with another user that doesn't have access to the external storage. This makes sure that the editor is not falsely overwritten with the owner.

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -391,6 +391,10 @@ class WopiController extends Controller {
 			return new JSONResponse([], Http::STATUS_FORBIDDEN);
 		}
 
+		// Set the user to register the change under his name
+		$this->userScopeService->setUserScope($wopi->getEditorUid());
+		$this->userScopeService->setFilesystemScope($isPutRelative ? $wopi->getEditorUid() : $wopi->getUserForFileAccess());
+
 		try {
 			if ($isPutRelative) {
 				// the new file needs to be installed in the current user dir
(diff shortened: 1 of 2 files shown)


juliushaertl:
@rullzer I pushed two more commits, as with the earlier fix, the filesystem was actually setup for the owner instead of the editor which caused the `OC\Files\View::file_put_contents` to fail as it operated on the file obtained with the editor user scope earlier. Therefore I also moved the filesystem setup to the beginning of the putFile controller method, to avoid any further discrepancies in which user is used for filesystem access.

I think for a code cleanup we could try moving the whole Wopi token handling for which user is acting in the anonymous request to a middleware to make sure the preconditions are properly set for all WopiController methods. I'll look into that in a follow up.

rullzer:
> @rullzer I pushed two more commits, as with the earlier fix, the filesystem was actually setup for the owner instead of the editor which caused the `OC\Files\View::file_put_contents` ~to fail~ to write a new file as the file exists check failed because it operated on the file obtained with the editor user scope earlier. Therefore I also moved the filesystem setup to the beginning of the putFile controller method, to avoid any further discrepancies in which user is used for filesystem access.
> 
> I think for a code cleanup we could try moving the whole Wopi token handling for which user is acting in the anonymous request to a middleware to make sure the preconditions are properly set for all WopiController methods. I'll look into that in a follow up.

yes. That would make a lot of sense indeed.

PR #815, opened 2020-02-02
Bump 3.6.0

juliushaertl:


juliushaertl:
Let's put that on hold for now as file creation seems to have issues on the demo servers.

PR #817, opened 2020-02-04
Do not use void return type hint as we still need to cover PHP 7.0

juliushaertl:


juliushaertl:
> Altough I would also argue dropping php7 support is reasonable. Esp since nextcloud 15 is also EOL

Yeah ... someday ... you know :wink:

PR #819, opened 2020-02-04
Set proper language tag and add some exceptions

### LINKED ISSUE #795: languageToBCP47 function is wrong
languageToBCP47 function is defined in src/helpers/index.js
The comment says: loleaflet expects a BCP47 language tag syntax
However, the function does not provide this syntax. 

The first problem is that region code is converted to lower case. It should be upper case, because the LOKit expects it to be in upper case. So for example, if you pass en-gb to LOKit, it will not understand it, and will use en-US instead of the intended en-GB.

The second problem is that this function creatively mixes language codes and locale codes. For example when the user selects language "English (US)" and locale "German (Switzerland)", the function generates the "en-ch" language code, which makes little sense, at least LOKit does not understand it. Even in theory it is not possible to get the right information from this string, because "ch" is ambiguous. What number format, or weekday names etc. would that imply? German? French? Italian? 

Third problem is that it only matches 2 letter language codes. For example Asturian (ast) will turn into Assamese (as). Quite a difference. :)

Probably loleaflet should accept an optional "locale" parameter besides "lang". And LOKit API should be extended. Currently it has only one "Language" parameter and it is not possible set a different UI language and locale (and default currency), as it is possible in desktop LibreOffice. 

As a first step, languageToBCP47 should return standard BCP47 language tags, preferably only such that loleaflet and LOKit can deal with. 
These are:
ab  ar   be     bo   ca           da   dz     eo  fa  ga  gug  hsb  it  kab  kmr-Latn  ks  lt   ml   my  nn   om     pt     rw     sd   sl       ss     szl  th  ts  ur   vi     zu
af  as   bg     br   ca-valencia  de   el     es  fi  gd  he   hu   ja  kk   kn        ky  lv   mn   nb  nr   or     pt-BR  sah    si   sq       st     ta   ti  tt  uz   xh
am  ast  bn     brx  cs           dgo  en-GB  et  fr  gl  hi   id   jv  kl   ko        lb  mai  mni  ne  nso  pa-IN  ro     sa-IN  sid  sr       sv     te   tn  ug  ve   zh-CN
an  az   bn-IN  bs   cy           dsb  en-ZA  eu  fy  gu  hr   is   ka  km   kok       lo  mk   mr   nl  oc   pl     ru     sat    sk   sr-Latn  sw-TZ  tg   tr  uk  vec  zh-TW

juliushaertl:
Fixes #795 

- [x] Add special handling to map 'de_DE' as a langauge to 'de'
- [x] Check for other whitelist entries that would make sense

### DIFF
--- a/src/helpers/index.js
+++ b/src/helpers/index.js
@@ -23,9 +23,30 @@
 import { getLanguage, getLocale } from 'nextcloud-l10n'
 
 const languageToBCP47 = () => {
+	let language = getLanguage().replace(/_/g, '-')
+	let locale = getLocale()
+
+	// German formal should just be treated as 'de'
+	if (language === 'de-DE') {
+		language = 'de'
+	}
+	// special case where setting the bc47 region depending on the locale setting makes sense
+	const whitelist = {
+		de: {
+			'de_CH': 'de-CH',
+			'gsw': 'de-CH',
+			'gsw_CH': 'de-CH'
+		}
+	}
+	const matchingWhitelist = whitelist[language]
+	if (typeof matchingWhitelist !== 'undefined' && typeof matchingWhitelist[locale] !== 'undefined') {
+		return matchingWhitelist[locale]
+	}
+
 	// loleaflet expects a BCP47 language tag syntax
-	return (getLanguage() + '-' + getLocale())
-		.replace(/^([a-z]{2}).*_([A-Z]{2})$/, (match, p1, p2) => p1 + '-' + p2.toLowerCase())
+	// when a the nextcloud language constist of two parts we sent both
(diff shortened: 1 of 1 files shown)


juliushaertl:
> Please do not convert the 2nd part to lower case. Collabora Online expects upper case.

Right, this is also how it is defined in BCP47, I'll fix that.

juliushaertl:
We don't need any conversion for the default since getLanguage should already return the proper format (e.g. 'en-GB'). So the only special case we handle for now is mapping 'de-DE' to 'de' as language and afterwards add a special case for swiss locale settings that should be passed to Collabora in the lang parameter.

PR #847, opened 2020-03-01
l10n: Changed grammar

rakekniven:
Reported at Transifex.

Does that fit to the function?

Signed-off-by: rakekniven <mark.ziegler@rakekniven.de>


juliushaertl:
> Does that fit to the function? 

Yes, thanks for the fix.

Issue #622, opened 2019-09-16
Collabora Online failes to load, when file is share by public link. No errors logged. 


BjoKaSH:
**Describe the bug**
Collabora Online fails to load with message "Failed to load Collabora Online - please try again later" when a user (not logged in) tries to access the file by password protected public share link.
The messages appears a few seconds after the dialogue box "Select your Nickname" has appeared.  The dialogue box disappears when the messages is shown.

If one is quick enough, one can manage to type a name in the dialogue box and hit enter before the messages appears.  Then the pages reloads, but ultimately fails with the same "Failed to load Collabora Online - please try again later".

If one now manually reloads the page, then the document opens in Collabora Online as expected.

For a shared by link file, this "quickly type something in the 'nickname' box, hit enter, wait for error, reload" seems to be the only way to open a file in Collabora online.

For logged-in users, no problem exists, Collabora Online just works as expected.

**To Reproduce**
Steps to reproduce the behaviour:
1. create an empty folder in NC
2. upload or create an office document (example here: a LibreOffice Calc document)
3. share the file (not the folder!) by link: Set a password, enable editing, disable download. Note down / copy the link
4. log out.  Make sure the NC really don't remember you. 
5. Try to access the link
6. NC will ask for the set password, provide it
7. NC will start loading, showing a "select your nickname" dialogue after a few seconds 
8. do nothing.  do not enter a nickname
9. After several seconds, the dialogue disappears and a message "Failed to load Collabora Online - please try again later" is briefly shown.  Page turns empty (not counting the standard footer)
10. Try reloading the page (remove the automatically added ?# from the URL before), steps 7 - 9 will repeat.  Try several times, if you like.
11. Try reloading the page, but quickly click the "select nickname" box as soon as it appears, type a character and hit enter.  You have (on my box) approx. 3 seconds to do so.
12. A new page will load, but stay empty. After some seconds a message "Failed to load Collabora Online - please try again later" is briefly shown.
13. Try reloading the page.  The document opens as expected.

**Expected behavior**
1. User clicks shared link
2. NC ask for password
3. NC ask for nickname
4. Document loads

**Screenshots**

**Client details:**
 - OS: Windows 7
 - Browser Firefox 60.9.0esr (64-Bit)
 - Device: ThinkPad laptop

## Server details

**Operating system**:  Ubuntu 16.04 LTS

**Web server:**  apache2

**Database:** mysql

**PHP version:**  7.0.33-0ubuntu0.16.04.6

[...]

PR #848, opened 2020-03-03
Properly obtain original source file for public share links

juliushaertl:
* Resolves: #622 
* Target version: master 

### Summary
- Move to database migrations
- Store share token with wopi token entry
  With this we are now able to properly obtain the correct file source if a logged-in user is opening a public share link


juliushaertl:
Integration tests need to be adjusted since the wopi token is not generated anymore without a guest name in 28c7dfd

juliushaertl:
@rullzer Reviews and testing would be highly appreciated :wink:

vpecinka:
thanks!

PR #922, opened 2020-04-24
Add details popover for the demo servers

juliushaertl:
image


juliushaertl:
Rebased so check if ci is now happy.

Issue #954, opened 2020-05-12
3.6.0 demo server: check for obvious sillies


timar:
In order to improve user experience, if the user selects to use public demo servers, we need to check for obvious sillies that won't work for sure.

* localhost as Nextcloud host name
* non-routable IP address of Nextcloud host
* http:// (demo servers use https://)

And warn the user, that their setup will not work.

PR #958, opened 2020-05-14
Resolves: check for obvious sillies #954

mrkara:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -71,11 +71,29 @@
 				<div>
 					<input id="demoserver" v-model="serverMode" type="radio"
 						name="serverMode" value="demo" class="radio"
-						:disabled="updating" @input="fetchDemoServers">
+						:disabled="updating || hasHostErrors" @input="fetchDemoServers">
 					<label for="demoserver">{{ t('richdocuments', 'Use a demo server') }}</label><br>
 					<p class="option-inline">
 						<em>{{ t('richdocuments', 'You can use a demo server provided by Collabora and other service providers for giving Collabora Online a try.') }}</em>
 					</p>
+					<div v-if="hasHostErrors" class="option-inline-emphasized">
+						<p>{{ t('richdocuments', 'Your NextCloud setup is not capable of connecting to the demo servers because;') }}</p>
+						<ul>
+							<li v-if="hostErrors[0]">
+								{{ t('richdocuments', 'it is a local setup (localhost)') }}
+							</li>
+							<li v-if="hostErrors[1]">
+								{{ t('richdocuments', 'it uses an insecure protocol (http)') }}
+							</li>
+							<li v-if="hostErrors[2]">
+								{{ t('richdocuments', 'it is unreachable from the Internet (possibly because of a firewall, or lack of port forwarding)') }}
+							</li>
+						</ul><br>
+						<p>
+							{{ t('richdocuments', 'For use cases like this, we offer instructions for a') }} <a title="Quick tryout with Nextcloud docker" href="https://www.collaboraoffice.com/code/quick-tryout-nextcloud-docker/" target="_blank"
+								rel="noopener" class="external">{{ t('richdocuments', 'Quick tryout with Nextcloud docker.') }}</a>
+						</p>
+					</div>
 					<div v-if="serverMode === 'demo'" class="option-inline">
 						<p v-if="demoServers === null">
 							{{ t('richdocuments', 'Loading available demo servers …') }}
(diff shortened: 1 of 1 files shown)


src/components/AdminSettings.vue
```diff
@@ -71,11 +71,21 @@
 				<div>
 					<input id="demoserver" v-model="serverMode" type="radio"
 						name="serverMode" value="demo" class="radio"
-						:disabled="updating" @input="fetchDemoServers">
+						:disabled="updating || hostErrors.some(x => x)" @input="fetchDemoServers">
 					<label for="demoserver">{{ t('richdocuments', 'Use a demo server') }}</label><br>
 					<p class="option-inline">
 						<em>{{ t('richdocuments', 'You can use a demo server provided by Collabora and other service providers for giving Collabora Online a try.') }}</em>
 					</p>
+					<div v-if="hostErrors.some(x => x)" class="option-inline">
```
juliushaertl, line 79: It would probably be a bit cleaner to move `hostErrors.some(x => x)` to a dedicated data property like hasErrors to make the template a bit more readable

src/components/AdminSettings.vue
```diff
@@ -71,11 +71,21 @@
...
+						<em><u><p>{{ t('richdocuments', 'Your NextCloud setup is not capable of connecting to the demo servers because;') }}</p>
```
juliushaertl, line 80: Please just us a p tag and do the styling using CSS instead of em/u tags. As a general design comment, we only use underlining in Nextcloud for links.

src/components/AdminSettings.vue
```diff
@@ -232,6 +242,7 @@ export default {
 		return {
 			serverMode: '',
 			serverError: Object.values(OC.getCapabilities()['richdocuments'].collabora).length > 0 ? SERVER_STATE_OK : SERVER_STATE_CONNECTION_ERROR,
+			hostErrors: [OC.getHost() === 'localhost' || OC.getHost() === '127.0.0.1', OC.getProtocol() !== 'https', false],
```
juliushaertl, line 245: The global OC.getHost/getProtocol functions are deprecated since they were mainly helpers to make javascript code testable in old days. `window.location.protocol` and `window.location.host` are the suggested repacements to use.

src/components/AdminSettings.vue
mrkara: In fact, that's what I wrote at first even the name was the same. :) But gave an error telling me that there was no var named hostErrors. (Maybe it wasn't generated yet? No idea.) So I went this way.

src/components/AdminSettings.vue
mrkara: Oh. Good to know. I had changed `window.location.protocol` and `window.location.host` to the OC calls on purpose after seeing the OC calls in the code. I'll change them back.

src/components/AdminSettings.vue
mrkara: Got it. :)

juliushaertl:
I just seen that the signoff is also missing, please check https://github.com/nextcloud/richdocuments/pull/958/checks?check_run_id=674531830 for details.

mrkara:
Fixed a one-char typo with the last update.

src/components/AdminSettings.vue
```diff
@@ -71,11 +71,21 @@
...
 					<input id="demoserver" v-model="serverMode" type="radio"
 						name="serverMode" value="demo" class="radio"
-						:disabled="updating" @input="fetchDemoServers">
+						:disabled="updating || hasHostErrors" @input="fetchDemoServers">
 					<label for="demoserver">{{ t('richdocuments', 'Use a demo server') }}</label><br>
 					<p class="option-inline">
 						<em>{{ t('richdocuments', 'You can use a demo server provided by Collabora and other service providers for giving Collabora Online a try.') }}</em>
 					</p>
+					<div v-if="hasHostErrors" class="option-inline-emphasized">
+						<p>{{ t('richdocuments', 'Your NextCloud setup is not capable of connecting to the demo servers because;') }}</p>
```
juliushaertl, line 80:
```suggestion
						<p>{{ t('richdocuments', 'Your NextCloud setup is not capable of connecting to the demo servers because:') }}</p>
```

src/components/AdminSettings.vue
```diff
@@ -71,11 +71,21 @@
...
+						<ul>
+							<li v-if="hostErrors[0]">{{ t('richdocuments', 'it is a local setup (localhost)') }}</li>
```
juliushaertl, line 82: The linter complains about the html single line formatting. Running `npm run lint:fix` should be able to fix that automatically :wink:

mrkara: I had tried `npm run dev --fix`, but seems like this is the right command. Thanks! :)

juliushaertl:
The last commit is missing a signoff message

mrkara:
> The last commit is missing a signoff message ;)

Lol. I didn't know it goes away after --amend's. :)

mrkara:
> > The last commit is missing a signoff message ;)
> 
> Lol. I didn't know it goes away after --amend's. :)

Oh. Got it now. It's because of the last commit I accepted.

juliushaertl:
Yeah, unfortunately there is no way to let github at that automatically when creating commits though the UI

PR #987, opened 2020-05-26
Built-In CODE Server status message on first load

merttumer:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -18,8 +18,25 @@ const PostMessages = new PostMessageService({
 	loolframe: () => document.getElementById('loleafletframe').contentWindow
 })
 
-const showLoadingIndicator = () => document.getElementById('loadingContainer').classList.add('icon-loading')
-const hideLoadingIndicator = () => document.getElementById('loadingContainer').classList.remove('icon-loading')
+const showLoadingIndicator = () => {
+	if (OC.appswebroots.richdocumentscode && Config.get('urlsrc').indexOf('proxy.php') >= 0) {
+		var url = Config.get('urlsrc').substr(0, Config.get('urlsrc').indexOf('proxy.php') + 'proxy.php'.length)
+		$.get(url + '?status').done(function(val) {
+			if (val && val.status && val.status === 'starting') {
+				document.getElementById('proxyLoadingIcon').classList.add('icon-loading-small')
+				document.getElementById('proxyLoadingMessage').textContent = t('richdocuments', 'Built-in CODE Server is starting up shortly, please wait.')
+			}
+		})
+	} else {
+		document.getElementById('loadingContainer').classList.add('icon-loading')
+	}
+}
+
+const hideLoadingIndicator = () => {
+	document.getElementById('loadingContainer').classList.remove('icon-loading')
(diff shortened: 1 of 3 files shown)


src/document.js
```diff
@@ -20,8 +20,20 @@ const PostMessages = new PostMessageService({
 
 const showLoadingIndicator = () => document.getElementById('loadingContainer').classList.add('icon-loading')
 const hideLoadingIndicator = () => document.getElementById('loadingContainer').classList.remove('icon-loading')
+const showLoading = () => {
+	if (OC.appswebroots.richdocumentscode && Config.get('urlsrc').indexOf('proxy.php') >= 0) {
+		$.get(OC.appswebroots.richdocumentscode + '/proxy.php?status').done(function(val) {
```
juliushaertl, line 25: Please add some error handling here in case the request fails for some reason

juliushaertl: Ah actually the timeout handler should then just kick in but there seems to be an issue with the timeout postmessage not reaching the parent frame. I'll have a look at that. Let's get this in so far.

juliushaertl:
@merttumer Please add a signoff mesage to your commit. For details see https://github.com/nextcloud/richdocuments/pull/987/checks?check_run_id=708829273

src/document.js
merttumer: I want to make change before get this in thanks

src/document.js
```diff
@@ -20,8 +20,20 @@ const PostMessages = new PostMessageService({
...
+			if (val && val.status && val.status === 'starting') {
+				$('#proxyLoadingIcon').addClass('icon-loading-small')
```
juliushaertl, line 27: We're actually trying to get rid of jquery :wink: please switch that to some plain javascript 

https://developer.mozilla.org/de/docs/Web/API/Document/getElementById
http://youmightnotneedjquery.com/#add_class
http://youmightnotneedjquery.com/#set_text

merttumer: OK!

src/document.js
```diff
@@ -18,8 +18,25 @@ const PostMessages = new PostMessageService({
...
 
-const showLoadingIndicator = () => document.getElementById('loadingContainer').classList.add('icon-loading')
-const hideLoadingIndicator = () => document.getElementById('loadingContainer').classList.remove('icon-loading')
+const showLoadingIndicator = () => {
+	if (OC.appswebroots.richdocumentscode && Config.get('urlsrc').indexOf('proxy.php') >= 0) {
+		var url = Config.get('urlsrc').substr(0, Config.get('urlsrc').indexOf('proxy.php') + 'proxy.php'.length)
+		$.get(url + '?status').done(function(val) {
+			if (val && val.status && val.status === 'starting') {
+				document.getElementById('proxyLoadingIcon').classList.add('icon-loading-small')
+				document.getElementById('proxyLoadingMessage').innerHTML = t('richdocuments', 'Built-in CODE Server is starting up shortly, please wait.')
```
juliushaertl, line 27: Let's use `textContent` instead of `innerHTML` otherwise translations which are synced from transifex could put random HTML in here.

PR #990, opened 2020-05-26
Inform user about web server configuration issues

merttumer:
Built-in CODE server needs a configuration in the webserver
when it is Nginx, to be able to access proxy.php

Co-authored-by: Muhammet Kara <muhammet.kara@collabora.com>

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -27,7 +27,11 @@
 			<p>{{ t('richdocuments', 'Collabora Online is a powerful LibreOffice-based online office suite with collaborative editing, which supports all major documents, spreadsheet and presentation file formats and works together with all modern browsers.') }}</p>
 
 			<div v-if="settings.wopi_url && settings.wopi_url !== ''">
-				<div v-if="serverError == 2" id="security-warning-state-failure">
+				<div v-if="serverError == 2 && isNginx && serverMode === 'builtin'" id="security-warning-state-failure">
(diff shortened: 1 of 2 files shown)


src/components/AdminSettings.vue
```diff
@@ -343,6 +349,19 @@ export default {
...
 		},
+		checkNginx() {
+			try {
+				axios.get(generateUrl('/apps/richdocuments/settings/webserver')).then((response) => {
+					return response.data
+				}).then(data => {
+					if (data.status === 'success' && data.data.message.indexOf('nginx') >= 0) {
+						this.isNginx = true
+					}
+				})
```
juliushaertl, line 360: We could actually also test if apps/richdocumentscode/proxy.php?status is returning a proper 200 response without following redirects. Then we could only show the error when the status is not starting anymore, which would also solve the issue with a failing connection during loolwsd startup.

lib/Controller/SettingsController.php
```diff
@@ -204,6 +204,24 @@ public function setSettings($wopi_url,
 
 		return new JSONResponse($response);
 	}
+	/**
+	 * @NoAdminRequired
+	 * @throws \Exception
+	 * @return JSONResponse
+	 */
+	public function checkWebServer() {
+		try {
```
juliushaertl, line 213: Maybe you could move the SERVER_SOFTWARE handover to https://github.com/nextcloud/richdocuments/blob/0bd69be42b242b651c25410d0cb6646042b7ff81/lib/Settings/Admin.php#L92 You can directly access it then in the AdminSettings.vue though this.settings without the need for an additional request.

src/components/AdminSettings.vue
merttumer: I guess that will be in a different pr

PR #993, opened 2020-05-27
Auto-enable built-in CODE server when available

mrkara:

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -27,17 +27,23 @@ namespace OCA\Richdocuments\AppInfo;
 use OC\Files\Type\Detection;
 use OC\Security\CSP\ContentSecurityPolicy;
 use OCA\Federation\TrustedServers;
+use OCA\Richdocuments\AppConfig;
 use OCA\Richdocuments\Capabilities;
 use OCA\Richdocuments\Preview\MSExcel;
 use OCA\Richdocuments\Preview\MSWord;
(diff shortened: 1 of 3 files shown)


lib/AppInfo/Application.php
```diff
@@ -131,4 +137,35 @@ public function updateCSP() {
...
+
+            $appConfig->setAppValue('wopi_url', $wopi_url);
+            $appConfig->setAppValue('disable_certificate_verification', 'yes');
+
+            $discoveryManager = $this->getContainer()->query(DiscoveryManager::class);
+            $capabilitiesService = $this->getContainer()->query(CapabilitiesService::class);
+
+            $discoveryManager->refretch();
+            $capabilitiesService->clear();
+            $capabilitiesService->refretch();
```
juliushaertl, line 168: This currently doesn't handle the case that the capabilities request to the built-in server is timing out. In that case $capabiltiesService->getCapabilities() will be an empty array and it will be only refetched after an hour.

My suggestion would be to add a check in https://github.com/nextcloud/richdocuments/blob/967165372c298c20b58fa158b475114c763e9791/lib/Service/CapabilitiesService.php#L56 to see if the built-in server url is configured and then try to refetch the capabilities.

mrkara: Just pushed a new commit to the PR, to handle the capabilities part. I appreciate it if you could check to see if the change makes sense.

PR #994, opened 2020-05-27
Platform and OS checks for the built-in server option

mrkara:

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -80,8 +80,10 @@
 						<em>{{ t('richdocuments', 'easy, just a bit slower than a normal server, and of course not with the usual excellent scalability, but fine for testing or personal use or small teams.') }}</em>
 					</p>
 					<p v-else class="option-inline">
-						<em>{{ t('richdocuments', 'This installation does not have ') }}<a title="Built-in CODE Server App" :href="appUrl" target="_blank"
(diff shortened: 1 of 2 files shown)


src/components/AdminSettings.vue
```diff
@@ -80,7 +80,8 @@
 						<em>{{ t('richdocuments', 'easy, just a bit slower than a normal server, and of course not with the usual excellent scalability, but fine for testing or personal use or small teams.') }}</em>
 					</p>
 					<p v-else class="option-inline">
-						<em>{{ t('richdocuments', 'This installation does not have a built in server') }}</em>
+						<em>{{ t('richdocuments', 'This installation does not have ') }}<a title="Built-in CODE Server App" :href="appUrl" target="_blank"
+							rel="noopener" class="external">{{ t('richdocuments', 'a built in server') }}</a></em>
```
juliushaertl, line 84: Constructing translations like that won't work since the structure of that sentence might be different on other languages. Maybe move to something like 
```suggestion
						<em>
							{{ t('richdocuments', 'This installation does not have a built in server') }}
							<a :href="appUrl" target="_blank">{{ t('richdocuments', 'Install it from the app store.') }}</a>
						</em>
```

mrkara:
> See inline

Thanks! Let me sign & push that.

PR #998, opened 2020-05-29
Re-fetch the proxy status when there is a problem + show a helpful message.

kendy:
This is to show the message all the time there is a problem; and
actually never hide it in case the error is not going to go away by
itself anyway.


juliushaertl:
@kendy Could you add a signoff message to your commit ?

PR #1021, opened 2020-06-08
Add CODEStatusService and use for auto-enable

mrkara:

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -35,6 +35,7 @@ use OCA\Richdocuments\Preview\OOXML;
 use OCA\Richdocuments\Preview\OpenDocument;
 use OCA\Richdocuments\Preview\Pdf;
 use OCA\Richdocuments\Service\CapabilitiesService;
+use OCA\Richdocuments\Service\CODEStatusService;
 use OCA\Richdocuments\Service\FederationService;
 use OCA\Richdocuments\WOPI\DiscoveryManager;
 use OCA\Viewer\Event\LoadViewer;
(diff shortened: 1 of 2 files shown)


mrkara:
We will also use the same service on the admin settings page when user selects the built-in CODE option

mrkara:
@juliushaertl Does it look okay so far? If so, can we merge this, and continue with a follow-up patch?

lib/AppInfo/Application.php
```diff
@@ -168,6 +169,24 @@ public function checkAndEnableCODEServer() {
 			$absoluteUrl = $urlGenerator->getAbsoluteURL($relativeUrl);
 			$wopi_url = $absoluteUrl . '?req=';
 
+            $CODEStatusService = $this->getContainer()->query(CODEStatusService::class);
```
juliushaertl, line 172: The repo uses tabs instead of spaces :wink:

lib/AppInfo/Application.php
```diff
@@ -168,6 +169,24 @@ public function checkAndEnableCODEServer() {
...
+            $statusResponse = $CODEStatusService->checkCODEProxyStatus($wopi_url);
```
juliushaertl, line 173: Hm, so in case an error happens, it would retry on every request and perform a request to the proxy each time. Not sure how to avoid that :/

lib/AppInfo/Application.php
mrkara: > The repo uses tabs instead of spaces wink

Got it. Corrected. IDE claims that it adheres to the current styling of the project, but seems like it failed to do that. :)

mrkara:
Based on the reviews/discussion on the PR, it seems like doing this for the auto-enable part is a bit cumbersome in terms of performance. And without that part, the whole statuscheck service seems like an overkill just to be used on the JS side since we already have platform checks in place, and a status check is done in document.js.

So I propose to close this for now. If the need arises in the future, we can reuse the checkProxyStatus function of document.js any time.

PR #1023, opened 2020-06-09
Fix saving 'disable certificate verification'

CySlider:
* Target version: master 

### Summary

The value is never fully equal to 'true', so checking on this option is never saved.
Instead it is fully equal to the boolean value true

### Checklist

- [ ] Sign-off message is added to all commits (Dunno)


juliushaertl:
Thanks a lot. Could you please sign off your commit. Details can be found here: https://probot.github.io/apps/dco/

juliushaertl:
Let's get that in without DCO since it is a very minor change

CySlider:
Thanks, sorry I did not even check it out but did it completely in Github and could not see this option as I have it in my IDE.

PR #1033, opened 2020-06-17
added UI defaults options to AdminSettings

merttumer:

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -193,6 +193,29 @@
 				</div>
 			</settings-checkbox>
 		</div>
+		<div v-if="isSetup" id="ui-defaults" class="section">
+			<h2>{{ t('richdocuments', 'UI Defaults') }}</h2>
+			<settings-checkbox v-model="settings.uidefaults.enabled" :label="t('richdocuments', 'Override UI Defaults')" hint=""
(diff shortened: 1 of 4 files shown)


merttumer:
This patch depends on unreleased feature, let's wait until it is released, thanks

src/components/AdminSettings.vue
```diff
@@ -193,6 +193,29 @@
 				</div>
 			</settings-checkbox>
 		</div>
+		<div v-if="isSetup" id="ui-defaults" class="section">
+			<h2>{{ t('richdocuments', 'UI Defaults') }}</h2>
+			<settings-checkbox v-model="settings.uidefaults.enabled" :label="t('richdocuments', 'Override UI Defaults')" hint=""
+				:disabled="updating" @input="updateUIDefaults" />
```
juliushaertl, line 199: I actually don't think we should have individual settings for this and instead pick sane defaults, what do you think @jancborchardt

merttumer: loolwsd.xml will also have these options and this will override them if checked.

jancborchardt: Agree with @juliushaertl here, we shouldn't expose these as settings and simply set our defaults.

PR #1109, opened 2020-08-12
Fix incorrect template file for presentation

merttumer:
Current template file is problematic with Master slides
It should be replaced with this one.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy:
Makes perfect sense to merge, yes :-)

PR #1110, opened 2020-08-12
Revert "Fix incorrect template file for presentation"

kendy:
Reverts nextcloud/richdocuments#1109


kendy:
Hmm, it turns out that we'll need an update of the template, the 2nd slide contains some text...

PR #1115, opened 2020-08-21
Use proper base template to be compatible with Nextcloud 20

juliushaertl:
Otherwise we will have two headerbars being rendered. This also works just fine with older Nextcloud versions.


juliushaertl:
Oops just created twice https://github.com/nextcloud/richdocuments/pull/1115

Issue #1151, opened 2020-09-10
Watermarking not working


WindfallProphet:
**Describe the bug**
Enabled watermark, but no watermark present.

**To Reproduce**
1. Enable Watermarking

**Expected behavior**
Expected to see watermark on documents

**Client details:**
 - OS: Arch Linux
 - Browser Firefox
 - Version  80.1
 - Device: x64 deskdop

## Server details
**Operating system**:
Centos 8 x86_64
**Web server:**
nginx/1.14.1
**Database:**
mysql v.15.1 -10.3.17-mariadb
**PHP version:**
php 7.4.10 
Modules [php]: apcu, bcmath, bz2, calendar, Core, ctype, curl, date, dom, exif, fileinfo, filter, ftp, gd, gettext, gmp, hash, iconv, imagick, intl, json, ldap, libxml, mbstring, mysqli, mysqlnd, openssl, pcntl, pcre, PDO, pdo_mysql, pdo_sqlite, Phar, posix, readline, Reflection, session, shmop, SimpleXML, sockets, sodium, SPL, sqlite3, standard, sysvmsg, sysvsem, sysvshm, tokenizer, xml, xmlreader, xmlwriter, xsl, Zend OPcache, zip, zlib
**Nextcloud version:**
Nextcloud 19.0.2
**Version of the richdocuments app**
3.7.4
**Version of Collabora Online**
4.2.6-2

<details>
<summary>Logs</summary>
"message":"Undefined index: watermark_{text,allTags,allGroups,linkTags,linkAll,linkRead,linkSecure,shareRead,enabled} at /var/www/nextcloud/apps/richdocuments/lib/Controller/SettingsController.php#237","userAgent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.83 Safari/537.36","version":"19.0.2.2","id":"5f596d10c293c"}
</details>

WindfallProphet:
Might be a Duplicate of #911

juliushaertl:
Thanks for reporting, I'm not sure where or how this broke but https://github.com/nextcloud/richdocuments/pull/1152 should fix it, testing and review are very welcome

WindfallProphet:
I'm getting no more watermark complaints in nextcloud.log, but I'm unable to view a file anonymously. I would have sworn this was working, but both pre and post #1152 are having the same issue. 

Watermarking disabled → anonymous browsing works
[...]

PR #1152, opened 2020-09-10
Properly check value types when updating watermark settings

juliushaertl:
I'm not sure why this has worked in the past, however this should fix issues when saving the watermark settings.

* Resolves: #1151


rullzer (approved):
:elephant:

juliushaertl:
Tests run fine locally.

PR #1153, opened 2020-09-10
Very minimal document templates for Collabora Online

timar:
These files contain no data at all. This guarantees that when the
user opens a new document in the app, the language of paragraph,
page size, cell date format, currency, etc. will be according to
the current locale.


juliushaertl (approved):
Sure :+1:

juliushaertl:
@timar I'm looking into preparing similar template for draw files. Is that something you could provide me with or do you have any hints on how i could create one myself? 

I've found that the language seems to be within styles.xml `fo:language="de" fo:country="DE"` but maybe there are other things to watch out for as well here.

PR #1180, opened 2020-10-09
Ensures <iframe> contains a non-empty title attribute

pedropintosilva:


pedropintosilva:
I noticed that there is no title and was wondering if this is the way to go

juliushaertl (approved):
Yep, doesn't hurt to have it :)

PR #1198, opened 2020-10-16
Use correct call to notify mobile

eszkadev:
This fixes copy/paste of the images in the document


timar:
@juliushaertl This fixes ADR-84.

kendy (approved):
I can confirm this works, and additionally fixes the ADR-83 too...

PR #1211, opened 2020-10-20
Passing some UI Defaults to loleaflet frame

merttumer:
These ui defaults determine some widget states
at the start of the document which is hidden or shown
Some of them such as Ruler Statusbar and Sidebar can be
hidden at start with this patch. If not set, collabora
online will assume they are true

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -80,6 +80,46 @@ const hideLoadingIndicator = () => {
 	document.getElementById('proxyLoadingMessage').textContent = ''
 }
 
+const generateCSSVarTokens = () => {
+	/* NC versus COOL */
+	var cssVarMap = {
+		'--color-main-text': '--co-color-main-text',
+		'--color-main-background': '--co-body-bg',
+		'--color-primary-text': '--co-primary-text',
+		'--color-primary-element': '--co-primary-element:--co-text-accent',
+		'--color-primary-element-light': '--co-primary-element-light',
+		'--color-error': '--co-color-error',
+		'--color-warning': '--co-color-warning',
+		'--color-success': '--co-color-success',
+		'--border-radius': '--co-border-radius',
+		'--border-radius-large': '--co-border-radius-large',
+		'--color-background-hover': '--co-background-hover',
+		'--color-background-dark': '--co-background-dark',
+		'--color-text-light': '--co-text-light',
+		'--color-text-lighter': '--co-text-lighter',
+		'--color-loading-light': '--co-loading-light',
+		'--color-loading-dark': '--co-loading-dark',
+		'--color-box-shadow': '--co-box-shadow',
+		'--color-border': '--co-border',
+		'--color-border-dark': '--co-border-dark',
+		'--border-radius-pill': '--co-border-radius-pill'
+	}
+	var str = ''
+	for (var cssVarKey in cssVarMap) {
+		var cStyle = window.parent.getComputedStyle(document.documentElement).getPropertyValue(cssVarKey)
+		if (!cStyle) {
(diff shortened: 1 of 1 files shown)


merttumer:
@juliushaertl Can you please review these patches? Thank you

src/document.js
```diff
@@ -80,6 +80,46 @@ const hideLoadingIndicator = () => {
...
+		'--color-loading-light': '--co-loading-light',
+		'--color-loading-dark': '--co-loading-dark',
+		'--color-box-shadow': '--co-box-shadow',
+		'--color-border': '--co-border',
+		'--color-border-dark': '--co-border-dark',
+		'--border-radius-pill': '--co-border-radius-pill'
+	}
+	var str = ''
+	for (var cssVarKey in cssVarMap) {
+		var cStyle = window.parent.getComputedStyle(document.documentElement).getPropertyValue(cssVarKey)
```
juliushaertl, line 109:
```suggestion
		var cStyle = window.getComputedStyle(document.documentElement).getPropertyValue(cssVarKey)
```

src/document.js
```diff
@@ -80,6 +80,46 @@ const hideLoadingIndicator = () => {
...
+		if (!cStyle) {
+			// try suffix -dark instead
+			cStyle = window.parent.getComputedStyle(document.documentElement).getPropertyValue(cssVarKey + '-dark')
```
juliushaertl, line 112:
```suggestion
			cStyle = window.getComputedStyle(document.documentElement).getPropertyValue(cssVarKey + '-dark')
```

src/document.js
```diff
@@ -176,7 +216,9 @@ const documentsMain = {
 
 			// form to post the access token for WOPISrc
 			const form = '<form id="loleafletform_viewer" name="loleafletform_viewer" target="loleafletframe_viewer" action="' + urlsrc + '" method="post">'
-				+ '<input name="access_token" value="' + accessToken + '" type="hidden"/></form>'
+				+ '<input name="access_token" value="' + accessToken + '" type="hidden"/>'
+				+ '<input name="ui_defaults" value="TextRuler=false;TextStatusbar=true;TextSidebar=false;PresentationSidebar=false;PresentationStatusbar=true;SpreadsheetSidebar=false" type="hidden"/'
```
juliushaertl, line 220:
```suggestion
				+ '<input name="ui_defaults" value="TextRuler=false;TextStatusbar=true;TextSidebar=false;PresentationSidebar=false;PresentationStatusbar=true;SpreadsheetSidebar=false" type="hidden"/>'
```

juliushaertl (changes requested):
See inline for smaller comments. In general window.parent should be avoided as it doesn't work across different domains, which might occur in federated or global scale setups.

Otherwise :+1: from my side.

src/document.js
juliushaertl: The CSS variables are also properly loaded inside of the document frame.

src/document.js
merttumer: Thank you for the reply. Initially it was like this but later we discovered that it does not work on Firefox like that. The CSS variables become <empty string>. The only solution I found was using window.parent.

juliushaertl: Indeed seems to be a firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1471231

Then let's get this in as it is and maybe see if we can work around that by not using `display: none;` but opacity of 0 until the frame is loaded.

juliushaertl (approved):

PR #1226, opened 2020-11-06
Move Collabora endpoint caching to distributed cache

juliushaertl:
Moves capabilties and discovery caching from appdata to a proper cache like redis. This should improve the performance avoiding unneeded file reads as discussed with @rullzer and @MorrisJobke

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -186,7 +186,7 @@ class Application extends App {
 
 			$discoveryManager->refretch();
 			$capabilitiesService->clear();
-			$capabilitiesService->refretch();
+			$capabilitiesService->refetch();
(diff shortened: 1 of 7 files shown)


rullzer:
> Moves capabilties and discovery caching from appdata to a proper cache like redis. This should improve the performance avoiding unneeded file reads as discussed with @rullzer and @MorrisJobke

yep. We recommend caches anyways. Esp on slow storages this should help a lot.

rullzer (approved):
Did not test. But it looks sane. Seems to also cleanup some logic along the way. Nice.

rullzer (approved):

MorrisJobke (approved):
Code makes sense 👍

PR #1238, opened 2020-11-10
Move to @nextcloud/capabilities and only register default mime types for viewer

juliushaertl:
Fixes https://github.com/nextcloud/server/issues/23522


rullzer (approved):
yay more js packages

PR #1243, opened 2020-11-12
Add migration to bigint columns

juliushaertl:


rullzer (approved):
Fine by me.
But there is also https://github.com/nextcloud/server/blob/2a054e6c04e0a40421510eb889cbf59f153c5177/lib/public/Migration/BigIntMigration.php#L34

nickvergessen (changes requested):
Please use the repair step way instead.

nickvergessen:
Add it to https://github.com/nextcloud/server/blob/3f691c52dcced41aadf5e41b21856dd6bdd92fa0/core/Command/Db/ConvertFilecacheBigInt.php#L54-L75
it already contains apps anyway. Maybe it should be migrated to the way AddMissing* commands work, so apps can do it via hooks instead.

MorrisJobke:
> Please use the repair step way instead.

So the way to go is, that this change should be added to an existing migration. In that way it is already applied on new installations, but not triggered on existing installations (to not cause huge downtimes). Then we add something to the server.

juliushaertl:
> Maybe it should be migrated to the way AddMissing* commands work, so apps can do it via hooks instead.

I guess that makes sense, yes. Adding more hardcoded app tables to server doesn't seem right to me.

> So the way to go is, that this change should be added to an existing migration. In that way it is already applied on new installations, but not triggered on existing installations (to not cause huge downtimes). Then we add something to the server.

Sounds good :+1:

juliushaertl:
Added the migration as `occ richdocuments:convert-bigint` for now so we can get this out as a release.

For future enhancements we can move this to https://github.com/nextcloud/server/issues/24100 once it is there.

lib/Migration/Version2060Date20200302131958.php
```diff
@@ -44,11 +44,11 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 				'notnull' => false,
 				'length' => 64,
 			]);
-			$table->addColumn('fileid', 'integer', [
+			$table->addColumn('fileid', 'bigint', [
 				'notnull' => true,
 				'length' => 4,
```
nickvergessen, line 49: drop or 20?

nickvergessen (approved):

PR #1256, opened 2020-11-16
Adapt built-in CODE url on host address change

mrkara:

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kyrofa:
This does appear to fix the issue, but it's interesting... richdocuments keeps disabling itself. It seems I can only enable it from the UI, not `occ` (and it's disabled upon initial install using localhost as the URL), and if I switch URLs is disables itself again. That seems related to this change, but I'm not familiar enough with the codebase to understand what's happening.

juliushaertl (approved):

PR #1259, opened 2020-11-18
Don't use a stream response on an empty file

rullzer:
Streaming an empty file does boom currently.
So lets use a differnt response for that.


timar (approved):
I tested it, works OK. Thanks!

PR #1266, opened 2020-11-23
Fix issues with federated editing in global scale setups

juliushaertl:
- Do not fail if computed styles cannot be obtained due to cross origin iframes
- Allow defining an allow list through gs.trustedHosts in config.php:
With this either a domain pattern like *.example.com can be defined or individual domains can be added as trusted sources for federated editing. Having individual domains listed has the benefit that the CSP can be set right away and there is no need to perform an additional reload for opening a document.

```
  // provide a whitelist pattern, this will require the files app to reload 
  // to open a document in order to set the proper csp for the opened document on the remote server
  'gs.enabled' => true,
  'gs.trustedHosts' => [
    '*.local.dev'
  ],
```

or

```
  // immediately add all trusted hosts to csp when loading the files app
  'gs.enabled' => true,
  'gs.trustedHosts' => [
    'nc1.local.dev',
    'nc2.local.dev'
  ],
```

As discussed with @rullzer, ideally we would just obtain the list of hosts from the lookup server once https://github.com/nextcloud/lookup-server/pull/43 is ready.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -41,6 +41,7 @@ use OCA\Viewer\Event\LoadViewer;
 use OCP\AppFramework\App;
 use OCP\AppFramework\QueryException;
 use OCP\EventDispatcher\IEventDispatcher;
+use OCP\GlobalScale\IConfig;
 use OCP\IPreview;
 
 class Application extends App {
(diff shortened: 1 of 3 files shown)


rullzer (approved):
Didn't test but looks good

PR #1269, opened 2020-11-24
Increase timeout if proxy is starting

eszkadev:

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/WOPI/DiscoveryManager.php
+++ b/lib/WOPI/DiscoveryManager.php
@@ -77,34 +77,8 @@ class DiscoveryManager {
 			$options['verify'] = false;
 		}
 
-		// first load with proxy can take some time - increase timeout in that case
-		$usesProxy = false;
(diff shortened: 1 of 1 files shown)


lib/WOPI/DiscoveryManager.php
```diff
@@ -77,6 +77,35 @@ public function fetchFromRemote() {
 			$options['verify'] = false;
 		}
 
+		// first load with proxy can take some time - increase timeout in that case
+		$usesProxy = false;
```
juliushaertl, line 81: Would probably be nice to move this into a dedicated function so the general flow is a bit more readable and we might be able to reuse this in case we need it for any other request that might run into the default timeout.

juliushaertl (approved):

juliushaertl (approved):

PR #1300, opened 2020-12-30
Update files.js

ealmonte32:
fixes proper settings url for: `Settings > Administration > Collabora Online `

* Target version: master 

screenshots below show issue and fix by including the `/index.php/` dir:

Screen Shot 2020-12-30 at 2 58 36 PM
Screen Shot 2020-12-30 at 2 58 43 PM

<br>
This is the proper URL with fix:

Screen Shot 2020-12-30 at 3 27 50 PM

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -52,7 +52,7 @@ const odfViewer = {
 
 	onEdit: function(fileName, context) {
 		if (!odfViewer.isCollaboraConfigured) {
-			const setupUrl = OC.generateUrl('/settings/admin/richdocuments')
+			const setupUrl = OC.generateUrl('/index.php/settings/admin/richdocuments')
 			const installHint = OC.isUserAdmin()
 				? `<a href="${setupUrl}">Collabora Online is not setup yet. <br />Click here to configure your own server or connect to a demo server.</a>`
 				: t('richdocuments', 'Collabora Online is not setup yet. Please contact your administrator.')


juliushaertl:
Please check your server configuration if you have somehow misconfigured pretty urls, this call works fine on my different setups with and without index.php

ealmonte32:
> Please check your server configuration if you have somehow misconfigured pretty urls, this call works fine on my different setups with and without index.php

I did not change ANY server configuration at all, I literally did a fresh install, went to Apps, installed the collabora and other apps, went to that settings area for collabora online and that is what happens, so not sure what pretty urls settings you are referring to, but I can assure you I did not change any "pretty url" settings.

Also, your different setups, when was the last time you did a fresh install and tried this? This install I did was a few days ago and on an ubuntu server and it was the only package installed after installing the server core itself..

juliushaertl:
Can you share your config php?

ealmonte32:
> Can you share your config php?

I have the /snap/nextcloud, does that make any difference?

Here is config.php:
https://pastebin.com/8m11sb4d

juliushaertl:
Ah so the page you are pointing is actually part of the richdocumentscode app, where the urls are not constructed correctly:

https://github.com/CollaboraOnline/richdocumentscode/blob/master/lib/Settings/Admin.php#L37-L39

Please open an issue or pull request at https://github.com/CollaboraOnline/richdocumentscode for that.

PR #1303, opened 2021-01-04
Cut of guest names so they fit into the database

nickvergessen:


juliushaertl (approved):
Does the trick for now :+1:

PR #1327, opened 2021-01-27
Handle errors when fetching remote file info more gracefully

juliushaertl:
This will handle errors that may occur when fetching the file info of a remote wopi token fails more gracefully and enhances the logging so that logs will actually help to understand what is going wrong there afterwards.

### DIFF
--- a/lib/Service/FederationService.php
+++ b/lib/Service/FederationService.php
@@ -173,7 +173,7 @@ class FederationService {
 				]
 			]);
 			$responseBody = $response->getBody();
-			$data = \json_decode($responseBody, true, 512, JSON_THROW_ON_ERROR);
+			$data = \json_decode($responseBody, true, 512);
 			$this->logger->debug('Reveived remote file details for ' . $remoteToken . ' from ' . $remote . ': ' . $responseBody);
 			return $data['ocs']['data'];
 		} catch (\Throwable $e) {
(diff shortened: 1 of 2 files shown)


lib/Service/FederationService.php
```diff
@@ -172,10 +172,12 @@ public function getRemoteFileDetails($remote, $remoteToken) {
...
 				]
 			]);
-			$data = \json_decode($response->getBody(), true);
+			$responseBody = $response->getBody();
+			$data = \json_decode($responseBody, true, 512, JSON_THROW_ON_ERROR);
+			$this->logger->debug('Reveived remote file details for ' . $remoteToken . ' from ' . $remote . ': ' . $responseBody);
 			return $data['ocs']['data'];
 		} catch (\Throwable $e) {
-			$this->logger->info('Unable to determine collabora URL of remote server ' . $remote);
+			$this->logger->error('Unable to fetch remote file details for ' . $remoteToken . ' from ' . $remote, ['exception' => $e]);
```
rullzer, line 180: Does this really need to be an error?

juliushaertl: Yes, in that case the file cannot be opened.

rullzer: ah ok :+1:

PR #1328, opened 2021-01-27
Prefix global scale nodes with the current protocol 

juliushaertl:
to obtain the remote collabora url properly in https://github.com/nextcloud/richdocuments/blob/a89f9bfc443876be527a0ed2b22bfbf315f54f19/lib/Service/FederationService.php#L81

I couldn't reproduce an actual issue with it in cases where there would be a proper redirect for http -> https but we can save that in such cases and this is less error prone.


juliushaertl:
Closing in favor of #1332

Issue #1341, opened 2021-02-03
Support Flat ODF files


timar:
The OpenDocument format supports the following two ways of document representation:
- zipped package of XML files plus additional binaries (images, etc.)
- single, flat, uncompressed XML file – this is called flat ODF

richdocuments supports only the first, although Collabora Online is capable of handling Flat ODF. 
Relevant file extensions and mime types:
.fodt application/vnd.oasis.opendocument.text-flat-xml
.fods application/vnd.oasis.opendocument.spreadsheet-flat-xml
.fodp application/vnd.oasis.opendocument.presentation-flat-xml

timar:
I tried #1342.
- Applied the patch.
- My config//mimetypemapping.json 
`
{
	"fodg": ["application/vnd.oasis.opendocument.graphicsi-flat-xml"],
	"fodp": ["application/vnd.oasis.opendocument.presentation-flat-xml"],
	"fods": ["application/vnd.oasis.opendocument.spreadsheet-flat-xml"],
	"fodt": ["application/vnd.oasis.opendocument.text-flat-xml"]
}
`
- My config/mimetypealiases.json
`
{
	"application/vnd.oasis.opendocument.presentation-flat-xml": "x-office/presentation",
	"application/vnd.oasis.opendocument.spreadsheet-flat-xml": "x-office/spreadsheet",
	"application/vnd.oasis.opendocument.text-flat-xml": "x-office/document"
}
`
- I ran `sudo -u www-data php occ maintenance:mimetype:update-js`

No effect :( I could not open my sample .fods file.

PR #1342, opened 2021-02-03
Support Flat ODF files

tintou:
* Resolves: #1341
* Target version: master 

### Summary

Adds missing supported types

### Checklist


juliushaertl:
Did you add some custom mimetype mapping to your nextcloud instance when testing this. As far as I see we currently cannot detect the flat odf mimetype properly within Nextcloud server and using mimetype detection on such a file itself at least for me always brings up text/xml for fodt files.

juliushaertl:
Thanks a lot for your pull request and we'll get your commit there merged with the new PR.

PR #1353, opened 2021-02-11
Apply new file name after rename

eszkadev:
If file was renamed inside editor and then
SaveAs dialog was opened the old name was sill
present. This fixes that case.


juliushaertl (approved):

juliushaertl:
Test failures are unrelated, most likely due to github actions using a newer ubuntu release now as we've seen that in other apps as well

PR #1386, opened 2021-02-25
Added IP Address option to the watermark

WindfallProphet:
...{ipAddr}
* Target version: master 

### Summary
I thought it was important to have IP address as an option on the watermark, especially if the file is shared publicly. 

This is working well on my machine, but if there's a better way to implement, let me know. 

### Checklist

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -196,12 +196,27 @@ class WopiController extends Controller {
 
 		if ($this->shouldWatermark($isPublic, $wopi->getEditorUid(), $fileId, $wopi)) {
 			$email = $user !== null && !$isPublic ? $user->getEMailAddress() : "";
+			if (!empty($_SERVER['HTTP_CLIENT_IP']))   
+			  {
+			    $ipAddr = $_SERVER['HTTP_CLIENT_IP'];
+			  }
+			//whether ip is from proxy
+			elseif (!empty($_SERVER['HTTP_X_FORWARDED_FOR']))  
+			  {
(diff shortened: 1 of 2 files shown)


lib/Controller/WopiController.php
```diff
@@ -196,12 +196,27 @@ public function checkFileInfo($fileId, $access_token) {
 
 		if ($this->shouldWatermark($isPublic, $wopi->getEditorUid(), $fileId, $wopi)) {
 			$email = $user !== null && !$isPublic ? $user->getEMailAddress() : "";
+			if (!empty($_SERVER['HTTP_CLIENT_IP']))
```
juliushaertl, line 199: The issue with this is that the request to the checkFileInfo endpoint is not done by the user but by the Collabora server, so we will not get the actual users ip here unless it is somehow sent to us by the Collabora server. 

cc @timar @thebearon I think that would be a good enhancement to the watermarking feature. Would this be something you could imagine implementing on the Collabora side, so that the WOPI requests would send over the original IP of the user in some additional header?

juliushaertl:
@WindfallProphet Thanks for your contribution. I'll be closing this for now as we will need support for getting the real request IP from collabora online. Feel free to open a feature request over at https://github.com/collaboraonline/online in that regard. We can then reopen this to finish once the requirements are there.

WindfallProphet:
The idea came about when we were trying to figure out how to run a remote class and share confidential materials with anonymous users. 

I'm open to doing whatever you want in regards to issues. It seems like a really good feature to have, though the use case might be limited.

Issue #1334, opened 2021-01-28
Dark theme colors


drikqlis:
**Describe the bug**
When using dark theme or breeze-dark theme menus are unreadable. See https://github.com/mwalbeck/nextcloud-breeze-dark/issues/183 and https://github.com/nextcloud/richdocuments/issues/1270

**To Reproduce**
1. Enable dark theme or breeze-dark
2. Open any document in Collabora.

**Screenshots**

**Client details:**
 - OS: [e.g. iOS] Windows 10
 - Browser [e.g. chrome, safari] Chrome, Firefox
 - Device: [e.g. iPhone6, desktop]

## Server details

**Operating system**: Ubuntu 18.04 LTS

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.3

**Nextcloud version:** 20.0.6

**Version of the richdocuments app**

**Version of Collabora Online**

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

Fuseteam:
screenshot showcasing the bug on version 20
![image](https://user-images.githubusercontent.com/10421851/108228752-eda5fa00-711d-11eb-9315-0c595e56f7d5.png)

elpraga:
I am having the same issue, it happens also on Android and when using the Android app. It is especially bad when selecting fonts since their names cannot be seen. 

Should this issue be reported here or on the Breeze Dark issue tracker?

juliushaertl:
A fix was already proposed in https://github.com/nextcloud/richdocuments/pull/1394

Issue #229, opened 2018-07-21
[Feature] Make NC theming match Collabora theme


Alquh:
**Steps to reproduce**
Install NC and theming app with Collabora
Navigate to settings and enable theming (if not enabled)
Change logo and color theme
Navigate to Collabora

**Expected behaviour**
Collabora's logo and color theme should match Theming settings

**Actual behaviour**
Collabora's default logo and theme are appearing

Rational for change: Theming for Collabora would be nice if it matched Nextcloud

juliushaertl:
Fixed with https://github.com/nextcloud/richdocuments/pull/1211

PR #1394, opened 2021-02-28
Improve dark theme experience

mwalbeck:
* Resolves: #1334 
* Target version: master 

### Summary
This pull request removes background colours and text colours from the list of variables being passed on to Collabora. Those colour variables aren't used for a lot of elements so they end up not really making a difference with the default theme but gives a bad dark theme experience.

From what I can tell, it isn't currently possible, to provide a proper dark theme experience using only colour variables in Collabora.

This won't affect #229 

### Checklist

### Screenshots
before
dark-before
after
dark-after
after with red primary colour
dark-red

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -83,8 +83,6 @@ const hideLoadingIndicator = () => {
 const generateCSSVarTokens = () => {
 	/* NC versus COOL */
 	var cssVarMap = {
-		'--color-main-text': '--co-color-main-text',
-		'--color-main-background': '--co-body-bg',
 		'--color-primary-text': '--co-primary-text',
 		'--color-primary-element': '--co-primary-element:--co-text-accent',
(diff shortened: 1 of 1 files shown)


juliushaertl:
Seems like a good solution to me as there is no dark mode in Collabora Online as of today, right @merttumer ?

kendy:
@mwalbeck, @juliushaertl: @merttumer will work on improving the dark theme shortly, but indeed, might be a good idea to merge the patch in the meantime, to avoid problems before it's done.

kendy (approved):
Let's do this before we improve the dark mode in Collabora Online.

juliushaertl (approved):

juliushaertl:
Thanks a lot for your contribution on that @mwalbeck :tada:

Issue #1379, opened 2021-02-24
Writing documents with enabled per-user encryption keys fails (Private Key missing for user)


marioklump:
**Describe the bug**
When saving a document with enabled per-user encryption keys, an exception `Private Key missing for user: please try to log-out and log-in again` is thrown. I debugged it and got the following insights:

- Before the document is passed to collabora, a new public share is added for the document (https://github.com/nextcloud/richdocuments/blob/master/lib/Controller/DocumentController.php#L270).
- Collabora is able to fetch the document by using the passed access_token. The document is fetch in incognito mode (https://github.com/nextcloud/richdocuments/blob/master/lib/Controller/WopiController.php#L349).
- When writing the document back to the nextcloud storage, the incognito mode isn't used. Instead the user is set to the document's editor to be able to trace the document's changes (https://github.com/nextcloud/richdocuments/blob/master/lib/Controller/WopiController.php#L407). This causes the usage of the user-specific encryption key.

I replaced the explicit setting of the user by enabling the incognito mode and tested it in my nextcloud setup: https://github.com/mpk-software/richdocuments/commit/a0fed2b940b735eb896559a8fd6cd70232d16cd5

This causes to use the public share key for writing the document instead of trying to determine the current user's encryption key (which seems not to be present in the session established by the collabora server). My above workaround is working and I'm able to save documents with enabled per-user-encryption.

What do you think is the best option for solving the problem with per-user encryption keys? Would it be okay to loose the information about who did the changes on the document (the goal of https://github.com/nextcloud/richdocuments/commit/ada925e4c5031213bf28f6792dd6a84548d05a1c)? Should we only use the incognito mode if per-user encryption is enabled?

**To Reproduce**
1. Enable per-user encryption.
2. Open a document by using collabora.
3. Click the "save" icon
4. Check the logs in Nextcloud's admin section (or `nextcloud.log` file).

**Expected behavior**
The document should be saved even if per-user-encryptions keys are enabled.

**Client details:**
 - OS: macOS
 - Browser chrome
 - Version 21
 - Device: MacBook Pro 2019

## Server details

**Database:**
MariaDB 10.2

**PHP version:**
7.4.14

**Nextcloud version:**
20.0.7.1

**Version of the richdocuments app**
3.7.14

**Version of Collabora Online**
6.4.6.2 (via Docker container)

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```

[...]

PR #1396, opened 2021-03-02
Avoid determination of document editor in per-user-encryption setups

marioklump:
* Resolves: #1379
* Target version: master 

### Summary
When saving a document with enabled per-user encryption keys, an exception `Private Key missing for user: please try to log-out and log-in again` is thrown.

This is caused by the determination of the user who is editing the document to be able to trace the document's changes (https://github.com/nextcloud/richdocuments/blob/master/lib/Controller/WopiController.php#L407). Setting a specific user causes the usage of the user-specific encryption key instead of using the incognito mode (like it's done in `getFile`: https://github.com/nextcloud/richdocuments/blob/master/lib/Controller/WopiController.php#L349).

### Checklist

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -22,17 +22,20 @@
 namespace OCA\Richdocuments\Controller;
 
 use OC\Files\View;
-use OCA\Richdocuments\Db\Wopi;
 use OCA\Richdocuments\AppConfig;
+use OCA\Richdocuments\Db\Wopi;
 use OCA\Richdocuments\Db\WopiMapper;
+use OCA\Richdocuments\Helper;
 use OCA\Richdocuments\Service\UserScopeService;
 use OCA\Richdocuments\TemplateManager;
 use OCA\Richdocuments\TokenManager;
-use OCA\Richdocuments\Helper;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\JSONResponse;
+use OCP\AppFramework\Http\StreamResponse;
+use OCP\AppFramework\QueryException;
+use OCP\Encryption\IManager as IEncryptionManager;
 use OCP\Files\File;
 use OCP\Files\Folder;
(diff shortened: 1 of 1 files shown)


marioklump:
@juliushaertl Do you know how to solve the failing tests? I'm using `OCA\Encryption\Util` to determine if the master-key is enabled, but it's not part of `christophwurst/nextcloud`. There seems not to be an `isMasterKeyEnabled()` method as part of  Nextcloud's public API.

juliushaertl:
> I'm using OCA\Encryption\Util to determine if the master-key is enabled, but it's not part of christophwurst/nextcloud. There seems not to be an isMasterKeyEnabled() method as part of Nextcloud's public API.

Yes, the server-side encryption is built in a modular way where the default encryption module (the encryption app that is shipped) is taking care of the implementation in regards to user/master key encryption, so that part is only in the app and not in the server encryption API. That is of course not ideal from an integrating apps perspective since we cannot be sure if the app is enabled or not. Due to that the app is disabled in the tests the autoloader is not able to inject the Util instance into the WopiController constructor. Probably the most suitable way would be to just avoid using dependency injection for the master key check and do something like this for the check:

```
$isMasterKeyEnabled = false;
try {
	$util = \OC::$server->query(\OCA\Encryption\Util::class);
	$isMasterKeyEnabled = $util->isMasterKeyEnabled();
} catch(\OCP\AppFramework\QueryException $e) {
	// No encryption module enabled
}
```

lib/Controller/WopiController.php
```diff
@@ -714,4 +729,19 @@ public function getTemplate($fileId, $access_token) {
 		}
 	}
 
+	/**
+	 * Check if the encryption module uses a master key.
+	 *
+	 * @return bool
+	 */
+	private function isMasterKeyEnabled()
+	{
```
juliushaertl, line 738:
```suggestion
	/**
	 * Check if the encryption module uses a master key.
	 */
	private function isMasterKeyEnabled(): bool {
```

Very small nitpick on the coding style and we can actually type hint on the master branch (and for the stable 4.x releases)

Otherwise :+1: from my side.

juliushaertl (approved):

Issue #1395, opened 2021-03-01
Could not establish connection to the Collabora Online server when URL has a trailing slash /


nebulade:
**Describe the bug**
When configuring the collabora office connector, it asks for the collabora server URL (`URL (and Port) of Collabora Online-server`)

If this URL due to copy and paste for example contains a trailing slash, then the error "Could not establish connection to the Collabora Online server." is shown, while it works fine after removing the slash.

**Expected behavior**
The code should be smart enough to remove trailing slashes automatically.

## Server details

**Operating system**: Cloudron

**Web server:** Apache

**Database:** Postgres

**PHP version:** 7.4

**Nextcloud version:** 20.0.8

**Version of the richdocuments app** 3.7.15

**Version of Collabora Online** 6.4-11

juliushaertl:
A fix is proposed in https://github.com/nextcloud/richdocuments/pull/1412

Testing is very welcome :)

PR #1412, opened 2021-03-08
Trim duplicate trailing slashes in wopi url

juliushaertl:
Fixes https://github.com/nextcloud/richdocuments/issues/1395

### DIFF
--- a/lib/Service/CapabilitiesService.php
+++ b/lib/Service/CapabilitiesService.php
@@ -89,7 +89,7 @@ class CapabilitiesService {
 		if ($remoteHost === '') {
 			return [];
 		}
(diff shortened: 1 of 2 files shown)


rullzer (approved):

eneiluj (approved):
Code seems fine but I couldn't reproduce the issue. It works fine with a trailing slash on my side. So I didn't try the fix.

juliushaertl:
> Code seems fine but I couldn't reproduce the issue. It works fine with a trailing slash on my side. So I didn't try the fix.

Yep, depends on how the reverse proxy is forwarding the requests to Collabora, so it might still work on some setups.

PR #1448, opened 2021-03-25
Migrate to @nextcloud/webpack-vue-config

skjnldsv:
- Update eslint and babel configs
- Migrate to proper stanhdard template
- Fix eslint


skjnldsv:
@juliushaertl should be mostly done, I left the commits split, do you want to take over and review? :)

juliushaertl:
Thanks for getting this started :) There still seems to be an issue with the resulting bundles, but I'll look into that:

image

skjnldsv:
> Thanks for getting this started :) There still seems to be an issue with the resulting bundles, but I'll look into that:

I released the first beta for @nextcloud/babel-config, can you try?

juliushaertl:
Works like a charm :heart_eyes:

juliushaertl (approved):

.eslintrc.js
```diff
@@ -1,86 +1,24 @@
...
-		parser: 'babel-eslint',
-		ecmaVersion: 6
+	settings: {
+		'import/resolver': {
+			node: {
+				paths: ['src'],
+				extensions: ['.js', '.vue'],
+			},
+		},
 	},
```
skjnldsv, line 20: This part was already included in the config 
https://github.com/nextcloud/eslint-config/blob/f36fb05ab5d3a42618fb6542524f89b6883e29d1/index.js#L24

I added it because of the typescript, but maybe you don't lint it?

.eslintrc.js
```diff
@@ -1,86 +1,24 @@
...
-	parserOptions: {
-		parser: 'babel-eslint',
-		ecmaVersion: 6
+	settings: {
+		'import/resolver': {
+			node: {
+				paths: ['src'],
+				extensions: ['.js', '.vue'],
+			},
+		},
```
skjnldsv, line 19: This part was already included in the config 
https://github.com/nextcloud/eslint-config/blob/f36fb05ab5d3a42618fb6542524f89b6883e29d1/index.js#L24

I added it because of the typescript, but maybe you don't lint it?

juliushaertl: Yeah those are not linted at the moment as I had some issues with regular eslint rules not working properly back then with the type identifiers, probably worth to have another look but let's keep that out of that PR for now. I've pushed another commit to get rid of the import resolver config for now then.

PR #1459, opened 2021-04-02
Reload file name for share feature after Save As

eszkadev:
There was a bug that after save as the sidebar was showing
the old file when we used share feature.


juliushaertl (approved):

juliushaertl:
CI failure unrelated as branch is based on 3.7.14 instead of master the older drone config is picked up.

PR #1502, opened 2021-04-22
Do not use libxml_disable_entity_loader on PHP 8 or later

juliushaertl:
Avoid deprecation logs on PHP 8

https://php.watch/versions/8.0/libxml_disable_entity_loader-deprecation

> In PHP 8.0 and later, PHP uses libxml versions from 2.9.0, which disabled XXE by default. libxml_disable_entity_loader() is now deprecated.

### DIFF
--- a/lib/WOPI/Parser.php
+++ b/lib/WOPI/Parser.php
@@ -39,9 +39,14 @@ class Parser {
 	 */
 	public function getUrlSrc($mimetype) {
 		$discovery = $this->discoveryManager->get();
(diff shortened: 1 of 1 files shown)


juliushaertl:
oci failure seems to be an issue with the github action that installs the php module, unrelated to this PR.

ChristophWurst (approved):

KimHEF:
Hey Thanks for your Solution.
I am not very familiar with PHP.... so i have to deactivate this Module?
Could you explain how?

Kind Regards.

PR #1501, opened 2021-04-22
Add federation/direct editing tests

juliushaertl:
Add test cases for federated editing and direct editing and extend the wopi cases with some checks on the PutFile and GetFile operations.

PR #1512, opened 2021-04-26
Enhanced federated editing

juliushaertl:
### Pending

- [x] Based on https://github.com/nextcloud/richdocuments/pull/1501
- [x] Document new API endpoint

## General concept

This is a rough summary on how the federated editing worked also before this PR:

Collabra Online offers support for collaborative editing of files that are shared through federated
sharing. The file will always be opened on the file owners instance (Source) and other users will
connect to that through their Nextcloud instance (Initiator).

In a federated scenario both Nextcloud servers need to add each other as trusted servers or in a
global scale environment a list of nodes can be added through the `gs.trustedHosts` setting in the
config.php file.

### Additions through this PR

When a Initiator opens a file that is located on an incoming federated share, a check will be
performed if the share owners instance supports federated editing. If that is the case a Initiator
token will be created, and the user will be redirected to the Source instance to open the file.

The source instance will then fetch the user and file details, create a WOPI token for the remote
user with those details and open the document with that.

This allows different new cases in federated editing as now not only the user information is preserved but the additional token types allow:
- Local reshares of incoming federated shares
- Reshares as link of incoming federated shares
- Direct editing for share links to allow users to open a share link on any instance with their user information through a direct editing link

### Wording

The following wordings are used related to federated editing:

- Source: The instance the original file is located at that was initially shared by the file owner
- Initiator (WOPI): The instance that opens the file
- Initiator (User): The instance the file open request was issued by the user
  - Normally this is the same as Initiator(WOPI), except when a public share link is opened by a user on another instance than the Initiator(WOPI)
    - E.g. Instance A requests a direct share link for a public share token on instance B that originates to a federated share from Instance C

### Implementation details

- Files will always be opened with the Collabora server that is provided by the Source instance
- When a initiator attempts to open a file
  - Initiator tokens will be created with the initiator details (user info, wopi token on the initator instance)
  - The user is being redirected to the source instance providing it with the initator token
  - The source instance then will request the information from the initiator tokens on the initiator instances and open the file in the context of the initiator
- Direct editing for share links
  - Allow users to open a share link on any instance with their user information through a direct editing link
  - The OCS API allows the user to request a direct editing link for the share link on Initiator(WOPI)
- Preserve editor information across federation
- Introduce token types
  - TOKEN_TYPE_USER
    - Regular WOPI token for opening a local file as a user on SOURCE 
  - TOKEN_TYPE_GUEST
    - Regular WOPI token for opening a public share link as a guest on SOURCE
  - TOKEN_TYPE_REMOTE_USER
    - Remote WOPI token to open a file initiated by a regular user of INITIATOR on SOURCE
    - remoteServer/remoteServerToken contain the reference to the INITIATOR instance
  - TOKEN_TYPE_REMOTE_GUEST
    - Remote WOPI token to open a file initiated by a guest of INITIATOR(WOPI) on SOURCE
  - TOKEN_TYPE_INITIATOR
    - serverHost contains the source instance host
    - if the Initiator(WOPI) is Initiator(User) the editorUid is used to provide the user info
    - if the Initiator(User) is located on a different instace
      - remoteServer may contain the INITIATOR(USER) host
      - remoteServerToken may contain the INITIATOR(USER) user token
      - Those can be used to obtain the editorUid from the INITIATOR(USER) instance
- Integration tests should cover all the different scenarios that may occur with federated sharing and local sharing in general

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -18,7 +18,7 @@ jobs:
       # do not stop on another job's failure
       fail-fast: false
       matrix:
-        php-versions: ['7.4']
+        php-versions: ['8.0']
         databases: ['sqlite']
         server-versions: ['master']
         scenarios: ['wopi', 'direct', 'federation']
@@ -84,7 +84,7 @@ jobs:
       # do not stop on another job's failure
       fail-fast: false
       matrix:
-        php-versions: ['7.4']
+        php-versions: ['8.0']
         databases: ['mysql']
         server-versions: ['master']
         scenarios: ['wopi', 'direct', 'federation']
@@ -157,7 +157,7 @@ jobs:
       # do not stop on another job's failure
       fail-fast: false
       matrix:
-        php-versions: ['7.4']
+        php-versions: ['8.0']
         databases: ['pgsql']
         server-versions: ['master']
         scenarios: ['wopi', 'direct', 'federation']
@@ -232,7 +232,7 @@ jobs:
       # do not stop on another job's failure
       fail-fast: false
       matrix:
-        php-versions: ['7.4']
+        php-versions: ['8.0']
         databases: ['oci']
         server-versions: ['master']
         scenarios: ['wopi', 'direct', 'federation']
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -18,11 +18,12 @@ jobs:
       # do not stop on another job's failure
       fail-fast: false
       matrix:
-        php-versions: ['7.3', '7.4']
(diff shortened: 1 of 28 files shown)


juliushaertl:
For now I moved the integration tests to run on PHP8 only as there seems to be an issue with the built-in php server when doing recursive requests which apparently seems to be fixed on PHP8.0.3+ only (https://github.com/php/php-src/commit/5cb25a2d32df8073df7c191dd6c7694b4495af62)

rullzer (approved):
Looks good. I didn't test I must admit.

PR #1513, opened 2021-04-27
Don't cache empty capabilities

### LINKED ISSUE #1441: Trying to open file when Collabora Online not running "forgets" it was set up
**To Reproduce**
- Open a file, with Collabora Online running (to confirm all is fine at the beginning),
- Stop `loolwsd` service,
- Try opening a file again, it'll spin, then time out, feel free to go back right away, no need to wait for the timeout,
- Start `loolwsd` service again,
- Try opening a file again.

**Expected & buggy behavior**
After the last step, the file should open.

Instead there is an error: `Collabora Online is not setup yet. Click here...`
The admin has to get into the Settings, and click Save on the URL, then it works again.

Built-in CODE Server (richdocumentscode) triggers a similar bug after upgrade, the error message is different (a generic failure message), but opening files fails, and clicking Save in Settings fixes it.

## Server details

**Nextcloud version:**
20.0.8
**Version of the richdocuments app**
3.7.17
**Version of Collabora Online**
6.4.7

eszkadev:
Fixes #1441

When caching is configured and request to fetch the
capabilities fails - don't cache the empty result.

This way we will avoid persisting error:
"Collabora Online is not setup yet".

To test bug:
1. use Nextcloud with caching & richdocuments
2. restart apache and stop Collabora Online server
3. open Nextcloud and try to open any document
   you will get "Collabora Online is not setup yet" error
4. start Collabora Online server
5. refresh page and try to open file again

before: error persisted until admin changed richdocuments configuration
after: file is opened with Collabora Online

### DIFF
--- a/lib/Service/CapabilitiesService.php
+++ b/lib/Service/CapabilitiesService.php
@@ -113,6 +113,10 @@ class CapabilitiesService {
 		}
 
 		$this->capabilities = $capabilities;
-		$this->cache->set('capabilities', $capabilities, 3600);
+		$ttl = 3600;
+		if (count($capabilities) === 0)
+			$ttl = 60;
+
+		$this->cache->set('capabilities', $capabilities, $ttl);
 	}
 }


lib/Service/CapabilitiesService.php
```diff
@@ -113,6 +113,7 @@ public function refetch() {
 		}
 
 		$this->capabilities = $capabilities;
-		$this->cache->set('capabilities', $capabilities, 3600);
+		if (count($capabilities) !== 0)
+			$this->cache->set('capabilities', $capabilities, 3600);
```
juliushaertl, line 117: I'm a bit concerned here about that this will cause every request to fetch the capabilities again so in case there is some issue where the request times out or takes very long this would add to all the requests where nextcloud tries to obtain the capabilities. How about just storing it with a much lower ttl in the cache in case of the exception above?

eszkadev: I think it will be good compromise, what ttl do you propose? 60 s ?

juliushaertl: Yep, sounds sane to me :+1:

thebearon: @juliushaertl What is the purpose of refetching the capabilities at this point? I'd assume, if there are connectivity problems, a timeout would be the expected behavior, and the already cached capabilities are suitable for that. Is that not the case?

The exception I can think of is if the server is being reconfigured, which is probably done very rarely,

eszkadev: @thebearon when capabilities are needed first there is a cache used (see `getCapabilities`), but when we don't have anything in the cache then `refetch` function is executed.
Exception above my code is fired for example when Collabora Online server is not started at all. When server is re/starting then from my experience we don't receive any exception here, but normal response (after some longer time when capabilities enpoint is ready).

thebearon: > @thebearon when capabilities are needed first there is a cache used (see `getCapabilities`), but when we don't have anything in the cache then `refetch` function is executed.

@eszkadev Right, but in my repro steps I opened online just a minute before, thus the capabilities must've gotten cached. Then I turn online off to simulate a connectivity problem, I try opening a file again, and somewhere around this point it refetches the empty capabilities, because when I try once again afterwards (with online available again), the cached "good" capabilities are gone. This refetch seems incorrect to me.

The fix of lowering the lifetime of cached empty capabilities is good, just an additional piece of fix is needed in my opinion, unless there's a sound reason why that refetch happens (obviously I don't know much about what's happening there ;)).

juliushaertl: > @eszkadev Right, but in my repro steps I opened online just a minute before, thus the capabilities must've gotten cached. Then I turn online off to simulate a connectivity problem, I try opening a file again, and somewhere around this point it refetches the empty capabilities, because when I try once again afterwards (with online available again), the cached "good" capabilities are gone. This refetch seems incorrect to me.

Do you have APCu/Redis caching setup on your instance? Otherwise it would fetch them always anyways

juliushaertl (approved):

lib/Service/CapabilitiesService.php
eszkadev: @thebearon For me it works (it remembers capabilities and opens editor, but receives 404 because Collabora is off), no refetching happens and no error. I cannot reproduce your steps

thebearon: > Do you have APCu/Redis caching setup on your instance? Otherwise it would fetch them always anyways

@juliushaertl I haven't set up anything special on my instance, so I presume I don't have those. Yet I'm still having this issue.

juliushaertl: Can you quickly check your nextclouds config.php for entries mentioning `memcache.local` and `memcache.distributed`?

thebearon: @juliushaertl I have neither in my config.php.

juliushaertl: OK, strange as then the capabilities should be refetched on every page load.

Just to make sure we don't miss something obvious here, do you reproduction steps include to reload the page after stopping/starting the online server? If not, that might be something for the additional fix mentioned via email for checking each time before opening a document.

thebearon: The steps do include reloading the page after stopping the server, because you get stuck on a loading screen (unless you wait that out), but not after restarting the service (I have uploaded a screencast to the issue). Indeed, it seems the additional fix could help with this.

On the other hand, I still wonder if it makes sense that the empty capabilities overwrite the cached non-empty capabilities. But I guess, with the other fix this won't be a problem.

eszkadev: @thebearon did you use browser refresh button? I think I saw in one of your videos you use 'Files' icon from the top menubar which isn't the same in result as browser refresh I think. (or am I wrong?)

eszkadev: @juliushaertl can you implement the mentioned fix on the client side? I did short research and I wasn't able to get it working. It seems for me `OC.getCapabilities()` uses only some field generated in the html, I don't know how to fetch capabilities on demand.

mmeeks:
I think we need to re-think this. It seems to me that checking for or caching capabilities for every Nextcloud request is a complete waste of CPU and/or resources generally. Also - if we are not setup properly - we already fail nicely on click on a file with a popup anyway - so ... I would suggest we:

* if Collabora Online plugin is installed - we should assume that it is there & working & up-to-date (I don't think we added new capabilities any time recently), right up until the click on the document to edit it. Ergo we don't need to check / fetch any capabilities during most requests.
* As/when we are serving up our iframe on-clicking a document etc. - then we can check/warn of failure (it seems we do that now already =)
* I suspect the useful thing is for the early/regular check is the Hub use-case where we want this installed by default - but not enabled unless the backend is "really there" - we assume that if we are in this "app-image" mode, and the app-image is not present on the file-system then we  don't show the UI elements etc. (?) - or is that even a thing ? Presumably we could do this by checking for the local app-image, and/or moving the delay/uncertainty into the config/setup piece and out of each request (?) - ie. if it worked once during setup - we assume by default it is working ever after ?

Or did I miss some use-case? what do we really use the capabilities for? can we be smarter about that ? or are we too tied to a synchronous check at the time of request?

lib/Service/CapabilitiesService.php
thebearon: @eszkadev i usually click on the Files icon, or sometimes the folder icon to get out of the stalled loading screen.

kendy:
@eszkadev says we should merge this, but he does not have the rights - so let me push the button :-)

In the meantime, @eszkadev will have a look if he can implement the @mmeeks ' suggestion in a separate PR.

@juliushaertl - can you please release a new version of richdocuments with this fix in?  Thank you in advance!

juliushaertl:
> @juliushaertl can you implement the mentioned fix on the client side? I did short research and I wasn't able to get it working. It seems for me OC.getCapabilities() uses only some field generated in the html, I don't know how to fetch capabilities on demand.

Yes, I'm not aware of any place where we refetch it but it is an OCS API so you can just send a GET request like this:

```
$.get(OC.linkToOCS('cloud') + '/capabilities?format=json').then(e => console.log(e.ocs.data.capabilities.richdocuments))
```

> @juliushaertl - can you please release a new version of richdocuments with this fix in? Thank you in advance!

Yes, I'll prepare a new release then with that.

PR #1528, opened 2021-05-05
Emit WOPI postmessages through the nextcloud event bus

juliushaertl:
Emit all post messages we receive from Collabora on the Nextcloud event bus.


MorrisJobke (approved):
Code looks good 👍

PR #1544, opened 2021-05-10
Retry to check if Collabora is configured

eszkadev:
If Collabora wasn't initially configured, retry to check the capabilities just before request to open the file before we will show the error.


eszkadev:
@juliushaertl follow up for #1513

juliushaertl (approved):
Great, thanks a lot :+1:

PR #1546, opened 2021-05-14
Fix wrong parameters set when creating a file from a template id

juliushaertl:
First commit will fail tests, the second fixes the issue with is a regression from https://github.com/nextcloud/richdocuments/commit/ac959b19ec62ae6def4d18690f0f4b96bdad3b75 where dropped `$isRemoteToken` parameter was not removed in the call.


skjnldsv (approved):

joergmschulz:
see https://github.com/nextcloud/richdocuments/issues/1563 - not all seems to be fixed

PR #1640, opened 2021-06-29
Allow to limit wopi requests to specific source hosts

juliushaertl:
This PR adds a dedicated allow list that if specified will limit the hosts that can perform requests against the WOPI endpoints. IP parsing and subnet matching was taken from https://github.com/nextcloud/limit_login_to_ip/blob/master/lib/LoginHookListener.php


LukasReschke (approved):
🚀

abdulmge (commented):
~~

PR #1663, opened 2021-07-13
Throttle on invalid share tokens

LukasReschke:


juliushaertl:
Updated the baseline since the ShareNotFound is not found as it extends from HintException which is not in OCP.

juliushaertl (approved):

juliushaertl:
MySQL test failures unrelated: https://github.com/nextcloud/richdocuments/issues/1665

PR #1707, opened 2021-08-11
Show avatar list if hide download is enabled

juliushaertl:
There is no .header-right div when hide download is enabled so we need to add it manually in order to properly show the avatar list in the header then. Before it was not visible at all.

image

### DIFF
--- a/css/viewer.scss
+++ b/css/viewer.scss
@@ -81,6 +81,10 @@
 	position: relative;
 	right: -15px;
 
+	.header-right-richdocuments & {
+		right: 0;
+	}
+
 	& > .richdocuments-avatar {
 		float: right;
 		border-radius: 50%;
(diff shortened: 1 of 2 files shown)


artonge (commented):
Can't we fix the template instead of hot fixing the DOM ?

juliushaertl:
We can but I'd still like to get this in as this can be released out of bounds as an intermediate fix :)

artonge (approved):

artonge:
> We can but I'd still like to get this in as this can be released out of bounds of the server as an intermediate fix :)

I am not sure that I understand your point, but I trust it is a good one :)

juliushaertl:
> I am not sure that I understand your point, but I trust it is a good one :)

We can do a richdocuments release to the app store with the fix independent of the actual server change that might not be backported or just come later when the next maintenance release is scheduled ;)

artonge:
Oh, I did not saw that this issue was on richdocuments...

PR #1710, opened 2021-08-12
Another attempt to fix height setting for Safari issues on iOS

### LINKED ISSUE #1608: Bit of file list visible after resizing window
- Open a document, and increase the height of the window a bit.

=> There is a bit of gap between Collabora Online and the bottom edge of the browser.

Screencast:
https://user-images.githubusercontent.com/19438782/121974252-4a898800-cd7f-11eb-9d8c-1a73029b3907.mp4

In Safari on macOS this gap is shown right away after opening a file:
file_list_shown_Safari

Nextcloud 20.0.10
richdocuments 3.8.2

Observed with Chrome and Firefox on Linux, and with Safari on macOS.

Possibly a regression from e927ce1bdbb41887fc3149431760f95631c98207? (at least in non-Safari browsers)
CC @eszkadev, @juliushaertl.

juliushaertl:
Another attempt to improve frame handling on iOS devices where the height needs to be set manually as 100vh css values do not work as expected.

Also fixes #1608 where the header value was wrongly set to 60px instead of 50px

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -17,7 +17,10 @@ const PostMessages = new PostMessageService({
 // Workaround for Safari to resize the iframe to the proper height
 // as 100vh is not the proper viewport height there
 const handleResize = () => {
-	document.getElementById('richdocumentsframe').style.maxHeight = (window.innerHeight - 60) + 'px'
+	const frame = document.getElementById('richdocumentsframe')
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
I can confirm it fixes #1608  for me

PR #1711, opened 2021-08-12
Allow guests to request a direct token for share links

juliushaertl:
Allow direct editing for share tokens also for guest users

- [x] Create a direct token without a user
- [x] Handle loading and share token validation
- [x] Figure out a way to set the guest name 
- [x] move guest name setting to a dedicated api endpoint where it can be set given the wopi file id and secret to avoid full page reload which does not work for direct tokens and saves generating an unused wopi token

### DIFF
--- a/lib/Controller/OCSController.php
+++ b/lib/Controller/OCSController.php
@@ -23,9 +23,9 @@
  */
 namespace OCA\Richdocuments\Controller;
 
+use Exception;
+use GuzzleHttp\Exception\BadResponseException;
 use OCA\Richdocuments\Db\DirectMapper;
-use OCA\Richdocuments\Db\Wopi;
-use OCA\Richdocuments\Helper;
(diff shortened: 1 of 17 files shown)


lib/Controller/OCSController.php
```diff
@@ -148,30 +153,47 @@ public function createPublic(
...
+						'shareToken' => $shareToken,
+						'path' => $path,
+						'password' => $password
+					],
+					'timeout' => 30
+				]);
+			} catch (\Exception $e) {
+				$response = new DataResponse([], HTTP::STATUS_FORBIDDEN);
+				$response->throttle();
+				return $response;
```
artonge, line 170: Shouldn't the status be internal server error instead of forbidden?
Also, would logging the exception be useful? Same for the other exception handlers.

artonge (approved):

PR #1728, opened 2021-08-26
Prevent not-needed early loading of the richdocuments app

juliushaertl:
There is no need to register the filesystem/dav types for the app so we can avoid loading it too early

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -10,8 +10,6 @@ You can also edit your documents off-line with the Collabora Office app from the
 	<licence>agpl</licence>
 	<author>Collabora Productivity based on work of Frank Karlitschek, Victor Dubiniuk</author>
 	<types>
-		<filesystem/>
-		<dav/>
 		<prevent_group_restriction/>
 	</types>
 	<documentation>


artonge:
Can't find information about this node here: https://docs.nextcloud.com/server/latest/developer_manual/app_development/info.html?highlight=info%20xml
Can you give me more context ? :)

juliushaertl:
Sure, the types are used to load groups of certain apps earlier than others.

E.g. authentication apps should be available as early as possible:
https://github.com/nextcloud/server/blob/b78f3a57d1d05fb5f81ec8361139b64a3e54d7b0/lib/base.php#L979

filesystem apps are loaded before all others with https://github.com/nextcloud/server/blob/b78f3a57d1d05fb5f81ec8361139b64a3e54d7b0/lib/base.php#L997 so their application bootstrap is ran early as well.

There was no concrete error case in the Collabora integration but you might want to have a look at the description in https://github.com/ONLYOFFICE/onlyoffice-nextcloud/pull/511#issue-718653808 for a bug caused by the too early loading in combination with the guests app https://github.com/ONLYOFFICE/onlyoffice-nextcloud/pull/511#issue-718653808

artonge (approved):
Ok, thanks, totally reversed the way I first understood it :)
Makes sense

PR #1750, opened 2021-09-10
Register middleware as a service for ancient versions

juliushaertl:
This ensures that we remain compatibility with older Nextcloud versions where registered middlewares were not resolved through the DI container. Fixes the app loading for times before https://github.com/nextcloud/server/pull/17569 was introduced.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -77,6 +77,14 @@ class Application extends App {
 		} catch (QueryException $e) {
 		}
 
+		$server = $this->getContainer()->getServer();
+		$this->getContainer()->registerService(WOPIMiddleware::class, function () use ($server) {
+			return new WOPIMiddleware(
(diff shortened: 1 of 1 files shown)


eneiluj:
@juliushaertl I'm not sure I get what this does.
So, with older Nextcloud versions:
``` php
$this->getContainer()->registerMiddleWare(WOPIMiddleware::class);
```
was ineffective because then `WOPIMiddleware` could not be obtained with something like
`$container->query(WOPIMiddleware::class)`?

juliushaertl:
> was ineffective because then WOPIMiddleware could not be obtained with something like
$container->query(WOPIMiddleware::class)

Yes, the previous code would just try to access the service named like that and not use the automatic resolution of classes in the app namespace as part of the `lib/` directory:

image

eneiluj (approved):
Thanks for the explanation.

PR #1781, opened 2021-09-26
richdocumentsframe use 100% of mobile screen height

Andreas-Kainz:
Signed-off-by: Andreas-Kainz <kainz.a@gmail.com>

* Resolves: #1608
* Target version: master 

### Summary
richdocumentframe use 50px less than screen height, which cause ain issue on mobile.

### Before
before

### After
after

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -19,7 +19,7 @@ const PostMessages = new PostMessageService({
 const handleResize = () => {
 	const frame = document.getElementById('richdocumentsframe')
 	if (frame) {
-		frame.style.maxHeight = (document.documentElement.clientHeight - 50) + 'px'
+		frame.style.maxHeight = (document.documentElement.clientHeight) + 'px'
 	}
 }
 window.addEventListener('resize', handleResize)


Andreas-Kainz:
@juliushaertl, @eszkadev, @pedropintosilva I only know that I still have the sizing issue on mobile (all browsers) at https://staging.eu.collaboraonline.com server. Maybe the sizing issue is already fixed. 

Please let me know.

juliushaertl:
The 50px offset was needed for conditions where the height could not be properly determined through css on larger iOS devices, though I haven't considered that we actually hide the Nextcloud toolbar on smaller ones.

It is determined by the viewport width https://github.com/nextcloud/richdocuments/blob/cef30f5ad01c9c0e2d06e272801064f709ac00b7/css/mobile.css#L1 so it would make sense to only use those 50px if the header is visible then.

Andreas-Kainz:
If I test try.nextcloud.com or the colabora staging and demo nextcloud account, everywhere I have this issue on my phone. If I remove the `max-height: 752px` everything work fine. I'm no expert, but I reviewed different machines and everywhere it work within the css rule.

Andreas-Kainz:
can the MR get to master?

Issue #1505, opened 2021-04-23
Document templates rendered two times.


top-logic:
**Describe the bug**
Template selector shows any template two times.
Doesn't matter if the templates comes from global setup or from personal settings or both.

**To Reproduce**
1. As NC-Admin upload a new template under Collabora Online Administration or just select a template folder in personal settings.
2. Click on "Create new Document" according to the uploaded template.

**Expected behavior**
Templates should rendered just one time.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/70580580/115844779-7a23b000-a420-11eb-8be2-33c2a2664ce7.png)

**Client details:**
 - OS: Linux Mint 20.1
 - Browser Firefox, Chrome, Chromium
 - Version 87.0, 90.0.4430.72, 89.0.4389.128
 - Device: Desktop

## Server details

**Operating system**: Ubuntu 20.04.2 LTS

**Web server:** Apache

**Database:** Mysql 8

**PHP version:** 7.4

**Nextcloud version:** 21.0.1

**Version of the richdocuments app** 4.0.4

**Version of Collabora Online** Docker image 6.4.8

Issue #1793, opened 2021-10-05
Drop user template directory


juliushaertl:
Since there is a user wide templates folder in Nextcloud now we should drop the user option to set a template and only make use of the core one.

joergmschulz:
if, finally, this results in a working setup, this would be good. See #1563

Issue #1563, opened 2021-05-18
documents created from templates are broken and cannot be saved


joergmschulz:
odt documents created from template cannot be saved 
---

(edited to reflect current status)

Environment: 
NC 22, any collabora server, any richdocuments 4.1.0 and more recent

How to reproduce the issue
---

1) create a document from a template, using name1.odt 
2) a document with this name is saved.
3) edit this document
4) try to save it. Saving will fail.
5) save it using another name using name2.odt. 
6) close richdocuments.
7) open and edit name1.odt 
8) save name1.odt, saving will fail.
9) open and edit name2.odt 
10) save name2.odt, saving will succeed.
11) open a local copy of name1.odt with libreoffice. It will be opened unnamed and can only be saved-as a new dockument. 
12) issue the command > file name1.odt: The result will be: ` OpenDocument Text Template`
13) unzip the file name1.odt
mimetype: `application/vnd.oasis.opendocument.text-template`

--> all documents created from templates are broken, the others work.

 
Here the log file of collabora; 
```

ERR  #29: Socket write returned -1 (EPIPE: Broken pipe)| ./net/Socket.hpp:1172
libreoffice-master    | wsd-00008-00035 2021-05-18 20:12:26.741516 [ websrv_poll ] ERR  #29: Socket write returned -1 (EPIPE: Broken pipe)| ./net/Socket.hpp:1172
libreoffice-master    | wsd-00008-00038 2021-05-18 20:12:41.934188 [ docbroker_001 ] ERR  Cannot store docKey [/index.php/apps/richdocuments/wopi/files/1342285_oc2577a033c1] as .uno:Save has failed in LOK.| wsd/DocumentBroker.cpp:981
libreoffice-master    | wsd-00008-00038 2021-05-18 20:12:41.934268 [ docbroker_001 ] ERR  error: cmd=storage kind=savefailed| ./common/Session.hpp:125 
```

Update
---
the last known working configuration is NC 20.0.9 and richdocuments 3.7.19

_Originally posted by @joergmschulz in https://github.com/nextcloud/richdocuments/issues/1500#issuecomment-843528961_

thebearon:
@joergmschulz please update your CODE, that will get rid of the JSON parse error. The other issue might be related to what #1559 is supposed to fix.

joergmschulz:
probably the same as https://github.com/nextcloud/richdocuments/issues/1541

joergmschulz:
[...]

PR #1794, opened 2021-10-05
Do not load template directories twice

juliushaertl:
Fixes https://github.com/nextcloud/richdocuments/issues/1505

Otherwise in cases where the new nextcloud template directory was set to the same location as the collabora user setting the templates were rendered twice.

We should also get rid of the user setting but that is something tracked in https://github.com/nextcloud/richdocuments/issues/1793


joergmschulz:
does that mean the use of templates will be possible now? #1563 ?

eneiluj (approved):

PR #1811, opened 2021-10-15
Support Flat ODF files

juliusknorr:
Rebased version of #1342


juliushaertl:
:+1: Let's get this in

PR #1816, opened 2021-10-16
url-encode filename when constructing the iframe url for the viewer

lu-fennell:
This fixes the problem mentioned
[here](https://github.com/nextcloud/server/issues/24618): when trying to
open a document from the dashboard in a Folder with special characters
like `&` or `+`, the viewer shows an error.

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/src/view/Office.vue
+++ b/src/view/Office.vue
@@ -116,7 +116,7 @@ export default {
 	},
 	methods: {
 		async load() {
-			const documentUrl = getDocumentUrlForFile(this.filename, this.fileid) + '&path=' + this.filename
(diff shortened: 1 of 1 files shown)


lu-fennell:
I didn't find any instructions for contributing,  so I hope it's fine just to send the PR like this. 

I see the complaint about signing off the commit, but I'm assuming an actual project member should sign-off and not me? 

Please let me know if I need to do something else.

juliushaertl:
Btw. the ticket you linked seems to be unrelated, maybe there was a mistake with that?

PR #1820, opened 2021-10-21
UI rework

juliusknorr:
This PR adapts the file display of Collabora to the new full screen mode by switching from the legacy viewer.js to use the viewer app instead. In addition it addresses some default handling like passing over the classic ui mode and making the notebookbar configurable through app config.

Further small enhancements:
- Use light viewer mode
- Pass over more css variables to collabora
- Temporary avatar listing display overlaying Collabora, this can be dropped with https://github.com/CollaboraOnline/online/pull/3529
- Start using inital state for passing document data to the frontend code

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -11,6 +11,7 @@
 
 namespace OCA\Richdocuments\Controller;
 
+use OCA\Richdocuments\AppInfo\Application;
 use OCA\Richdocuments\Db\Wopi;
 use OCA\Richdocuments\Events\BeforeFederationRedirectEvent;
 use OCA\Richdocuments\Service\FederationService;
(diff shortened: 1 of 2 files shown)


eneiluj (commented):
* Use light viewer mode: :+1: Very nice. One small problem though: The viewer sidebar button and the Collabora sidebar button overlaps when Collabora's menu is hidden via the top-right arrow button. Not a big deal.

* Pass over more css variables to collabora: :+1: 

* Temporary avatar listing display overlaying Collabora: Very nice too. The avatars are not aligned with the Collabora top-right close button. Maybe we could make this button the same size than the avatars and the viewer's sidebar button.

* Start using inital state for passing document data to the frontend code: :+1: 

Is the classic Collabora menu used intentionally instead of the ribbon one?

juliushaertl:
> Is the classic Collabora menu used intentionally instead of the ribbon one?

Yes, that was discussed before with @jancborchardt

eneiluj (approved):

PR #1855, opened 2021-11-10
Expose initial state when using direct editing as well

juliushaertl:
Follow up to #1820 to also expose initial state values when using direct editing


luka-nextcloud (approved):
Looks good to me!

PR #1938, opened 2021-12-20
Move filetype styles to separate css file and load it with viewer

### LINKED ISSUE #1936: 5.0.1: Icons are still missing
grafik

juliushaertl:
Fixes #1936


eneiluj (commented):
LGTM but why is viewer.scss imported in files.js and not in viewer.js?

juliushaertl:
Right, viewer is actually misleading here as it was called like that before we had a viewer app itself :wink: Let me rename it then right away :+1:

luka-nextcloud (approved):

PR #1939, opened 2021-12-23
Add cypress tests

juliushaertl:
Implement simple cypress tests for opening and creating a new document for all supported file types.

To be extended over time, just so that we finally have the structure in place.

There is also a few things that should be moved to @nextcloud/cypress like file upload and folder creation.


juliushaertl:
Turns out, test failures expected due to https://github.com/nextcloud/server/pull/30385 and i was just not on the latest master locally

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,143 @@
...
+        containers: [1, 2, 3, 4]
+        php-versions: [ '7.4' ]
+        databases: [ 'sqlite' ]
+        server-versions: [ 'master' ]
+
+    name: runner ${{ matrix.containers }}
+
+    services:
+      collabora:
+        image: collabora/code:21.11.2.4.1
```
juliushaertl, line 28: We need to handle the welcome popup separately with the follow up release during testing.

cypress/e2e/open.spec.js
```diff
@@ -0,0 +1,122 @@
...
+				.should('be.visible')
+				.and('have.class', 'modal-mask')
+				.and('not.have.class', 'icon-loading')
+
+			cy.get('#collaboraframe').iframe().should('exist').as('collaboraframe')
+			cy.get('@collaboraframe').within(() => {
+				cy.get('#loleafletframe').iframe().should('exist').as('loleafletframe')
+			})
+
+			cy.get('@loleafletframe').find('#main-document-content').should('exist')
```
marcelklehr, line 94: Can we use the waitForCollabora collabora command here?

juliushaertl: Yep good catch. I replaced all occurrences with the command and cleaned up the unused commands that have been moved to the package already.

marcelklehr (approved):

PR #1959, opened 2022-01-04
Don't wait DOMContentLoaded to register the viewer handler

### LINKED ISSUE #1958: Viewer handler registration is done too late
It seems that
https://github.com/nextcloud/richdocuments/blob/368fd1e887c7688fded9d6883f49855cfebd0705/src/viewer.js#L31-L41
is done too late and is not effective with latest server master branch, viewer and richdocument master branches.

For example, in my dev setup, browser debug logs mention 4 registered handlers but when browsing the handlers array, 6 are actually added after the object has been printed:
riri
The failing ones are `richdocuments` and `pdf`.

I'll check if recent changes in server caused that.

As a temporary fix (or even definitive if it makes sense), we can register the handler without waiting for `DOMContentLoaded`

``` js
document.addEventListener('DOMContentLoaded', function(event) {
```

eneiluj:
fixes #1958

We can now rely on script loading dependencies. In fact, waiting for DOMContentLoaded was even too late and the registration was not effective.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -28,14 +28,12 @@ import { getCapabilities } from '@nextcloud/capabilities'
 
 const supportedMimes = getCapabilities().richdocuments.mimetypes
 
-document.addEventListener('DOMContentLoaded', function(event) {
-	if (OCA.Viewer) {
-		OCA.Viewer.registerHandler({
-			id: 'richdocuments',
-			group: null,
-			mimes: supportedMimes,
-			component: Office,
-			theme: 'light',
-		})
-	}
-})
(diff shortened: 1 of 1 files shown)


eneiluj:
@juliushaertl Oh I saw your comment in #1958 after making this PR. Same conclusion on our side with @mejo- :grin:.

mejo- (approved):

eneiluj:
The backport to stable23 is not mandatory as it works fine there but it removes this frontend warning:
`Please do NOT wait for the DOMContentLoaded before registering your viewer handler`

PR #2003, opened 2022-01-27
Blank timestamp is equivalent to null

Ashod:
The request->getHeader() function returns an empty string
even when the header in question doesn't exist. Here we
treat blank timestamps as if they weren't set, which translates
to skipping the timestamp mismatch and forcing the PUT operation.

Resolves the "overwrite" option in the conflict resolution dialog not working.


juliushaertl (approved):

juliushaertl:
Good catch 👍

PR #2007, opened 2022-01-31
Keep mime-type indices consecutive.

rotdrop:
Some JS parts of the code need the mime-type list as array. However,
when using array_filter() indices may become non-consecutive and
non-zero based. In this case json_encode/decode will not generate a JS
array but a JS object. This commit fixes this problem by wrapping the
potentially sparse array with a call to array_values().

Signed-off-by: Claus-Justus Heine <himself@claus-justus-heine.de>

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


juliushaertl (approved):

juliushaertl:
Good catch, thanks a lot for your contribution :)

juliushaertl:
Stable 4 apparently doesn't need that as it has no draw support.

Issue #2009, opened 2022-02-01
Issue on public page share links


juliushaertl:
<img width="1792" alt="Screenshot 2022-02-01 at 13 07 56" src="https://user-images.githubusercontent.com/3404133/151965891-296aa8cb-2e77-447c-8741-1bb7d30a6391.png">

PR #2014, opened 2022-02-01
Fix frame absolute position in public file share

eneiluj:
This one of the many ways to solve the style issue of #2009.


juliushaertl:
Awesome. Still wondering how we broke that, but fix makes sense, so LGTM 👍

juliushaertl (approved):

PR #2015, opened 2022-02-01
Avoid too specific psalm annotation

juliushaertl:
Make psalm on master happy again


juliushaertl:
Merging to get CI green as it is not a code change anyways

PR #2029, opened 2022-02-08
Avoid too specific psalm annotation

juliushaertl:
backport of #2015


juliushaertl:
Merging since adjusting CI branches only

PR #2035, opened 2022-02-11
Add alert when trying to insert cloud image into shared doc

NickWingate:
Signed-off-by: NickWingate <nick.wingate@collabora.com>

* Resolves: https://github.com/CollaboraOnline/online/issues/2859
* Target version: master 

### Summary
When trying to insert an image from nextcloud in a shared document the users gets an alert instead of an infinitely loading file prompt

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/view/FilesAppIntegration.js
+++ b/src/view/FilesAppIntegration.js
@@ -144,30 +144,31 @@ export default {
 	insertGraphic(insertFile) {
 		if (isPublic) {
 			console.error('[FilesAppIntegration] insertGraphic is not supported')
-		}
-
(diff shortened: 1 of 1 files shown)


juliushaertl:
Thanks for your pull request. We can actually make sure that the option is not even offered by COOL by making the following WOPI checkFileInfo response return false if the document opened is a share link.

juliushaertl:
This has been addressed now with https://github.com/nextcloud/richdocuments/pull/2175 thanks for the PR anyways :)

PR #2051, opened 2022-02-16
🐛 Fix CSP violation when Nextcloud server has so-called 'service root'

ldidry:
* Target version: master 

### Summary

When Nextcloud has a service root part, e.g. it is https://example.org/nextcloud, there was a CSP violation.

```
Refused to send form data to 'https://coolwsd.example.org/browser/739da71/cool.html?WOPISrc=https%3A%2F…gl&title=Foobar.ods&lang=fr&closebutton=1&revisionhistory=1' because it violates the following Content Security Policy directive: "form-action 'self'".
```

Similar to PR #1000

### Checklist

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -159,6 +159,7 @@ class DocumentController extends Controller {
 		$policy = new ContentSecurityPolicy();
 		$policy->addAllowedFrameDomain($wopiDomain);
 		$policy->allowInlineScript(true);
+		$policy->addAllowedFormActionDomain($wopiDomain);
 		$response->setContentSecurityPolicy($policy);
 
 		$featurePolicy = new FeaturePolicy();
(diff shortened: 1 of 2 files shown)


lib/Controller/DocumentController.php
```diff
@@ -159,6 +159,9 @@ private function setupPolicy($response) {
 		$policy = new ContentSecurityPolicy();
 		$policy->addAllowedFrameDomain($wopiDomain);
 		$policy->allowInlineScript(true);
+		if (method_exists($policy, 'addAllowedFormActionDomain')) {
```
tcitworld, line 162: `addAllowedFormActionDomain` exists since Nextcloud 17 so the check is not needed.

tcitworld (changes requested):

lib/Controller/DocumentController.php
ldidry: Should I remove it from `lib/AppInfo/Application.php` too, then?

tcitworld: Yes, you may.

ldidry: Done.

juliushaertl (approved):

tcitworld (approved):

PR #2053, opened 2022-02-16
Do not hide speadsheet statusbar by default

juliushaertl:
Otherwise users might miss how they can switch between different sheets of a spreadsheet file
Screenshot 2022-02-16 at 15 07 48


jancborchardt (approved):
Yup, totally! While for documents the statusbar can be hidden by default, for spreadsheets we need to keep showing it. :)

juliushaertl:
Yes, that actually was Safari on macOS. I can also dig a bit more into that tomorrow.

juliushaertl:
The richdocuments frame seems to be having the correct height, but seems to be somehow related to the flex layout and a height on the toolbar that is being set on the branding:
Screenshot 2022-02-18 at 12 26 08

PR #2054, opened 2022-02-16
Map es-419 to es-MX

juliushaertl:
Steps to reproduce:

- Set your users language to "Español (Latin America)" (Nextcloud language code es_419)
- Try to open a document

A follow up fix to handle language codes with numbers more gracefully will follow on COOL but this will make sure documents can be properly loaded already in the meantime.


eneiluj (approved):
:+1:
I had to rebase on master to get the CSP fix to make it work on my setup.
So I can also say this PR works on top of master too :grin:.

PR #2055, opened 2022-02-16
[stable3.8] Map es-419 to es-MX

juliushaertl:
Backport of #2054 

Steps to reproduce:

- Set your users language to "Español (Latin America)" (Nextcloud language code es_419)
- Try to open a document

A follow up fix to handle language codes with numbers more gracefully will follow on COOL but this will make sure documents can be properly loaded already in the meantime.

- [ ] Test against lowest supported Nextcloud version to double check

### DIFF
--- a/src/helpers/index.js
+++ b/src/helpers/index.js
@@ -30,6 +30,12 @@ const languageToBCP47 = () => {
 	if (language === 'de-DE') {
 		language = 'de'
 	}
+
+	// es-419 should be mapped as this is not considered a valid locale string in COOL
+	if (language === 'es-419') {
+		language = 'es-MX'
+	}
+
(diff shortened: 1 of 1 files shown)


juliushaertl:
Seems PHPUnit / php7.3-mysql-master (pull_request) Failing after 44s — php7.3-mysql-master is running against server master so failure is unrelated and expected, will open a pr to address this.

eneiluj:
Is the stable3.8 branch supposed to work fine with the latest Collabora server?

juliushaertl:
Yes, should work just fine.

Let's close this since it has been fixed upstream in https://github.com/LibreOffice/core/commit/1145909db26849350fb4f89392f17ddaeef69951 so our workaround is no longer required.

PR #2057, opened 2022-02-17
[stable3.8] Run tests against proper stable branch

juliushaertl:
Lets see how happy CI is with that


eneiluj:
:+1: Let's get this in.

PR #2069, opened 2022-02-23
Adapt branding css variables

juliushaertl:
Avoid passing font face and pass over theming cache buster instead of the url due to css variable filtering in place in online.


pedropintosilva:
Seems to work, thanks : )

----

I tested with default setup and then by setting a custom logo via settings
When custom logo is present:
- customLogo = http://localhost/nextcloud/index.php/apps/theming/image/logo?useSvg=1&v=3

When custom logo not present:
- customLogo = false

luka-nextcloud (approved):

PR #2093, opened 2022-03-10
Refactor building of the checkFileInfo response for cleaner code

juliushaertl:
Some refactoring that has been lying around for quite some time, mainly to make the checkFileInfo response building a bit more split to logical parts.


juliushaertl:
Need to dive into the failures in more depth, they might be actually wrong test case parts where for a public link the user is being assumed to be a logged in user.

PR #2104, opened 2022-03-23
Locking

juliushaertl:
- [x] Requires https://github.com/nextcloud/server/pull/31676
- [x] Requires https://github.com/nextcloud/files_lock/pull/54


lib/Controller/WopiController.php
```diff
@@ -580,9 +591,11 @@ public function putFile($fileId,
 	 * @return JSONResponse
 	 * @throws DoesNotExistException
 	 */
-	public function putRelativeFile($fileId,
-					$access_token) {
+	public function postFile(string $fileId, string $access_token): JSONResponse {
```
Raudius, line 594: What motivated this name change?

The postFile operation does not exist on Microsoft's WOPI spec
https://docs.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/putrelativefile

lib/Controller/WopiController.php
```diff
@@ -595,7 +608,22 @@ public function putRelativeFile($fileId,
...
+			case 'RENAME_FILE':
+				break; //FIXME: Move to function
+			default:
+				break; //FIXME: Move to function and add error for unsupported method
+		}
+
+
 		$isRenameFile = ($this->request->getHeader('X-WOPI-Override') === 'RENAME_FILE');
 
 		if (!$wopi->getCanwrite()) {
```
Raudius, line 629: Can a user without write permission lock the file?

lib/Controller/WopiController.php
```diff
@@ -711,6 +739,82 @@ public function putRelativeFile($fileId,
...
+
+	/**
+	 * @throws NotFoundException
+	 * @throws GenericFileException
+	 * @throws LockedException
+	 * @throws ShareNotFound
+	 */
+	protected function wrappedFilesystemOperation(Wopi $wopi, callable $filesystemOperation): void {
+		$retryOperation = function () use ($filesystemOperation) {
+			$this->retryOperation(function () use ($filesystemOperation) {
```
Raudius, line 803: Is it possible to pass the callable directly?

```php
$this->retryOperation($filesystemOperation);
```

lib/Controller/WopiController.php
Raudius: Resolving this because I understood the code a bit better from your other comment. I am assuming the rights handling should be done by each operation, In this case its handled by the lock manager

lib/Controller/WopiController.php
juliushaertl: Yes, good catch, addressed

Raudius (approved):

PR #2106, opened 2022-03-24
Replace deprecated String.prototype.substr()

CommanderRoot:
### Summary

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### Checklist

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -24,7 +24,7 @@ let checkingProxyStatus = false
 
 const checkProxyStatus = () => {
 	checkingProxyStatus = true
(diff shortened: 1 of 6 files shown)


juliushaertl (approved):

juliushaertl:
Could you maybe rebase and push again to trigger GitHub action execution? It seems that the last run staled for some reason.

PR #2109, opened 2022-03-25
Refactor connectivity handling

juliushaertl:
- Refactor connectivity handling
  - If requests to the Collabora server fail on the backend we mark the connectivity as failed but keep the cached responses, as in most cases this is a temporary issue. This also allows for more sane error reporting to the user in the future.
  - only perform heavy checks in the background job or when saving
- Improved error reporting for admins
  - Move setup check messages out of the large settings component
  - Show actual error messages in case connecting doesn't work
  - Check frontend reachability by fetching the Collabora domain from the browser
  - Validate setup also when loading the admin page

- [ ] Add cypress test that configures a faulty server and opens a file
- [ ] Add cypress test that configures a valid server, stops it and opens a file (if easily possible)

## Follow up
- [ ] Get status on initial state
- [ ] On timeout manually check status through separate requests
- [ ] Lazy setup for richdocumentscode
- [ ] Make public wopi url configurable as it is currently overwrritten on save


juliushaertl:
- [ ] check for browser html url from discovery and make a quick sanity check of the resulting page to detect misconfigured reverse proxy (200 response)

luka-nextcloud (approved):

juliushaertl:
Waiting to get #3152 in first

juliushaertl:
Replacing this with https://github.com/nextcloud/richdocuments/pull/3315 as I needed some refactoring for URL handling.

Issue #2089, opened 2022-03-07
CSP Error still exists in iOS and Android App


t-markmann:
The bug from https://github.com/nextcloud/richdocuments/issues/2052 and https://github.com/nextcloud/richdocuments/issues/2060 is fixed in the current versions of richdocuments, but it still occurs in the iOS and Android App.

schniepp:
I confirmed that 4.2.3 did work on Android, but 5.0.3 does not. 

Trying to open it, the document appears to load forever. After a few seconds, I get the message "Loading takes very long...". From that it never advances.

brtptrs:
At the moment we can not edit/open any files on android:
Mobile:
NC 3.19.1
Android 10

NC 22.x
Collabora/Richdocuments 4.2.5

Collabora Online :
COOLWSD
21.11.2.4 (git hash: e8ac597)
LOKit
Collabora Office 21.06.20.1 (git hash: 1bc9bd4)

juliushaertl:
Potential fix is in https://github.com/nextcloud/richdocuments/pull/2126 testing is very welcome :)

brtptrs:
Yes this seems to work.
Thank you!

t-markmann:
Will there soon be a release with this fix? Thanks!

PR #2126, opened 2022-04-06
Fixes CSP on Nextcloud installations not on root directory

Raudius:
### Summary
Fixes an issue affecting installations where Nextcloud is not installed on the root directory of the server. When accessing links to richdocuments files these would fail to open due to CSP violation issues when trying to resolve the collabora server scripts.

These requests were incorrectly applying the CSP for OCS/WebDav request and not allowing Collabora scripts to run. 

Fixes https://github.com/nextcloud/richdocuments/issues/2089

### Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -200,7 +200,7 @@ class Application extends App implements IBootstrap {
 
 		// Do not apply CSP rules on WebDAV/OCS
 		// Ideally this could be a middleware running after the controller execution before rendering the result to only do it on page response
-		if ($container->getServer()->getRequest()->getScriptName() !== '/index.php') {
+		if (!str_ends_with($container->getServer()->getRequest()->getScriptName(), '/index.php')) {
 			return;
 		}
 
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -200,7 +200,9 @@ class Application extends App implements IBootstrap {
 
 		// Do not apply CSP rules on WebDAV/OCS
 		// Ideally this could be a middleware running after the controller execution before rendering the result to only do it on page response
-		if (!str_ends_with($container->getServer()->getRequest()->getScriptName(), '/index.php')) {
+		$script_name_parts = explode('/', $container->getServer()->getRequest()->getScriptName());
+		$script_file = end($script_name_parts);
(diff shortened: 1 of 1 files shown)


lib/AppInfo/Application.php
```diff
@@ -200,7 +200,7 @@ public function updateCSP() {
 
 		// Do not apply CSP rules on WebDAV/OCS
 		// Ideally this could be a middleware running after the controller execution before rendering the result to only do it on page response
-		if ($container->getServer()->getRequest()->getScriptName() !== '/index.php') {
+		if (!str_ends_with($container->getServer()->getRequest()->getScriptName(), '/index.php')) {
```
juliushaertl, line 203: Could you switch to something like substr_compare as we need to still support PHP 7.4 for the master branch and 7.3 when back porting this fix to stable release branches for Nextcloud 22/23?

juliushaertl (changes requested):
Small comment about backward compatibility but tested and works otherwise 👍

lib/AppInfo/Application.php
```diff
@@ -200,7 +200,9 @@ public function updateCSP() {
 
 		// Do not apply CSP rules on WebDAV/OCS
 		// Ideally this could be a middleware running after the controller execution before rendering the result to only do it on page response
-		if ($container->getServer()->getRequest()->getScriptName() !== '/index.php') {
+		$script_name_parts = explode('/', $container->getServer()->getRequest()->getScriptName());
+		$script_file = end($script_name_parts);
+		if ($script_file !== 'index.php') {
```
juliushaertl, line 205:
```suggestion
		$scriptNameParts = explode('/', $container->getServer()->getRequest()->getScriptName());
		$scriptFile = end($scriptNameParts);
		if ($scriptFile !== 'index.php') {
```

To stick to https://docs.nextcloud.com/server/latest/developer_manual/getting_started/codingguidelines.html#objects-functions-arrays-variables 😉

lib/AppInfo/Application.php
juliushaertl: Now that was fast 😁

juliushaertl (approved):

juliushaertl:
Let's merge sincr the failures are unrelated and CSP handling ist not covered anyways by those tests yet.

PR #2142, opened 2022-04-11
Add gs.trustedHosts to form-action csp

mickenordin:
This will allow loadbalanced collabora servers to serve frames. 

* Target version: master 

### Summary
Without a way to add individual servers to form-action csp header it is not possible to load balance collabora servers

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -229,6 +229,7 @@ class Application extends App implements IBootstrap {
 				foreach ($trustedList as $server) {
 					$policy->addAllowedFrameDomain($server);
 					$this->addTrustedRemote($policy, $server);
+					$policy->addAllowedFormActionDomain($server);
 				}
 			}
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

mickenordin:
> Makes sense code wise, though I still cannot reproduce the CSP error locally. Since you mention load balanced environments, is the URL that is used for collabora different from the configured on in the Nextcloud admin settings with that?

So the set up is this: We have three collabora servers that is loadbalanced on wopi-src in ha-proxy. The load balanced url is, in this case,  https://document.drive.test.sunet.se which is configured in richdocuments app. The servers themselfs are called document1.drive.test.sunet.se, document2.drive.test.sunet.se and document3.drive.test.sunet.se. csp needs to allow the individual servers, not only the load balanced address, since the individual servers will connect directly to nextcloud.

juliushaertl:
Failures are unrelated and fixed on master, tests pass fine locally.

Issue #2123, opened 2022-04-04
WOPI token expiry


juliushaertl:
- [x] WOPI tokens should have an expiry set and stored in the richdocuments_wopi table
- [ ] Expiry should be configurable by an app config value
- [ ] A background job should clean up expired tokens
- [x] On every WOPI operation it should validate that the token is not expired https://github.com/nextcloud/richdocuments/blob/9ecf2aafc64f8e366c7bae5a73c37da8452bb604/lib/Db/WopiMapper.php#L155
- [ ] Hand over access_token_ttl in
  - https://github.com/nextcloud/richdocuments/blob/master/src/document.js#L186
  - https://github.com/nextcloud/richdocuments/blob/master/src/document.js#L241
- [x] Verify how expiry is handled in Collabora
- [ ] File upstream issue about expired token handling to possibly block further editing or trigger session renewal

From the WOPI spec:
- Access tokens should expire (become invalid) automatically after a period of time. Hosts can use the [access_token_ttl](https://docs.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/concepts#access_token_ttl) property to specify when an access token expires
- access_token_ttl: The access_token_ttl property tells a WOPI client when an access token expires, represented as the number of milliseconds since January 1, 1970 UTC (the date epoch in JavaScript). Despite its misleading name, it does not represent a duration of time for which the access token is valid. The access_token_ttl is never used by itself; it is always attached to a specific access token.

Collabora should properly handle the access_token_ttl, but needs some validation that it works as expected with a low one (e.g. 60 seconds)

Raudius:
> WOPI tokens should have an expiry set and stored in the richdocuments_wopi table

:heavy_check_mark: This is already the case! 

>  Hand over access_token_ttl in

Should be handled in this PR:
https://github.com/nextcloud/richdocuments/pull/2149

> Verify how expiry is handled in Collabora

:heavy_check_mark:  Could not find any information regarding the TTL in the docs, but dug into their server code a bit and verified through testing: a warning is given to the user 15 minutes prior to the expiry of the token.

Raudius:
> On every WOPI operation it should validate that the token is not expired

Implemented in: #2150

juliushaertl:
Moved the renewal topic to a separate issue https://github.com/nextcloud/richdocuments/issues/2170

PR #2149, opened 2022-04-14
Pass TTL token to Collabora

Raudius:
* Resolves: https://github.com/nextcloud/richdocuments/issues/2123
* Target version: master 

### Summary
Adds light-enforcement of the TTL token by passing the token to the Collabora server. This will provide users a warning when their token is about to expire (15 minutes prior to expiry).

Note: this PR does not enforce expiration on our back-end yet, so a user will be able to continue editing documents beyond the expiry of their token. This will come on #2150 .

### Checklist

- [ ] Code is properly formatted
- [ ] Sign-off message is added to all commits
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #2150, opened 2022-04-14
Enforce token expiry on WOPI controllers

Raudius:
* Resolves: #2123 
* Target version: master 

### Summary
Implements handling of expired tokens.
Required adding and handling of custom exceptions within the controllers, the exception types drive which response code to return, as per [the WOPI specification](https://docs.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/putfile#status-codes).

### TODO

- [x] Depends on: https://github.com/nextcloud/richdocuments/pull/2149

### Checklist


lib/Db/WopiMapper.php
```diff
@@ -146,18 +154,14 @@ public function getWopiForToken($token) {
...
 		/** @var Wopi $wopi */
 		$wopi = Wopi::fromRow($row);
 
-		//TODO: validate.
-		if ($wopi->getExpiry() > $this->timeFactory->getTime()){
-			// Expired token!
-			//http_response_code(404);
-			//$wopi->deleteBy('id', $row['id']);
-			//return false;
+		if ($wopi->getExpiry() < $this->timeFactory->getTime()){
```
juliushaertl, line 163: Shall we clean up the token right away here as well assuming that Collabora properly indicates the token invalidation and does not retry other operations with the same token afterwards? Otherwise I'd also be fine to rely on the background job for the cleanup.

Raudius: Collabora will continue using an expired token. 

For example if the user clicks the "save" button collabora will try to save it with the expired token, then if it receives a 401 it will show this messagee
test-odt - Files - Nextcloud 4-19-2022 5-20-06 PM

But the user can continue to edit the doocument and can even download the document.

juliushaertl: All right, then lets stick with that as it is and probably suggest to improve that on the upstream side at https://github.com/CollaboraOnline/online

juliushaertl (approved):

PR #2171, opened 2022-04-28
Add token_ttl app config

Raudius:
* Resolves: #2123
* Target version: master 

### Summary

Adds a `token_ttl` app config value which determines the amount of seconds newly created tokens are valid for.

### Checklist


Raudius:
A few thoughts:

- Should we clamp the config value? What values would be resonable? (1 hour - 24 hours?)
- The WOPI spec allows a TTL of `0`, for indefinite or undetermined length TTL. Do we want to support this?

juliushaertl (approved):

Issue #387, opened 2019-02-07
Revision cannot access documents in memory with Groupfolders


hriekehof:
**Describe the bug**
When i access the revision history from collabora from a file within a groupfolder in NC 15 i get the following error:
- Failed, reading the document from memory. Please contact administrator

**To Reproduce**
1. Put a document into a group folder
2. Make some changes and then try to recall this changes in the revision history window from collabora
3. See error

**Expected behavior**
See the document revision you select in the revision monitor

**Screenshots**
![image](https://user-images.githubusercontent.com/9348635/52401950-4409e800-2ac3-11e9-9863-7a327fd34d8e.png)

The error in englisch is: reading from memory failed

**Client details:**
 - OS: Win10
 - Browser: chrome
 - Version: 71
 - Device: desktop

## Server details

**Operating system**: official nextcloud docker image 15.0.2

**Web server:** apache

**Database:** mariadb

**PHP version:** 7

**Nextcloud version:** 15.0.2

**Version of the richdocuments app** 3.2.0

**Version of Collabora online** 4.0.0.2

<details>

<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
[no app in context] Error: OCP\Lock\LockedException: "files/9d3d8fb84dfca6cb0074460c449bace8" is locked, existing lock on file: none at <<closure>>

 0. /var/www/html/lib/private/Files/Storage/Common.php line 778
    changeLock("files/9d3d8fb84dfca6cb0074460c449bace8", 2)
 1. /var/www/html/lib/private/Files/Storage/Wrapper/Wrapper.php line 615
    changeLock("/__groupfolders ... x", 2, OC\Lock\MemcacheLockingProvider {})
 2. /var/www/html/lib/private/Files/Storage/Wrapper/Jail.php line 469
    changeLock("/__groupfolders ... x", 2, OC\Lock\MemcacheLockingProvider {})
 3. /var/www/html/lib/private/Files/Storage/Wrapper/Wrapper.php line 615
    changeLock("Bestelliste Deu ... x", 2, OC\Lock\MemcacheLockingProvider {})
 4. /var/www/html/lib/private/Files/Storage/Wrapper/Wrapper.php line 615
    changeLock("Bestelliste Deu ... x", 2, OC\Lock\MemcacheLockingProvider {})
 5. /var/www/html/lib/private/Files/Storage/Wrapper/Wrapper.php line 615
[...]

PR #2172, opened 2022-04-28
File history fix behaviour in group folders

Raudius:
* Resolves: #387
* Target version: master 

### Summary
Fixes retrieval of history revisions from files in group folders. 
Additionally this PR also allows seeing past revisions from within the viewer (in read only). This behaviour was already implemented but broken in a previous change.

### Checklist

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -34,7 +34,6 @@ use OCA\Richdocuments\TemplateManager;
 use OCA\Richdocuments\TokenManager;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Db\DoesNotExistException;
-use OCP\AppFramework\Db\TokenExpiredException;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\JSONResponse;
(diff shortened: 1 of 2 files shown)


lib/Controller/WopiController.php
```diff
@@ -125,7 +126,8 @@ public function __construct(
 		UserScopeService $userScopeService,
 		FederationService $federationService,
 		IEncryptionManager $encryptionManager,
-		IGroupManager $groupManager
+		IGroupManager $groupManager,
+		IVersionManager $versionManager
```
juliushaertl, line 130: Unfortunately there is no public API (as the namespace is OCA instead of OCP) so we cannot use dependency injection here as the files_version app can be disabled by admins and then the class would not be available.

We'll need to stick for getting the class from the server container and catch the ContainerExceptionInterface just incase.

Raudius: Nice catch, fixed :)

juliushaertl (approved):
Tested and works 🎉

juliushaertl:
Rebased for a clean history as it was based on the token invalidation.

Issue #2125, opened 2022-04-05
When opening a document many times, some users stay at "0% connecting"


Frando:
**Describe the bug**

I am using Nextcloud Office/Richdocuments with `collabora/code:21.11.3.4.1`, both installed via Docker. The setup works well, until a few users start to edit the same document. At around 5 or 6 users, people start to see a message "Nextcloud office / 0% connecting", and no further changes. The first users can continue to edit the document.

![nextcloud](https://user-images.githubusercontent.com/43627/161731826-e9e54571-bad8-4452-a1df-861e800d4bdf.png)

Interestingly, when users connect or disconnect, the user bubbles still appear and disappear. Also, the websocket connection to collabora is established, and I see the same pattern of messages as in the working editing session.
![nextcloud2](https://user-images.githubusercontent.com/43627/161732703-3ffc7fe7-963e-4f0a-b8b0-a0a92d1d141b.png)

So it seems the connection to Collabora is established and seems to be working. Also, in the Collabora admin panel, the number of connections includes the sessions that stay at "0% connecting". I also couldn't find any errors in the Collabora logs.

On the tabs that stayed at "0% connecting", the browser console log contained one error that was not present in the logs of the working tabs.
```
Uncaught TypeError: app.sectionContainer.getSectionWithName(...) is null
    _hideWizard https://collabora.rdlnet.de/browser/2ea9477/bundle.js:1
    onAdd https://collabora.rdlnet.de/browser/2ea9477/bundle.js:1
    addControl https://collabora.rdlnet.de/browser/2ea9477/bundle.js:1
    initializeBasicUI https://collabora.rdlnet.de/browser/2ea9477/bundle.js:1
    <anonymous> https://collabora.rdlnet.de/browser/2ea9477/bundle.js:1
    <anonymous> https://collabora.rdlnet.de/browser/2ea9477/bundle.js:1
    <anonymous> https://collabora.rdlnet.de/browser/2ea9477/bundle.js:1
bundle.js:1:2270506
```
See below for full logs.

**To Reproduce**
1. Install Nextcloud and Collabora via docker-compose and configure according to instructions
2. Start opening the same document many times (~6+)
3. Experience that the first documents open as intendended, and from the 6th or 7th on they stay at "0% connecting"

**Expected behavior**
All documents open, or the Collabora logs would reveal an error, or the system would run out of RAM.

**Screenshots**
See above

**Client details:**
 - OS: Linux
 - Browser Firefox
 - Version 98
[...]

Issue #2017, opened 2022-02-03
Document not loading - TypeError: fileName is undefined


himpierre:
**Describe the bug**
When opening a document, collabora loads but the document area stays empty.

**To Reproduce**
1. Open a document (does not matter if docx or odt)
2. collabora loads but document area stays empty

**Expected behavior**
Document loads completely.

**Client details:**
 - OS: Linux
 - Browser Firefox
 - Version 96

## Server details

**Operating system**: Linux

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 8.0.15

**Nextcloud version:** 23.0.1

**Version of the richdocuments app** 5.0.1

**Version of Collabora Online** 21.06.13-1

#### Browser log (data/nextcloud.log is empty)
```
Error during post message handler 
Object { msgId: "App_LoadingStatus", args: {…}, deprecated: false }
 TypeError: fileName is undefined
    getModelForFile merged-index.js:4636
    getFileModel FilesAppIntegration.js:190
    initAfterReady FilesAppIntegration.js:80
    postMessageHandler Office.vue:224
    handlePostMessage postMessage.tsx:115
    handlePostMessage postMessage.tsx:109
    v postMessage.tsx:45
    v postMessage.tsx:44
    <anonymous> Office.vue:82
    <anonymous> viewer.js:36
    <anonymous> viewer.js:36
[postMessage.tsx:120:12](webpack:///richdocuments/src/services/postMessage.tsx)
[viewer] Received post message 
Object { msgId: "Action_Load_Resp", args: {…}, deprecated: false }
[Office.vue:214](webpack:///richdocuments/src/view/Office.vue)
Uncaught TypeError: app.sectionContainer.getSectionWithName(...) is null
    _hideWizard https://office.domain.de/browser/739da71/bundle.js:1
    onAdd https://office.domain.de/browser/739da71/bundle.js:1
    addControl https://office.domain.de/browser/739da71/bundle.js:1
    initializeBasicUI https://office.domain.de/browser/739da71/bundle.js:1
    <anonymous> https://office.domain.de/browser/739da71/bundle.js:1
    <anonymous> https://office.domain.de/browser/739da71/bundle.js:1
    <anonymous> https://office.domain.de/browser/739da71/bundle.js:1
[bundle.js:1:2260334](https://office.domain.de/browser/739da71/bundle.js)
[viewer] Received post message 
Object { msgId: "Action_Load_Resp", args: {…}, deprecated: false }
[...]

PR #2187, opened 2022-05-03
When getFileList function returns null, getFileModel function returns…

gokaysatir:
… error.

This commit enables getFileModel function to return null safely. So the real issues that were hidden behind this error should be more visible now.

Signed-off-by: Gökay Şatır <gokaysatir@collabora.com>

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/view/FilesAppIntegration.js
+++ b/src/view/FilesAppIntegration.js
@@ -190,10 +190,10 @@ export default {
 		if (this.fileModel) {
 			return this.fileModel
 		}
-		if (!this.getFileList() || !this.getFileList()?.getModelForFile() || !this.getFileList()._updateDetailsView) {
+		if (!this.getFileList() || !this.getFileList().getModelForFile || !this.getFileList()._updateDetailsView) {
 			return null
 		}
-		this.getFileList()._updateDetailsView && this.getFileList()._updateDetailsView(this.fileName, false)
+		this.getFileList()._updateDetailsView(this.fileName, false)
 		this.fileModel = this.getFileList().getModelForFile(this.fileName)
 
 		if (this.fileModel && this.fileModel.on) {


gokaysatir:
We have a report from a user with below title.
"COOL doesn't load in Firefox"
And there are a couple of GitHub issues here & there related to this bug:

https://github.com/nextcloud/richdocuments/issues/2017
https://github.com/nextcloud/richdocuments/issues/2125

So i chased the issue and fixed a couple of things on Collabora Online. I also saw that a function wasn't working properly. This commit solves that issue.

The work i have done so far can not solve the issue itself. It only pushes the unrelated bugs out of the way.

To reproduce the initial bug that i was chasing:
* Have two or more (logged in nextcloud) users try to open the same file on Firefox.

With above steps, the bug can be almost always reproduced.

If we merge the commit of this PR, we will see no errors on JS console but the original issue will stay.

Lastly, from my notes:
"There is an error on Nextcloud "Viewer.vue" file that initiates "FilesAppIntegration.js (richdocuments)" file without the required parameters."

^This is the point i can't chase any more. @juliushaertl i'll be happy if someone can help me to find the root cause of the bug, thank you :)

juliushaertl:
@gokaysatir I've been trying to reproduce any issue with firefox but wasn't able to see any hidden error that you mentioned in our call. Do you have any additional details on that like the exact error message or stack trace?

Steps taken can be seen in the screen recording, but this was basically two users opening the same time in FF with your patch.

https://user-images.githubusercontent.com/3404133/168336140-40580afd-72ec-4b51-aa0c-cd5a2cb553f5.mov

juliushaertl (approved):
Change makes sense so LGTM but would be still good to chase the other reported issue ;)

juliushaertl:
I think this might be the additional fix needed https://github.com/nextcloud/richdocuments/pull/2212 in addition. @gokaysatir Any chance you could test if that fixes the issue for you as well?

gokaysatir:
> I think this might be the additional fix needed #2212 in addition. @gokaysatir Any chance you could test if that fixes the issue for you as well?

Of course, let me have  a look and inform you.

gokaysatir:
> I think this might be the additional fix needed #2212 in addition. @gokaysatir Any chance you could test if that fixes the issue for you as well?

Hello, i tried it again with & without the PR you pointed. I can confirm that it solves the issue. Hopefully my test is complete and we won't see that problem again:) Thanks for the fixes.

Issue #2094, opened 2022-03-12
`FilesAppIntegration.getFileModel` causes "TypeError: fileName is undefined"


msebas:
Hello,

yesterday I updated 
Nextcloud to 23.0.2 (docker container)
DB: Mariadb
Reverseproxy: nginx

Alongside I installed
Collabora: 21.11.2.4.1
Richtdocuments: 5.0.3

Everything runs in one docker-compose setup.

When I tried to open a freshly created document I got the error message 
```
Error during post message handler 
Object { msgId: "App_LoadingStatus", args: {…}, deprecated: false }
 TypeError: fileName is undefined
    getModelForFile merged-index.js:4636
    getFileModel FilesAppIntegration.js:190
    initAfterReady FilesAppIntegration.js:80
    postMessageHandler Office.vue:231
    handlePostMessage postMessage.tsx:115
    handlePostMessage postMessage.tsx:109
    C postMessage.tsx:45
postMessage.tsx:120:12
```

I took al look to `FilesAppIntegration.getFileModel:190` and there the line is
`if (!this.getFileList() || !this.getFileList()?.getModelForFile() || !this.getFileList()._updateDetailsView) {`

I fixed this assuming that `getModelForFile` needs the current file name to get
`if (!this.getFileList() || !this.getFileList()?.getModelForFile(this.fileName) || !this.getFileList()._updateDetailsView) {`
rebuild and reinstalled the app and then the editor started working fine.

I was unable to find any other case in the repositories where `getModelForFile` does not take an argument, so I just wanted to report this in case it is a real bug. 

Kind regards

msebas:
Well, at least line 190 was not changed from 5.01 to 5.03. 
@aov-arnold: Are you able to build the app from source? If yes I would like to ask you to check if 
https://github.com/msebas/richdocuments/tree/fix%232094 
resolves the problem (the first one, the second one, with the `app.sectionContainer.getSectionWithName(...) is null` in the error message appearing when multiple people edit the same file.). If yes I could turn the branch into a pull request to get the bug fixed in the next version.

aov-arnold:
@msebas I will try tomorrow! 
**Like that?**
- uninstall NC office
- clone your repo to apps directory
- install Node.js and npm in docker container
- run npm ci and npm run dev

aov-arnold:
After setting the Nextcloud min version to 23 I could activate the app.
[...]

PR #2212, opened 2022-05-16
Use visibility instead of display to have the iframe already allocate the size

juliushaertl:
This should address the various reportings where firefox failed loading a document when multiple users joined it due to the Collabora iframe requiring the size of the iframe before the Frame_Ready post message is emitted. We currently hide the iframe until that message is received to avoid having multiple loading indicators of different types. By moving the hiding to use visibility this should already allocate the size properly.

Huge thanks to @alangecker for the in depth debugging and insights provided in  https://github.com/nextcloud/richdocuments/issues/2125#issuecomment-1126820265

Needs a bit more long term testing on my side to verify that it always works reliably. A hot fix to apply manually is available in https://github.com/nextcloud/richdocuments/issues/2125#issuecomment-1127547720 for anyone who would want to help testing this.

Fixes https://github.com/nextcloud/richdocuments/issues/2125
Fixes https://github.com/nextcloud/richdocuments/issues/2094

### DIFF
--- a/src/view/Office.vue
+++ b/src/view/Office.vue
@@ -58,7 +58,7 @@
 					<ActionButton icon="icon-menu-sidebar" @click="share" />
 				</Actions>
 			</div>
-			<iframe v-show="showIframe"
+			<iframe :style="{visibility: showIframe ? 'visible' : 'hidden' }"
 				id="collaboraframe"
 				ref="documentFrame"
 				:src="src" />


luka-nextcloud (approved):

juliushaertl:
Another confirmation https://github.com/nextcloud/richdocuments/issues/2125#issuecomment-1128017851

juliushaertl:
Does not affect stable4 as we do not have the extra loading indicator there.

PR #2221, opened 2022-05-17
Fix viewer styling to adjust to upstream viewer changes

### LINKED ISSUE #2190: Viewer integration broken on master
We need to adjust likely due to https://github.com/nextcloud/viewer/pull/1229

Screenshot 2022-05-06 at 14 58 39

juliushaertl:
Fixes #2190 

While not nice, since the viewer added a `position: relative;` on the parent, this is currently the only way that we can overlay the viewer header and make the richdocuments frame full screen.

### DIFF
--- a/src/view/Office.vue
+++ b/src/view/Office.vue
@@ -337,6 +337,11 @@ export default {
 		flex-direction: column;
 		background-color: var(--color-main-background);
 		transition: opacity .25s;
+
+		.viewer & {
+			height: 100vh;
+			top: -50px;
+		}
 	}
 
 	iframe {


src/view/Office.vue
```diff
@@ -337,6 +337,11 @@ export default {
 		flex-direction: column;
 		background-color: var(--color-main-background);
 		transition: opacity .25s;
+
+		.viewer & {
```
vinicius73, line 341: Is this `&` needed?

eneiluj: @vinicius73 It does not works without it :grin: because `#richdocuments-wrapper` is not a parent of `.viewer` in the DOM.

This rule means it will only affect `#richdocuments-wrapper` if found inside `.viewer`.

eneiluj (approved):
Works fine!

PR #2233, opened 2022-05-20
Add php-cs-fixer

juliushaertl:


lib/Events/BeforeFederationRedirectEvent.php
```diff
@@ -2,45 +2,43 @@
...
-        $this->redirectUrl = $redirectUrl;
-    }
-
-    public function getRedirectUrl() {
-        return $this->redirectUrl;
-    }
-
-}
\ No newline at end of file
+	/** @var Node */
```
vinicius73, line 10: We will still use `@var` annotations, or we can migrate to `Type declarations`

juliushaertl: We can migrate those for Nextcloud 24 or newer, yes.

PR #2234, opened 2022-05-20
Cleanup CSP handling and avoid unreliable reload logic

juliushaertl:


juliushaertl:
Backport to stable4 in https://github.com/nextcloud/richdocuments/pull/2247

PR #2258, opened 2022-06-03
Move document params to initial state

juliushaertl:
Avoid building inline javascript in the template to pass over the required metadata for the document opening.

This also fixes an issue where special characters in a filename like `It's.odt` caused the save as dialog to show double escaped html.

- [x] Integration tests currently parse the inline js https://github.com/nextcloud/richdocuments/blob/9bb9661795e69d2a035522affe0913626ae3f02c/tests/features/bootstrap/RichDocumentsContext.php#L200-L213

### DIFF
--- a/lib/Controller/DirectViewController.php
+++ b/lib/Controller/DirectViewController.php
@@ -175,15 +175,10 @@ class DirectViewController extends Controller {
 				'token_ttl' => $wopi->getExpiry(),
 				'urlsrc' => $urlSrc,
 				'path' => $relativePath,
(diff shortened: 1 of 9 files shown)


juliushaertl:
- [ ] Further app enabled checks calling `OCA.AppId` or `oc_appswebroot` https://github.com/nextcloud/richdocuments/pull/2503/files/3428c1f944fa228818309e3c38431268ad92839f#diff-ce81efd19e432a3cd0658d52cccd7c8fbd4b01f4ada9b174f949578fe9af8578

marcelklehr (approved):

Issue #2267, opened 2022-06-08
Make use of share attributes for prevent download


juliushaertl:
- Based on https://github.com/nextcloud/server/pull/32482
- Get attributes of the file opened
- Set the the appropriate WOPI checkFileInfo fields from https://sdk.collaboraonline.com/docs/advanced_integration.html#checkfileinfo-extended-response-properties
  - DisablePrint
  - DisableExport
  - HideExportOption
  - HidePrintOption

PR #2280, opened 2022-06-20
Prevent download on user shares

eneiluj:
Depends on https://github.com/nextcloud/server/pull/32482
Implements #2267 

Set the `hideDownload` attribute of the Wopi token when it's created. Get this from the share permissions. If multiple shares give access to the file, only one with download disabled will lead to hiding it in Collabora.

- [ ] https://github.com/nextcloud/server/pull/33448

### DIFF
--- a/lib/TokenManager.php
+++ b/lib/TokenManager.php
@@ -22,6 +22,7 @@
 namespace OCA\Richdocuments;
 
 use OC\Files\Filesystem;
+use OCA\Files_Sharing\SharedStorage;
 use OCA\Richdocuments\Db\Direct;
 use OCA\Richdocuments\Db\WopiMapper;
 use OCA\Richdocuments\Db\Wopi;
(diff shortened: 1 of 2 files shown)


Raudius (approved):
LGTM :)

PVince81:
would this also apply for documents accessible through public link share ?
note that in this case there's the "hide download" option already, which is stored in a different field. maybe that logic exists already

eneiluj:
@PVince81 Yes it already exists for public link shares: The view-only mode is enabled according to the "hide download" share attribute.

Will the download permission be relevant in public shares? I mean, do we need to adapt the way we enable the view-only mode for public shares so that we check the "hide download" value AND the download permission?

PVince81:
@eneiluj you don't need to adapt it then. When resharing with a public link, the download permission is converted to "hide download" to be able to benefit from existing logic.

the perm from the share attributes is only intended for local shares.

juliushaertl:
- [ ] Needs another full test round after server PR is merged now

CarlSchwan:
Server PR is merged for master, I'm in the progress of backporting it

juliushaertl (approved):

lib/TokenManager.php
```diff
@@ -147,6 +148,21 @@ public function getToken($fileId, $shareToken = null, $editoruid = null, $direct
 						}
 					}
 				}
+
+				// disable download if at least one shared access has it disabled
+				foreach ($files as $file) {
+					$storage = $file->getStorage();
+					// using string as we have no guarantee that "files_sharing" app is loaded
+					if ($storage->instanceOfStorage('OCA\Files_Sharing\SharedStorage')) {
```
CarlSchwan, line 156:
```suggestion
					if ($storage->instanceOfStorage(OCA\Files_Sharing\SharedStorage::class)) {
```

Should work too even if the app is disabled

juliushaertl:
- [ ] We should make sure this also works if getAttributes is not available in case stable24 is not updated yet to the back ported version

PR #2283, opened 2022-06-24
Wrong french translation "Bureau" for "Office"

proxyconcept:
* Resolves: # 
* Target version: master 

### Summary

In french, the word "Bureau" means "Office" but also "Desktop" !

So in the context of this application, "Bureau" is a wrong choice (the previous untranslated version "Office" was better).

### TODO

An office suite is called in french "Suite bureautique".


thebearon:
I believe translations for richdocuments are done via transifex, not sure about details, @juliushaertl, please provide those.

juliushaertl:
Yes, translations should be handled through https://www.transifex.com/nextcloud/

Issue #1894, opened 2021-12-03
Edit pdf files with Collabora Online isn't available any more


Andreas-Kainz:
**Describe the bug**
In the past there was the option to open a pdf file with collabora online which isn't possible any more.

Andreas-Kainz:
Issue with NextCloud Hub II where you switch to nextcloud office

sebastiaanveld:
Just ran into this, hope this gets fixed.

brtptrs:
This regression breaks a number of internal administrative workflows for us, and many users are complaining. To help moving this forward, I'll be happy to do some testing once some code is available.

PR #2298, opened 2022-07-04
Fix: allow opening PDF with Collabora

Raudius:
* Resolves: #1894 
* Target version: master 

### Summary
Fixes regression where PDFs could not be opened with collabora since we moved to using the Viewer app (https://github.com/nextcloud/richdocuments/pull/1820).

Adds the context menu option to edit PDFs, and registers the actions with the files app. For the action handler we call the new viewer method `openWith` (https://github.com/nextcloud/viewer/pull/1273).

### TODO
- [x] Depends on: https://github.com/nextcloud/viewer/pull/1273

### Checklist

### DIFF
--- a/css/filetypes.scss
+++ b/css/filetypes.scss
@@ -13,3 +13,11 @@
 .icon-filetype-draw {
 	background-image: url('../img/x-office-drawing.svg');
 }
+
+.icon-richdocuments {
+	background-image: url(../img/app-dark.svg);
+}
+
+body.dark .icon-richdocuments {
+	background-image: url(../img/app.svg);
+}
--- a/css/filetypes.scss
+++ b/css/filetypes.scss
@@ -16,8 +16,5 @@
 
 .icon-richdocuments {
 	background-image: url(../img/app-dark.svg);
-}
-
-body.dark .icon-richdocuments {
-	background-image: url(../img/app.svg);
+	filter: var(--background-invert-if-dark);
 }
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -37,3 +37,28 @@ if (OCA.Viewer) {
 		theme: 'light',
 	})
 }
+
+// TODO: Viewer.openWith introduced with https://github.com/nextcloud/viewer/pull/1273
+//       This check can be replaced with `if(OCA.Viewer)` once NC 24 is EOL.
+if (OCA.Viewer.openWith) {
+	const supportedMimes = OC.getCapabilities().richdocuments.mimetypes.concat(OC.getCapabilities().richdocuments.mimetypesNoDefaultOpen)
+	const actionName = 'Edit with ' + OC.getCapabilities().richdocuments.productName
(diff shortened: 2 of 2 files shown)


src/viewer.js
```diff
@@ -37,3 +37,28 @@ if (OCA.Viewer) {
 		theme: 'light',
 	})
 }
+
+// TODO: Viewer.openWith introduced with https://github.com/nextcloud/viewer/pull/1273
+//       This check can be replaced with `if(OCA.Viewer)` once NC 24 is EOL.
+if (OCA.Viewer.openWith) {
+	const supportedMimes = OC.getCapabilities().richdocuments.mimetypes.concat(OC.getCapabilities().richdocuments.mimetypesNoDefaultOpen)
```
juliushaertl, line 44: We should only register optional ones here. Additionally we should probably add pdf to the default mime types if files_pdfviewer is disabled.

css/filetypes.scss
```diff
@@ -13,3 +13,11 @@
 .icon-filetype-draw {
 	background-image: url('../img/x-office-drawing.svg');
 }
+
+.icon-richdocuments {
+	background-image: url(../img/app-dark.svg);
+}
+
+body.dark .icon-richdocuments {
```
juliushaertl, line 21: - [ ] Double check if .dark is still a valid selector, likely has changed on server with the css refactoring

eneiluj: You can use this in NC >= 25:
``` css
filter: var(--background-invert-if-dark);
```
and keep
``` css
body.dark .icon-richdocuments {
    background-image: url(../img/app.svg);
}
```
to stay compatible with NC < 25

eneiluj (changes requested):

css/filetypes.scss
```diff
@@ -13,3 +13,12 @@
 .icon-filetype-draw {
 	background-image: url('../img/x-office-drawing.svg');
 }
+
+.icon-richdocuments {
+	background-image: url(../img/app-dark.svg);
+	filter: var(--background-invert-if-dark);
+}
+
+body.dark .icon-richdocuments {
```
juliushaertl, line 22: Let's drop it since we have one branch per major Nextcloud version ;)

Raudius: Done! We will add it manually for the backports?

src/viewer.js
Raudius: In the case onlyoffice and collabora are both installed I think onlyoffice usually is set as the default for opening Office Open XML mimetypes.

If we only register the optional ones, then there would be no way to open docx with collabora in this case.

juliushaertl: I'd say this is something we should then address separately. Maybe we can indeed have a setting for that or detect if there is an other app in use (ref #2301)

juliushaertl: My main concern is that we add the extra action also in cases when it is the only option to open with Collabora (which is probably the most common case).

Raudius: Makes sense, fixed!

juliushaertl (approved):

Issue #1859, opened 2021-11-11
Collabora still enabled in Nextcloud Mobile App and shared folders


moritz76:
Hello,

Collabora is only shared with certain groups in Nextcloud. In the browser with logged in user it works as well. In the groups for which it is not shared, the documents cannot be opened and no new documents can be created.
However, in the Android app I can create and open documents, even in shared folders there is the possibility to open documents. This should not be.
Also in the settings (as user) you have a tab to Collabora .

1. Go to Admin Settings
2. Enable Collabora  only for specific group
3. switch to user who is not in the specific group
4. Generate a share link, switch to another browser and the documents will be opened with Collabora
5. Switch in the mobile app, also here you can create and open documents with Collabora

**Expected behavior**
Users who are not in the specific group for collabora must not have any access to Collabora.

**Client details:**
 - OS: Windows 10, Android, iOS
 - Browser Firefox, Chrome

## Server details

**Operating system**:
Ubuntu 20.04 LTS

**Web server:**
nginx
**Database:**

**PHP version:**
8.0

**Nextcloud version:**
22.2.0
**Version of the richdocuments app**

**Version of Collabora Online**
4.2.3

Collabora Online with Built-in CODE Server in use

PR #2328, opened 2022-07-25
Limit capabilities to users that are enabled

juliushaertl:
Resolves https://github.com/nextcloud/richdocuments/issues/1859

This PR addresses some issues that happen if usage of Collabora is limited to certain groups:
- First commit makes sure that direct editing capabilities are not exposed to clients where the user doesn't belong to an allowed group
- Second commit makes sure that the scripts for public pages are only loaded if the share owner is member of such a group
- Move to use the LoadViewer and ShareLinkAccessedEvent events (skip when back porting below 23)

### DIFF
--- a/lib/PermissionManager.php
+++ b/lib/PermissionManager.php
@@ -24,7 +24,7 @@ namespace OCA\Richdocuments;
 use OCA\Richdocuments\AppInfo\Application;
 use OCP\IConfig;
 use OCP\IGroupManager;
-use OCP\IUser;
+use OCP\IUserSession;
 
 class PermissionManager {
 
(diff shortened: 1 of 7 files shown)


lib/PermissionManager.php
```diff
@@ -47,16 +53,24 @@ private function splitGroups($groupString) {
...
 		$enabledForGroups = $this->config->getAppValue(Application::APPNAME, 'use_groups', '');
 		if ($enabledForGroups === '') {
 			return true;
 		}
 
 		$groups = $this->splitGroups($enabledForGroups);
-		$uid = $user->getUID();
 		foreach ($groups as $group) {
-			if ($this->groupManager->isInGroup($uid, $group)) {
+			if ($this->groupManager->isInGroup($userId, $group)) {
```
Raudius, line 73: Wondering if we could use `IAppManager::isEnabledForUser` instead

juliushaertl: Richdocuments has its own setting as we need to load the app in early stages to make sure that preview generation can register and act on file system operations. That will disable the server built-in app limitation based on groups. Also this would block the app bootstrap for guests.

https://github.com/nextcloud/server/blob/66ffedcb1662088007711d132df7a630a313c572/lib/private/App/AppManager.php#L232-L233
https://github.com/nextcloud/server/blob/66ffedcb1662088007711d132df7a630a313c572/lib/private/App/AppManager.php#L56-L65

Raudius: That makes sense, thanks for explaining :)

PR looks good then!

Raudius (approved):

Raudius:
I will take care of the backports ;)

juliushaertl:
stable23 https://github.com/nextcloud/richdocuments/pull/2354

PR #2330, opened 2022-07-25
Prefer node with update permission and consolidate getting the file

juliushaertl:
Fixes saving issues when a file is available multiple times in a users home with different permissions.

Steps to reproduce
- Have a groupfolder where user1 has write permissions and user2 read only
- Share a file from user1 to user2 with write permissions
- Try to open the file as user2 for editing

Most of the changes are code cleanup and making use of the getFileForWopiToken method where appropriate


marcelklehr (approved):
Seems good, I did not test

PR #2337, opened 2022-07-27
Use file list instead of dir attribute

PVince81:
I'm working on removing "#dir" so this code needs to be adjusted.

I don't have the env to test this, can someone test with both regular list and public list ?


PVince81:
ref: server PR that sunsets the #dir attribute: https://github.com/nextcloud/server/pull/33373

juliushaertl:
Pushed some more fixes to also cover usages with getElementById and a helper to wrap the logic

juliushaertl (approved):

PR #2350, opened 2022-08-03
Emit event for each user on document open

juliushaertl:
This may be useful for reporting apps to listen to document opening by users


Raudius (approved):
Looks good.

One thing though, is that if we want to do this with Text as well we would need to reimplement the event. Is it something that could be moved to the server?

juliushaertl:
So far there is only a use case for Collabora, but good point to keep that in mind in case additional requirements come up. 👍

PR #2356, opened 2022-08-05
Log error when fetching the capabilities fails

juliushaertl:
It can be quite annoying to debug a failing setup if the request to capabilities fails (e.g. because the certificate is not in the Nextcloud certificate store) and there is no trace about it in the logs. 🙈

### DIFF
--- a/lib/WOPI/DiscoveryManager.php
+++ b/lib/WOPI/DiscoveryManager.php
@@ -81,11 +81,7 @@ class DiscoveryManager {
 			$options['timeout'] = 180;
 		}
 
-		try {
-			return $client->get($wopiDiscovery, $options);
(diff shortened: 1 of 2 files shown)


lib/WOPI/DiscoveryManager.php
```diff
@@ -81,11 +81,7 @@ public function fetchFromRemote() {
 			$options['timeout'] = 180;
 		}
 
-		try {
-			return $client->get($wopiDiscovery, $options);
-		} catch (\Exception $e) {
-			throw $e;
-		}
```
marcelklehr, line 88: heh

marcelklehr (approved):

PR #2372, opened 2022-08-09
fix safari does not focus to document frame on startup

merttumer:
To reproduce the issue on safari or a webkit browser:
Open a document and just start typing
and the typing does appear even though cursor is blinking
on the document.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -197,6 +197,7 @@ const documentsMain = {
 
 			$('#revViewer').append(form)
 			$('#revViewer').append(frame)
+			$('#loleafletframe_viewer').focus()
 
 			// submit that
 			$('#loleafletform_viewer').submit()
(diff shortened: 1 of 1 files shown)


merttumer:
@juliushaertl Hello, can you please review this? Thank you :)

juliushaertl (approved):
Tested and works.

juliushaertl:
@merttumer I still get a js console error about auto focus, but maybe that is unrelated as it works with your change:

Screenshot 2022-08-09 at 10 46 25

merttumer:
> @merttumer I still get a js console error about auto focus, but maybe that is unrelated as it works with your change:
> 
> Screenshot 2022-08-09 at 10 46 25

thanks for the review. I am aware of that issue and it comes from our side. I am merging this then.

PR #2375, opened 2022-08-10
[stable22] Limit capabilities to users that are enabled

max-nextcloud:
* Resolves: #1859 
* Target version: stable4 (stable22 and older)
* backport of the first two commits of https://github.com/nextcloud/richdocuments/pull/2328

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -73,9 +73,11 @@ class Application extends App implements IBootstrap {
 	}
 
 	public function boot(IBootContext $context): void {
+		$context->injectFn(function(ITemplateManager $templateManager, IL10N $l10n, IConfig $config, PermissionManager $permissionManager) {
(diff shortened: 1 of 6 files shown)


max-nextcloud:
@juliushaertl I think i was able to backport the relevant changes. However i don't quite understand the other changes that happened between stable4 and stable23. Please check the changes to `lib/AppInfo/Application.php` in the second commit to make sure they still make sense this way.
Thanks!

max-nextcloud:
@juliushaertl I think the remaining complains in static analysis are false positives that require an update to the psalm baseline. My local php install is not compatible with this. Could you check if my evaluation is correct and update the psalm baseline?

juliushaertl (approved):
Tested and works, pushed a very small commit to not expose new file actions through templates when it is disabled and will prepare a forward port pr for that.

PR #2390, opened 2022-08-17
Fix: hide download option inside Collabora for "hide download" shares

Raudius:
* Resolves: https://github.com/nextcloud/richdocuments/issues/2386
* Target version: master 

### Summary
Corrects logic for "hideDownload" property.


Raudius:
I found its still possible to by-pass the "hide download" permission in instance shares in other ways:
 * Save as (from within Collabora)
 * Copy file (from Files app)
 * Syncing (Webdav clients)

juliushaertl (approved):

PR #2405, opened 2022-08-25
Implement direct editing API

### LINKED ISSUE #721: Implement direct editing API
https://github.com/nextcloud/server/pull/17625

Raudius:
* Resolves: https://github.com/nextcloud/richdocuments/issues/721
* Target version: master 

### Summary
Implements the direct editing API for mimetypes supporrted by Collabora.

### TODO

- [ ] Test on desktop client
- [ ] Test on mobile clients

### Checklist

### DIFF
--- a/lib/DirectEditing/DirectEditor.php
+++ b/lib/DirectEditing/DirectEditor.php
@@ -1,0 +1,198 @@
+<?php
+/**
+ * @copyright Copyright (c) 2022 Raul Ferreira Fuentes <raul@nextcloud.com>
+ *
+ * @author Raul Ferreira Fuentes <raul@nextcloud.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Richdocuments\DirectEditing;
+
+use OCA\Richdocuments\AppConfig;
+use OCA\Richdocuments\AppInfo\Application;
+use OCA\Richdocuments\Capabilities;
+use OCA\Richdocuments\Controller\DocumentTrait;
+use OCA\Richdocuments\Service\InitialStateService;
+use OCA\Richdocuments\TokenManager;
+use OCP\AppFramework\Http\NotFoundResponse;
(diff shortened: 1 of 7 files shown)


lib/DirectEditing/PresentationCreator.php
```diff
@@ -0,0 +1,55 @@
...
+	public function getId(): string {
+		return 'richdocuments_presentation';
+	}
+
+	public function getName(): string {
+		return $this->l10n->t('presentation');
+	}
+
+	public function getExtension(): string {
+		return 'odp';
```
juliushaertl, line 49: Would be good to check for the preferred default document format Open Document Format or Office Open XML and pick the extension / mime type depending on that as done here:

https://github.com/nextcloud/richdocuments/blob/9bb9661795e69d2a035522affe0913626ae3f02c/lib/TemplateManager.php#L459

juliushaertl:
- [ ] Test with desktop and android
- [ ] Ping Tobi about that we need to disable the old mechanism in android
- [ ] Check back with Marino about iOS

lib/DirectEditing/DirectEditor.php
```diff
@@ -0,0 +1,149 @@
...
+	 *
+	 * This behavior is similar to the current direct editing implementation in collabora where we generate a one-time token and switch over to the regular wopi token for the actual editing/saving process
+	 *
+	 * @param IToken $token
+	 * @return Response
+	 */
+	public function open(IToken $token): Response {
+		$token->useTokenScope();
+		try {
+			$session = $this->apiService->create($token->getFile()->getId());
```
juliushaertl, line 134: Seems there is still some logic missing here to make this work ;)

juliushaertl:
@Raudius I released and pushed a small typo fix for the namespace, but seems this is not in a functional state yet. Since you mentioned testing it locally, maybe you didn't push your recent changes? :)

lib/DirectEditing/PresentationCreator.php
Raudius: Good call! Added this to the text/presetnation/spreadsheet creators.

Unsure what to do with the graphics one, as far as I know there is no `odg` equivalent in Microsoft Office.

juliushaertl:
- [ ] start by emitting loaded next to documentLoaded in richdocuments
- [ ] Create a npm package to standardise the postMessages that are exchanged between editor and mobile apps
  - [ ] Migrate "Interfaces called by the clients" to post messages being triggered by the mobile apps through [window.postMessage()](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage)
    - For the client called methods we either require them (onClose) or they can be skipped by editors (OCA.RichDocuments.documentsMain.postGrabFocus / OCA.RichDocuments.documentsMain.postAsset)
  - [ ] Move postMessage parsing from https://github.com/nextcloud/richdocuments/blob/81b09056fdcc7d3a845bf1ba0c2e7f4d179d1c10/src/helpers/mobile.js over to the library
  - [ ] Library API would need methods to 
    - [ ] emit a message to mobile apps (callMobileMessage)
    - [ ] register a listener for incoming messages from the mobile apps (listenForMobileMessage)
  - [ ] List of supported messages is documented and validated in the library

juliusknorr:
Meanwhile replaced by https://github.com/nextcloud/richdocuments/pull/5656

Issue #2418, opened 2022-08-31
Add file open endpoint for Collabora Office


juliushaertl:
The idea is to have a button inside of Collabora Online / LibreOffice that allows the user to switch to editing online by clicking it.

Richdocuments would:
- [ ] Provide an endpoint with a path parameter to open the file for the user
  - It should be enough to get the file id and redirect to the files app with the openfile parameter
- [ ] Expose the endpoint URL through capabilities, so that it can be picked up by the client

The desktop client would expose a `.editonline.ini` (naming could change) file in the root sync folder which contains the edit url as well as the userid and relative path, which can then be read by Collabora Office.

- [ ] Fetch the endpoint from the capabilities
- [ ] For each sync folder put a .ini file in the root, containing the format below

### Example if the Nextcloud root is synced to ~/Nextcloud, there would be a ~/Nextcloud/.editonline.ini:

````
[editserver]
    endpoint = https://nextcloud.example.com/index.php/apps/richdocuments/editfromlocal
    userid = admin
    relativepath = /
```

Considering a file Document.odt is opened, Collabora Office could then construct url to open in the browser:
https://nextcloud.example.com/index.php/apps/richdocuments/editfromlocal?user=admin&path=/Document.odt

### Example if the Nextcloud subdirectory /Engineering is synced to ~/Nextcloud, there would be a ~/Nextcloud/.editonline.ini:

````
[editserver]
    endpoint = https://nextcloud.example.com/index.php/apps/richdocuments/editfromlocal
    userid = admin
    relativepath = /Engineering
```

Considering a file Document.odt is opened, Collabora Office could then construct url to open in the browser:
https://nextcloud.example.com/index.php/apps/richdocuments/editfromlocal?user=admin&path=/Engineering/Document.odt

PR #2419, opened 2022-09-01
Implement endpoint to open files in the browser

juliushaertl:
Implements #2418 

Adds the following entry to the richdocuments capabilities:
```
<editonline_endpoint>http://nextcloud.dev.local/index.php/apps/richdocuments/editonline</editonline_endpoint>
```

A file can be opened through:
http://nextcloud.dev.local/index.php/apps/richdocuments/editonline?path=/Test.odt&userId=admin

- If the user is not logged in they will be redirected to the login page
- If the user is logged in they will see an error
- If the passed user id is different from the currently logged in user, they will see a hint about that


juliushaertl:
Was done as part of #2975

PR #2427, opened 2022-09-06
[stable24] Viewer top margin is not needed

eneiluj:
On NC 24, adding a `-50px` margin hides the collabora menu and the viewer buttons.


Raudius:
With this branch I get a double header

image

juliushaertl:
@eneiluj Hiding the viewer header was a explicit design decision discussed with @jancborchardt

Was there any report for this to be a bug?

PR #2428, opened 2022-09-06
Open PDF files by default when PDF viewer is disabled

Raudius:
Related to https://github.com/nextcloud/files_pdfviewer/issues/649

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -23,6 +23,7 @@
 
 namespace OCA\Richdocuments;
 
+use OCP\App\IAppManager;
 use OCA\Richdocuments\Service\CapabilitiesService;
 use OCP\Capabilities\ICapability;
 use OCP\IL10N;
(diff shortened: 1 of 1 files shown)


lib/Capabilities.php
```diff
@@ -82,14 +83,17 @@ class Capabilities implements ICapability {
 	private $capabilitiesService;
 	/** @var PermissionManager */
 	private $permissionManager;
+	/** @var AppManager */
+	private $appManager;
 
 	private $capabilities = null;
 
-	public function __construct(IL10N $l10n, AppConfig $config, CapabilitiesService $capabilitiesService, PermissionManager $permissionManager) {
+	public function __construct(IL10N $l10n, AppConfig $config, CapabilitiesService $capabilitiesService, PermissionManager $permissionManager, AppManager $appManager) {
```
juliushaertl, line 91:
```suggestion
	public function __construct(IL10N $l10n, AppConfig $config, CapabilitiesService $capabilitiesService, PermissionManager $permissionManager, IAppManager $appManager) {
```

Should use the IAppManager from the OCP namespace which is considered the public and stable API.

juliushaertl (commented):
Small comment regarding the app manager, but otherwise LGTM 👍

lib/Capabilities.php
Raudius: Fixed! 

Also modified my include path in PhpStorm to just include the public API, should avoid this in the future :)

PR #2442, opened 2022-09-15
Local editing

Raudius:
* Target version: master 

### Summary
Adds a button to the Collabora editor which enables the file to be opened locally.

Clicking the button will:
  * Trigger a warning that the file will close for all users currently viewing the file with Collabora
  * Save the file
  * Close all currently open collabora sessions of the file
  * Unlock the file (if locked)
  * Open an `nc://` url to be handled by the desktop client

### TODO

- [ ] Remove the button when viewing read-only files (csv, pdf)?

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -1,5 +1,5 @@
 import { emit } from '@nextcloud/event-bus'
-import { getRootUrl } from '@nextcloud/router'
+import { getRootUrl, imagePath } from '@nextcloud/router'
 import { getRequestToken } from '@nextcloud/auth'
 import Config from './services/config.tsx'
 import { setGuestName, shouldAskForGuestName } from './helpers/guestName'
@@ -11,9 +11,10 @@ import {
 	isDirectEditing,
 	isMobileInterfaceAvailable,
 } from './helpers/mobile'
-import { getWopiUrl, getSearchParam } from './helpers/url'
+import { getWopiUrl, getSearchParam, getNextcloudUrl } from './helpers/url'
 
 import '../css/document.scss'
+import axios from '@nextcloud/axios'
 
 const PostMessages = new PostMessageService({
 	parent: window.parent,
@@ -307,6 +308,15 @@ const documentsMain = {
 
 						if (Config.get('userId') === null) {
 							PostMessages.sendWOPIPostMessage('loolframe', 'Hide_Menu_Item', { id: 'insertgraphicremote' })
+						} else {
+							PostMessages.sendWOPIPostMessage('loolframe', 'Insert_Button', {
+								id: 'Open_Local_Editor',
+								imgurl: window.location.protocol + '//' + getNextcloudUrl() + imagePath('richdocuments', 'launch.svg'),
+								mobile: false,
+								label: t('richdocuments', 'Open in local editor'),
+								hint: t('richdocuments', 'Open in local editor'),
+								insertBefore: 'undo',
+							})
 						}
 
 						emit('richdocuments:wopi-load:succeeded', {
@@ -442,6 +452,23 @@ const documentsMain = {
 					case 'File_Rename':
 						documentsMain.fileName = args.NewName
 						break
+					case 'Clicked_Button':
+						if (parsed.args.Id === 'Open_Local_Editor') {
+							documentsMain.UI.initiateOpenLocally()
+						} else {
+							console.debug('[document] Unhandled `Clicked_Button` post message', parsed)
+						}
+						break
+					case 'Get_Views_Resp':
+						if (documentsMain.openingLocally) {
+							documentsMain.UI.removeViews(parsed.args)
+							documentsMain.unlockFile()
+								.catch(_ => {}) // Unlocking failed, possibly because file was not locked, we want to proceed regardless.
+								.then(() => {
+									documentsMain.openLocally()
+								})
+						}
+						break
 					default:
 						console.debug('[document] Unhandled post message', parsed)
 					}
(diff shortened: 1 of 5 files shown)


src/document.js
```diff
@@ -500,6 +523,57 @@ const documentsMain = {
 				$(document.body).removeClass('claro')
 			})
 		},
+
+		initiateOpenLocally() {
+			OC.dialogs.confirmDestructive(
```
juliushaertl, line 528: Could we maybe only issue the confirm action if there are other users present in the document? The list of views should be available.

juliushaertl:
> Remove the button when viewing read-only files (csv, pdf)?

Might still be relevant to open in a preferred viewer app I'd say.

src/document.js
```diff
@@ -500,6 +523,57 @@ const documentsMain = {
...
+			.catch(_ => {}) // Unlocking failed, possibly because file was not locked, we want to proceed regardless.
+			.then(() => {
+				documentsMain.openLocally()
+			})
+	},
+
+	openLocally() {
+		if (documentsMain.openingLocally) {
+			documentsMain.openingLocally = false
+			window.location.href = 'nc://open/' + Config.get('userId') + '@' + window.location.host + documentsMain.fullPath
```
juliushaertl, line 575: @allexzander Does the nc:// protocol handle nextcloud servers installed in a subdirectory? e.g. under (https://example.com/nextcloud/) ? Would the subdirectory then need to be contained in the open url ?

juliushaertl, line 575: Maybe we can adapt the logic from https://github.com/nextcloud/server/pull/34171/files#diff-dedceb236d4709ee562b67512f8ad10bce9dc7a494e9b1cfdabe8ea8ad33ca25R2810-R2818 to also take the port into account

src/document.js
```diff
@@ -307,6 +308,15 @@ const documentsMain = {
 
 						if (Config.get('userId') === null) {
 							PostMessages.sendWOPIPostMessage('loolframe', 'Hide_Menu_Item', { id: 'insertgraphicremote' })
+						} else {
+							PostMessages.sendWOPIPostMessage('loolframe', 'Insert_Button', {
+								id: 'Open_Local_Editor',
+								imgurl: 'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIGhlaWdodD0iMjRweCIgdmlld0JveD0iMCAwIDI0IDI0IiB3aWR0aD0iMjRweCIgZmlsbD0iIzAwMDAwMCI+PHBhdGggZD0iTTAgMGgyNHYyNEgweiIgZmlsbD0ibm9uZSIvPjxwYXRoIGQ9Ik0xOSAxOUg1VjVoN1YzSDVjLTEuMTEgMC0yIC45LTIgMnYxNGMwIDEuMS44OSAyIDIgMmgxNGMxLjEgMCAyLS45IDItMnYtN2gtMnY3ek0xNCAzdjJoMy41OWwtOS44MyA5LjgzIDEuNDEgMS40MUwxOSA2LjQxVjEwaDJWM2gtN3oiLz48L3N2Zz4=',
+								mobile: false,
+								label: 'Open in local editor',
+								hint: 'Open in local editor',
```
juliushaertl, line 317:
```suggestion
								label: t('richdocuments', 'Open in local editor'),
								hint: t('richdocuments', 'Open in local editor'),
```

src/document.js
```diff
@@ -307,6 +308,15 @@ const documentsMain = {
 
 						if (Config.get('userId') === null) {
 							PostMessages.sendWOPIPostMessage('loolframe', 'Hide_Menu_Item', { id: 'insertgraphicremote' })
+						} else {
+							PostMessages.sendWOPIPostMessage('loolframe', 'Insert_Button', {
+								id: 'Open_Local_Editor',
+								imgurl: 'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIGhlaWdodD0iMjRweCIgdmlld0JveD0iMCAwIDI0IDI0IiB3aWR0aD0iMjRweCIgZmlsbD0iIzAwMDAwMCI+PHBhdGggZD0iTTAgMGgyNHYyNEgweiIgZmlsbD0ibm9uZSIvPjxwYXRoIGQ9Ik0xOSAxOUg1VjVoN1YzSDVjLTEuMTEgMC0yIC45LTIgMnYxNGMwIDEuMS44OSAyIDIgMmgxNGMxLjEgMCAyLS45IDItMnYtN2gtMnY3ek0xNCAzdjJoMy41OWwtOS44MyA5LjgzIDEuNDEgMS40MUwxOSA2LjQxVjEwaDJWM2gtN3oiLz48L3N2Zz4=',
```
juliushaertl, line 314: Maybe we could use something like https://github.com/bhovhannes/svg-url-loader to still have a .svg file in the repo and just build the data representation during build, but not a blocker ;)

juliushaertl:
@Raudius I don't fully remember, but the icon was agreed on with @jancborchardt, right? 

Screenshot 2022-09-22 at 09 25 00

src/document.js
Raudius: I had it that way originally but changed it because the user might accidentally click it (when trying to save for example), and they might panic when the document suddenly closes.

Also we might want to add in the dialog that it only works if the desktop client (v3.6.0+) is installed and running

src/document.js
Raudius: Changed to use the `imagePath` function. 

Realised why it didnt work before, `imagePath` generates a relative path which would open from the collabora hostname. Made sure to prepend the NC instance url :)

Raudius:
> @Raudius I don't fully remember, but the icon was agreed on with @jancborchardt, right?

Icon and tooltip label were amended with Jan's suggestions :)

src/document.js
```diff
@@ -307,6 +308,15 @@ const documentsMain = {
 
 						if (Config.get('userId') === null) {
 							PostMessages.sendWOPIPostMessage('loolframe', 'Hide_Menu_Item', { id: 'insertgraphicremote' })
+						} else {
+							PostMessages.sendWOPIPostMessage('loolframe', 'Insert_Button', {
+								id: 'Open_Local_Editor',
+								imgurl: window.location.protocol + '//' + getNextcloudUrl() + imagePath('richdocuments', 'launch.svg'),
+								mobile: false,
+								label:  t('richdocuments', 'Open in local editor'),
+								hint:  t('richdocuments', 'Open in local editor'),
```
juliushaertl, line 317:
```suggestion
								label: t('richdocuments', 'Open in local editor'),
								hint: t('richdocuments', 'Open in local editor'),
```

eslint complains about that

juliushaertl (approved):
One eslint error but let's get this in once fixed 👍

PR #2451, opened 2022-09-20
Always show the close button

juliushaertl:
Avoid being locked in with direct editing on iOS.


pedropintosilva (approved):
Tested on iPad (NC server master + richdocuments checkout this branch + CO)
It works as expected. The close button is always visible. It works in both modes (read only and edit mode).

PR #2483, opened 2022-10-03
Cleanup frontend dev dependencies

juliushaertl:
Replaces https://github.com/nextcloud/richdocuments/pull/2440


julien-nc (approved):
Works fine with Node 16 and Npm 8

PR #2490, opened 2022-10-07
Always enable editing for admins

juliushaertl:
Make sure that admins always can work with Collabora to avoid breaking settings

### DIFF
--- a/lib/PermissionManager.php
+++ b/lib/PermissionManager.php
@@ -74,6 +74,10 @@ class PermissionManager {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


eneiluj (approved):
Funny to think of admins locking themselves out.
:+1: 

Wouldn't it make sense to check `isAdmin` before the `use_groups` config value? I don't know which is more costly between `isAdmin` and `config->getAppValue`...

juliushaertl:
Shouldn't matter much in terms of performance, especially since the admin check is probably the edge case ;)

PR #2502, opened 2022-10-14
Add config option to turn on Collabora feature lock for read only users

juliushaertl:
Implementing the option to enable [feature locking](https://sdk.collaboraonline.com/docs/installation/Configuration.html?highlight=isuserlocked#feature-locking) for all users that are not within edit groups

Can be enabled with:

```
occ config:app:set richdocuments read_only_feature_lock --value=yes
```

### DIFF
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -22,6 +22,8 @@ class AppConfig {
 
 	public const SYSTEM_GS_TRUSTED_HOSTS = 'gs.trustedHosts';
 
+	public const READ_ONLY_FEATURE_LOCK = 'read_only_feature_lock';
+
 	private $defaults = [
 		'wopi_url' => '',
 		'timeout' => 15,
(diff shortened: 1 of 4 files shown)


julien-nc (commented):
Looks good.

* What's the effect of enabling `IsUserLocked` in the file info?
* What's the use case of disabling `read_only_feature_lock`? Is it just to override admin settings?

juliushaertl:
> What's the effect of enabling IsUserLocked in the file info?

Admins can limit functionality in the coolwsd.xml and configure a modal (e.g. for an upselling path at hosting providers)

> What's the use case of disabling read_only_feature_lock? Is it just to override admin settings?

By default it is disabled, when turned on every user that does not have edit permission will be considered locked with the above mentioned options from Collabora.

julien-nc (approved):
:+1:

PR #2503, opened 2022-10-14
Disable sharing for public links and versions if app is disabled

juliushaertl:
- Do not pass the revision history parameter if files_versions is disabled
- Do not enable sharing for public links

### DIFF
--- a/src/helpers/url.js
+++ b/src/helpers/url.js
@@ -33,6 +33,9 @@ const getSearchParam = (name) => {
 }
 
 const getWopiUrl = ({ fileId, title, readOnly, closeButton, revisionHistory }) => {
+	// Only set the revision history parameter if the versions app is enabled
+	revisionHistory = revisionHistory && window?.oc_appswebroots?.files_versions
+
 	// WOPISrc - URL that loolwsd will access (ie. pointing to ownCloud)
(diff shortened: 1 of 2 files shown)


src/helpers/url.js
```diff
@@ -33,6 +33,9 @@ const getSearchParam = (name) => {
 }
 
 const getWopiUrl = ({ fileId, title, readOnly, closeButton, revisionHistory }) => {
+	// Only set the revision history parameter if the versions app is enabled
+	revisionHistory = revisionHistory && window?.oc_appswebroots?.files_versions
+
```
julien-nc, line 38: Is this a reliable way to check if an app is enabled? I mean are we sure it won't get removed anytime soon?

If not we could get the information via initial state I guess.

julien-nc (commented):
Make sense!

Just a question.

src/helpers/url.js
juliushaertl: Yeah, ideally we should have initial state for that, but I'd stick to this for now to keep it back portable and maybe pass over more stuff through initial state when getting back to #2258

julien-nc: :+1:

julien-nc (approved):

PR #2512, opened 2022-10-14
[stable24] Use correct server version for integration tests

juliushaertl:


juliushaertl:
All green, merging to keep moving with green CI again

PR #2550, opened 2022-10-20
Keep passing light theme values to Collabora as long as the dark mode is not available there

juliushaertl:
Otherwise the primary tinted colors are wrong and lead to dark icons on dark background

Fixes https://github.com/nextcloud/richdocuments/issues/2302

## Considerations for Nextcloud 24 and lower

For back porting we will need to add the removal of the user accessibility stylesheet, e.g. with this:

```
document.querySelector("link[href*='accessibility\/css\/user']").remove()
```

Quite ugly but I don't see a better way to get the light mode css variables.

### DIFF
--- a/src/helpers/coolParameters.js
+++ b/src/helpers/coolParameters.js
@@ -61,12 +61,13 @@ const generateCSSVarTokens = () => {
 		'--border-radius-pill': '--co-border-radius-pill',
 	}
 	let str = ''
+	const element = document.getElementById('documents-content') ?? document.documentElement
 	try {
 		for (const cssVarKey in cssVarMap) {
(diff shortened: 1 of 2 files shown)


juliushaertl:
I remember there was a ticket for this but cannot find it 🙈

Raudius:
> I remember there was a ticket for this but cannot find it see_no_evil

Is it the ticket 2302? 

(Not linking it so that it doesnt show up as mentioned in case its the wrong one :p)

Raudius (approved):

juliushaertl:
> > I remember there was a ticket for this but cannot find it see_no_evil
> 
> Is it the ticket 2302?
> 
> (Not linking it so that it doesnt show up as mentioned in case its the wrong one :p)

Yes, thanks :)

PR #2560, opened 2022-10-21
Add app config to override wopi checkFileInfo

juliushaertl:
Allows to fine tune options e.g. with:

```
occ config:app:set richdocuments wopi_override --value='{"EnableShare":false}'
```

### DIFF
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -162,4 +162,16 @@ class AppConfig {
 	private function splitGroups(string $groupString): array {
 		return explode('|', $groupString);
 	}
+
+	/**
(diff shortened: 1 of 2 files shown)


marcelklehr (approved):

marcelklehr:
Psalm error:

ERROR: NullArgument - lib/AppConfig.php:170:89 - Argument 3 of OCP\IConfig::getAppValue cannot be null, null value provided to parameter with type string (see https://psalm.dev/057)
		if ($wopiOverride = $this->config->getAppValue(Application::APPNAME, 'wopi_override', null)) {

Issue #1380, opened 2021-02-24
Notify users on mentions in document comments


thkoch2001:
## Current status

- [x] Evaluate filetype/core requirements in https://github.com/CollaboraOnline/online/issues/1529
- [x] Implement handling of mentions in core
- [x] Implement some user listing mechanism on the Nextcloud side
- [x] Implement handling of mentions in online
- [ ] Listen to the post message for inserted mention and emit a notifcation to the user

## Pull requests

- https://github.com/CollaboraOnline/online/pull/5334
- https://github.com/CollaboraOnline/online/pull/5093
- https://github.com/nextcloud/richdocuments/pull/2576

## Initial request

**Is your feature request related to a problem? Please describe.**
- As an owner of a document I want to get notified on any comment that gets added to my document in order to provide a timely answer.
- As a commenter of a document I want to be able to mention a person in a comment and thus raise their awareness of an action or input that is needed from their side.
- As a reader of a document I want to be able to subscribe to a document to get informed about all comments to that document.
- As a participant in a comment thread I want to be notified of all updates to all new comments to a comment thread.
- As a participant in a comment thread I want to be notified when the thread gets resolved.

**Describe the solution you'd like**
Destroy any reason to use Google Docs ever again.

**Describe alternatives you've considered**
None.

**Additional context**
https://support.google.com/docs/answer/65129

juliushaertl:
Other solutions have their own file format but when exporting to a regular office file, the mention is purely identified by the users email address. For the Nextcloud context we'd ideally have the user id or federated cloud id present as some kind of metadata, in order to uniquely identify the mentioned user. We are currently discussion with Collabora how we could store mentions in the comments in the actual file.

juliushaertl:
Findings so far, ODF does not seem to have any good fit for storing a unique user identifier (like the user id or federated cloud id) within the document, so while we could probably process mentions while being edited in COOL there is no good way we found until now to persist those properly. @kendy is still looking into options from the Collabora side.

juliushaertl:
@Raudius Noticed that we had this ticket for that and assigned you since we talked about that you would look into that.

Raudius:
> @Raudius Noticed that we had this ticket for that and assigned you since we talked about that you would look into that.

Cool.

One thing to note is that with the current implementation this would only work in the document text (not in the comments). I recall that due to document format reasons getting it to work on comments is a lot more challenging.

juliushaertl:
Reopening for the remaining task:

[...]

PR #2576, opened 2022-10-24
Implement user mentions API

Raudius:
* Resolves: https://github.com/nextcloud/richdocuments/issues/1380
* Target version: master 

### Summary
Implements the post message API to allow user mentions in Collabora


src/document.js
```diff
@@ -604,13 +612,17 @@ const documentsMain = {
 				return
 			}
 
-			const list = result.data.ocs.data.map((user) => {
+			return result.data.ocs.data
+		}).catch(() => {
+			return documentsMain.users
+		}).then((users) => {
+			const list = users.map((user) => {
```
juliushaertl, line 619: Could we do this right away if there is no user session available so that we skip the failing request?

juliushaertl (commented):
Just one minor comment and eslint needs some love. Otherwise good from my side 👍

src/document.js
Raudius: Yep nice suggestion, I changed the logic around

PR #2579, opened 2022-10-25
Listen for event during preview rendering and apply secure view options

juliushaertl:

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Raudius (approved):
LGTM :)

Only thing I noticed is that you close a read-only watermarked file, you get a pretty bare view. 

<details>
<summary>Screenshot:</summary>

image

</details>

Might be something to look into with the design team after this PR is merged :)

juliushaertl:
Let me actually have a look at that, because closing is something we can probably disable in the collabora view :)

juliushaertl:
Might need some extra backend handling to hand over if we are on a single or folder share, will open a follow up issue for that.

PR #2580, opened 2022-10-25
Fix the theming adjustments when accesibility app is disabled

Raudius:
* Resolves: https://help.nextcloud.com/t/nextcloud-office-does-not-load/148321
* Target version: master 

### Summary
When the accessibility app the query-selector returns `null` which causes a failure.


Raudius:
@juliushaertl 

richdocuments is broken in NC23.:see_no_evil:  

So it would be nice to get this in and do another release this week

juliushaertl (approved):

PR #2604, opened 2022-11-08
Show warning if "WOPI allow-list" not configured

Raudius:
### Summary

Adds warning if allow-list is not configured. Without this setting all WOPI requests are served to any IP, which allows users to download file which they might not have access to (download disabled share, secure view...) via WOPI requests.


juliushaertl (approved):

juliushaertl:
Looks good, Some eslint errors still need fixing though ;)

Issue #1738, opened 2021-08-30
"Edit" button is positioned too low on mobile devices


scroom:
**Describe the bug**
As described above. Unfortunately, the button is positioned so low that you can hardly click it. Attached is a screenshot from an iPad and the Safari browser. However, the problem also occurs on other browsers. On chrome on the iPad the button isn't even visible.

**To Reproduce**
1. If logged in to Nextcloud click on an Office file
2. Collabora is loading
3. the "edit" button is positioned so low that is hardly clickable or even invisible and so Collabora Office could not be used to edit documents.

**Expected behavior**
Button should be visible and clickable

**Screenshots**
<img width="810" alt="2021-08-29 20-48-55-0" src="https://user-images.githubusercontent.com/5347731/131410414-e5364341-d7f1-4298-8511-dd829ea6c8df.png">
![IMG_0079](https://user-images.githubusercontent.com/5347731/131410430-696a61b5-9f7b-4ccb-a257-b4ab270fd77d.PNG)

**Client details:**
 - OS: IOS 
 - Browser: Safari and even worse Chrome
 - Version Chrome 92.0.4515.90
 - Device: iPad

## Server details

**Operating system**: Ubuntu 20.04

**Web server:** Nginx

**Database:** MySQL

**PHP version:** 7.4

**Nextcloud version:** 21.0.4

**Version of the richdocuments app** 4.2.2

**Version of Collabora Online** 6.4

juliushaertl:
Can you maybe share where exactly you opened the document, e.g. if it was on Talk or the Dashboard or the regular files app?

scroom:
It has been the files app.

thebearon:
Based on the screenshots, it looks like the files have been opened from the dashboard to me.

scroom:
> Based on the screenshots, it looks like the files have been opened from the dashboard to me.

Oops. You're right. I thought I had opened the files from the Files app. But after a test I just did, I can confirm that the Edit button in Safari and in Chrome on the iPad are displayed properly when Collabora is launched from the Files app.
However, when the app is launched from the Dashboard, the button is displayed too low.

scroom:
Is it more of a bug in the Dashboard app then? @rullzer

rullzer:
Why am I tagged here?

scroom:
> Why am I tagged here?

I've seen that you are one of the main developers of the dashboard, and I didn't knew where it would be best to post this bug.

[...]

PR #2605, opened 2022-11-08
Another attempt to work around Safari issues

juliushaertl:
The problems with Safari still seem to cause some issues, therefore this
consolidates the fixes into a helper file and provides enable/disable
methods to be turned on during document loading/closing.

There are two issues:
- Safari does have a weired understanding of the document height in css,
  so we need to resize containers accordingly to fit the visible
  viewport when the keyboard is shown
- Safari does not always respect fixed positioning on the body element
  and still allows scrolling if the keyboard is shown, therefore
  manually setting the scrollTop is required when resizing or attempts to
  scroll

Fixes https://github.com/nextcloud/richdocuments/issues/1738

Some things left to still check:
- [ ] See if we actually need the interval callback
- [ ] Maybe we can only run the code on iOS to avoid side effects with other platforms

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -4,6 +4,7 @@ import { getRequestToken } from '@nextcloud/auth'
 import Config from './services/config.tsx'
 import { setGuestName, shouldAskForGuestName } from './helpers/guestName.js'
 import { getUIDefaults, generateCSSVarTokens, getCollaboraTheme } from './helpers/coolParameters.js'
+import { enableScrollLock } from './helpers/safariFixer.js'
 
 import PostMessageService from './services/postMessage.tsx'
 import {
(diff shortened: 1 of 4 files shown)


juliushaertl:
@pedropintosilva I remember you also had an iOS device for testing, so I'd be curious if you could also give this a test run. This should avoid scrolling the collabora UI out of the visible area in both Safari through the Nextcloud interface as well as when using the in app web view in the Nextcloud iOS app.

pedropintosilva (approved):
I have tested on iPad and this seems to fix the issue both when using (normally) safari and also when using safari via home-screen shortcut (so the website opens without safari top bar etc)

PHP-Expert:
Any updates on this?

gokaysatir:
I think this can be merged.

PR #2626, opened 2022-11-17
Pass the proper target path/name for new files created through direct editing

juliushaertl:
### Summary

Steps to reproduce:
- Create a new file on mobile apps
- Try to save as 
- See that the filename now shows the actual filename

### Checklist


marcelklehr (approved):
looks good

PR #2657, opened 2022-12-02
introduce zotero integration for Nextcloud Office (Collabora Online)

lpranam:
Signed-off-by: Pranam Lashkari <lpranam@collabora.com>

* Target version: master 

### Summary
Collabora is working on integrating Zotero online, allowing users to add citations from personal library collections. We need their API key to get the Zotero library collection access from the user account. This PR enables users to specify API keys in the personal office settings. If the key is not specified then the user will not be displayed any Zotero feature in UI.

Initial Zotero PR in Collabora online: https://github.com/CollaboraOnline/online/pull/5521

### Checklist

### DIFF
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -285,13 +285,22 @@ class SettingsController extends Controller {
 	 * @param $value
 	 * @return JSONResponse
 	 */
-	public function setPersonalSettings($templateFolder) {
+	public function setPersonalSettings($templateFolder,
+		$zoteroAPIKeyInput) {
 		$message = $this->l10n->t('Saved');
 		$status = 'success';
 
 		if ($templateFolder !== null) {
 			try {
 				$this->config->setUserValue($this->userId, 'richdocuments', 'templateFolder', $templateFolder);
+			} catch (PreConditionNotMetException $e) {
+				$message = $this->l10n->t('Error when saving');
+				$status = 'error';
+			}
+		}
+		if ($zoteroAPIKeyInput !== null) {
+			try {
+				$this->config->setUserValue($this->userId, 'richdocuments', 'zoteroAPIKey', $zoteroAPIKeyInput);
 			} catch (PreConditionNotMetException $e) {
 				$message = $this->l10n->t('Error when saving');
 				$status = 'error';
--- a/src/personal.js
+++ b/src/personal.js
@@ -1,5 +1,6 @@
 import '../css/admin.scss'
 import { generateFilePath } from '@nextcloud/router'
+import { showError } from '@nextcloud/dialogs'
 
 $(function() {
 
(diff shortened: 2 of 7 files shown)


src/personal.js
```diff
@@ -38,6 +48,24 @@ $(function() {
 		})
 	}
 
+	PersonalSettings.prototype.updateZoteroAPIKey = function(key) {
+		const self = this
+		this._updateSetting({ zoteroAPIKeyInput: key }, function() {
+			self.zoteroAPIKeyInput.value = key
+		}, function() {
+
+		})
```
juliushaertl, line 57:
```suggestion
		this._updateSetting({ zoteroAPIKeyInput: key }, () => {
			this.zoteroAPIKeyInput.value = key
		}, (error) => {
			console.error('Failed to update the Zotero API key', error)
		})
```

Would also be great to add a call to https://www.npmjs.com/package/@nextcloud/dialogs `showError` to display a message to the user if this fails.

lib/Controller/SettingsController.php
```diff
@@ -297,6 +298,14 @@ public function setPersonalSettings($templateFolder) {
 				$status = 'error';
 			}
 		}
+		if ($zoteroAPIKeyInput !== null) {
+			try {
+				$this->config->setUserValue($this->userId, 'richdocuments', 'zoteroAPIKey', $zoteroAPIKeyInput);
```
juliushaertl, line 303: Is there a way we could verify that the API key is valid here already? Even if it is only used by online?

timar:
I also found a build error:
```
ERROR in /home/collabora/online-buildscripts/staging/builddir/richdocuments-nextcloud/src/personal.js
  27:1  error  Expected indentation of 3 tabs but found 4  indent
```

src/personal.js
lpranam: used showError :)

lib/Controller/SettingsController.php
lpranam: it can be done, but then i'll have to call the zotero API, which requires `api.zotero.org` to be added to connect-src for CSP

src/personal.js
```diff
@@ -38,6 +49,24 @@ $(function() {
 		})
 	}
 
+	PersonalSettings.prototype.updateZoteroAPIKey = function(key) {
+		const self = this
+		this._updateSetting({ zoteroAPIKeyInput: key }, function() {
+			self.zoteroAPIKeyInput.value = key
+		}, function(error) {
+			showError('Failed to update the Zotero API key')
```
timar, line 57: use the t() function for localization

lib/Controller/SettingsController.php
timar: Why? You'd call Zotero API from php not from the browser.

lpranam:
updated PR to consider capabilities

src/personal.js
```diff
@@ -38,6 +49,24 @@ $(function() {
 		})
 	}
 
+	PersonalSettings.prototype.updateZoteroAPIKey = function(key) {
+		const self = this
+		this._updateSetting({ zoteroAPIKeyInput: key }, function() {
+			self.zoteroAPIKeyInput.value = key
+		}, function(error) {
+			showError(t('Failed to update the Zotero API key'))
```
timar, line 57: t('richdocuments', 'Failed to update the Zotero API key')

timar (changes requested):

timar (approved):

juliushaertl (approved):
Looks good to me 👍

PR #2688, opened 2022-12-20
Log info message on wopi ip mismatch

juliushaertl:
Add some logging at info level to make debugging non-working allow lists easier

### Checklist


max-nextcloud (approved):
looks good to me. I can imagine that will make debugging much easier.

PR #2723, opened 2023-01-04
fix: Open pdf files by default on share links if files_pdfviewer is disabled

juliushaertl:
### Summary

Fixes an issue where pdf files were not opened in Collabra on public share links when the files_pdfviewer app is disabled. If the pdf mimetype is in the default list it should not be part of the optional mimetypes.

### Checklist

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -104,6 +104,7 @@ class Capabilities implements ICapability {
 		if (!$this->capabilities) {
 			$collaboraCapabilities = $this->capabilitiesService->getCapabilities();
 			$filteredMimetypes = self::MIMETYPES;
+			$optionalMimetypes = self::MIMETYPES_OPTIONAL;
 			// If version is too old, draw is not supported
 			if (!$this->capabilitiesService->hasDrawSupport()) {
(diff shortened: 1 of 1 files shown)


lib/Capabilities.php
```diff
@@ -114,13 +115,14 @@ public function getCapabilities() {
...
+				$optionalMimetypes = array_diff($optionalMimetypes, ['application/pdf']);
 			}
 
 			$this->capabilities = [
 				'richdocuments' => [
 					'version' => \OC::$server->getAppManager()->getAppVersion('richdocuments'),
-					'mimetypes' => $filteredMimetypes,
-					'mimetypesNoDefaultOpen' => self::MIMETYPES_OPTIONAL,
+					'mimetypes' => array_values($filteredMimetypes),
+					'mimetypesNoDefaultOpen' => array_values($optionalMimetypes),
```
marcelklehr, line 125: Hah, I hate having to remember this in PHP so much!

marcelklehr (approved):
Didn't test, but seems reasonable!

gokaysatir (approved):
Tested the patch, thanks for the fix :)

Issue #2697, opened 2022-12-21
Creating new document from template fails to open Collabora Online (NotPermittedException is logged)


kevinkuan1969:
Describe the bug
We can create any new document with all the templates except the global template added from the Settings-Adminstration-Office-Global Templates. 

To Reproduce
1. Go to File, select + to add new document  
2. Pick the global template from the selection 
3. Click create
4. Loading new document
5. Document loading failed, please try again later.

Expected behavior

Screenshots

![image](https://user-images.githubusercontent.com/80019930/208930193-7a2ba001-8aa9-4ec7-9607-3f4549138287.png)

Client details:
 - OS - MS Windows 11
 - Browser - MS Edge 
 - Version 108.0.1462.54 (Official build) (64-bit)
 - Device: Desktop

Server details

Operating system: Ubuntu 20.14

Web server: 

Database: MySQL 8.0.31

PHP version: 8.1.13

Nextcloud version: 24.0.7

Version of the Collabora Online - Build-in CODE server: 22.5.802

Version of Nextcloud Office: 6.3.2

<details>
<summary>Logs</summary>

Cannot find any error in the log file and will provide as request

juliushaertl:
Any chance you could share the template file you used and the collabora server logs ?

kevinkuan1969:
![image](https://user-images.githubusercontent.com/80019930/208938228-8e74b958-30ec-4524-9c6c-8e3461a8fb8c.png)
Using the Build-in Code and did not install Collabora Online to any server.

Here the sample of template generated from MS Office 2019

kevinkuan1969:
The sample of template to upload to global template in Nextcloud Office
[Global Template - Realestate.odt](https://github.com/nextcloud/richdocuments/files/10278924/Global.Template.-.Realestate.odt)

kevinkuan1969:
![image](https://user-images.githubusercontent.com/80019930/209305105-3bb07e17-b2f7-4da2-a5b4-fb904736d0b8.png)

Instead of using build-in CODE, i have setup up the new Collaboration Online server hosted the same server. All works perfectly except when select the global template. It failed to load but it works after second attempt.
 
No error found in collabora_error.log file but the the error found in nextcloud.log:

{"reqId":"1mC1toFJdUJeYlcjhHKH","level":3,"time":"2022-12-23T08:40:14+00:00","remoteAddr":"127.0.0.1","user":"--","app":"richdocuments","method":"GET",
[...]

PR #2740, opened 2023-01-10
Fix creating new files from user templates

juliushaertl:
- Add a test case for the failure described in https://github.com/nextcloud/richdocuments/issues/2697
- Fix it by making sure that the WOPIMiddleware also allows to return the template


cypress/e2e/templates.spec.js
```diff
@@ -0,0 +1,64 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+describe('Create new office files from templates', function() {
+
+	let randUser
+	before(function() {
+		cy.createRandomUser().then(user => {
+			randUser = user
+			cy.login(user)
```
max-nextcloud, line 25: no need to login here i think.

cypress/e2e/templates.spec.js
```diff
@@ -0,0 +1,64 @@
...
+		})
+	})
+	beforeEach(function() {
+		cy.login(randUser)
+		cy.createFolder(randUser, 'Templates-user')
+		cy.uploadFile(randUser, 'templates/presentation.otp', 'application/vnd.oasis.opendocument.presentation', '/Templates-user/presentation.otp')
+		cy.setPersonalTemplateFolder(randUser, '/Templates-user')
+		cy.login(randUser)
```
max-nextcloud, line 33: Do you need to login again here?

cypress/support/commands.js
```diff
@@ -31,6 +31,24 @@ Cypress.Commands.add('logout', (route = '/') => {
 	})
 })
 
+Cypress.Commands.add('createFolder', (user, target) => {
+	cy.login(user)
```
max-nextcloud, line 35: I'd just assume you're logged in here.

cypress/e2e/templates.spec.js
```diff
@@ -0,0 +1,64 @@
...
+	before(function() {
+		cy.createRandomUser().then(user => {
+			randUser = user
+			cy.login(user)
+		})
+	})
+	beforeEach(function() {
+		cy.login(randUser)
+		cy.createFolder(randUser, 'Templates-user')
+		cy.uploadFile(randUser, 'templates/presentation.otp', 'application/vnd.oasis.opendocument.presentation', '/Templates-user/presentation.otp')
```
max-nextcloud, line 31: If you had multiple `it` blocks in this `describe` you'd try to create this folder multiple times. Maybe move into the `before` or the `it` block?

max-nextcloud (approved):

cypress/e2e/templates.spec.js
max-nextcloud: I guess it's because uploadFile will log you out. I think we have an uploadFile in text cypress that uses the current login and preserves it: https://github.com/nextcloud/text/blob/master/cypress/support/commands.js#L55-L76

juliushaertl: Ah thanks for the pointer, let me adapt to that the to avoid clearing.

cypress/support/commands.js
juliushaertl: I'd keep this for now to avoid that the command call could interfere with previously called logins. Feels like those could be harder to debug and I think the additional check shouldn't hurt too much.

nevertheless those should go into the lib at some point

PR #2745, opened 2023-01-12
Load viewer scripts depending on share ownership

juliushaertl:
- refactor: Use existing method to check if users can edit
- fix: Do not consider guests to be in groups
- fix: Load viewer depending on share owners on public links
- fix: Expose capabilities also for guests

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


juliushaertl:
e2e tests for this would also be nice to have, but require some more effort. I'll try to work on that on the sides a bit but would be good to get this in already to fix this issue.

lib/PermissionManager.php
```diff
@@ -64,7 +64,7 @@ private function userMatchesGroupList(?string $userId = null, ?array $groupList
 		}
 
 		if ($userId === null) {
-			return true;
+			return false;
```
julien-nc, line 67: Maybe add an explicit comment there to explain how unauthenticated users are handled.

julien-nc (changes requested):
Did not try it but looks good.

lib/PermissionManager.php
juliushaertl: Definitely, added :)

julien-nc (approved):
:+1:

PR #2767, opened 2023-01-23
Avoid exception when uploading epub

eszkadev:
When Collabora Online exported epub file to the storage WopiController tried to get access token what fails because in discovery there is no epub entry (Collabora Online cannot open epub files).
Export functionality doesn't require url in response at all as it doesn't open newly uploaded file. Let's just return status OK as upload was successful.

This is related to "export to storage" functionality: https://github.com/CollaboraOnline/online/pull/5738


timar (approved):

juliushaertl (approved):

juliushaertl:
Rebased so we have CI passing again and adjusted the code style for parenthesis on if-statements to match our standard.

PR #2782, opened 2023-02-01
allow to add ttf files

szaimen:
See https://github.com/nextcloud/server/blob/37bb33c5799b834dfef3fb73936bd0e5a4773fb8/resources/config/mimetypemapping.dist.json#L192

### DIFF
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -38,6 +38,7 @@ class SettingsController extends Controller {
 	// TODO adapt overview generation if we add more font mimetypes
 	public const FONT_MIME_TYPES = [
 		'font/ttf',
+		'application/font-sfnt',
 		'font/opentype',
 		'application/vnd.oasis.opendocument.formula-template',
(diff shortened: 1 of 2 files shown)


szaimen:
BTW: it seems like the api used is unreliable:
image
See https://developer.mozilla.org/en-US/docs/Web/API/File/type

juliushaertl (approved):

juliushaertl:
We could detect this on the server side by reading the file with https://github.com/nextcloud/server/blob/f5c361cf44739058b79f322576a1bad2d8c142d9/lib/private/Files/Type/Detection.php#L226 but let's get this in as it seems like a reasonable first step.

szaimen:
> We could detect this on the server side by reading the file with https://github.com/nextcloud/server/blob/f5c361cf44739058b79f322576a1bad2d8c142d9/lib/private/Files/Type/Detection.php#L226 but let's get this in as it seems like a reasonable first step.

sounds good 👍

PR #2814, opened 2023-02-20
Integrate the link picker

### LINKED ISSUE #2813: Link picker
Based on https://github.com/CollaboraOnline/online/pull/5852 we should:

- [x] Extend checkFileInfo to set the feature flag enabled
- [x] Listen for the post message to show the link picker dialog and emit the result to Collabora
- [x] Listen for the event to fetch the link preview metadata and emit the results to Collabora

julien-nc:
closes #2813

* [x] Extend checkFileInfo to set the feature flag enabled
* [x] Listen for the post message to show the link picker dialog and emit the result to Collabora
* [x] Listen for the event to fetch the link preview metadata and emit the results to Collabora

Tested without using a CODE server which supports the feature. Tested by reacting to `UI_Mention` messages. Worked fine :grin:.

Problems:
* [x] The link picker does not get the provider list even if `RenderReferenceEvent` is dispatched (I checked it really is). Could it be that a script is loaded too early?
* [x] I'm not so sure with the base64 encoding of the link thumbnail.
* [ ] Should we check the mimetype of the thumbnail request response? If so, we need to adjust `/core/references/preview/HASHED_REF` to provide a mimetype, for now, `application/octet-stream` is returned.

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -223,6 +223,7 @@ class WopiController extends Controller {
 			'DownloadAsPostMessage' => $wopi->getDirect(),
 			'SupportsLocks' => $this->lockManager->isLockProviderAvailable(),
 			'IsUserLocked' => $this->permissionManager->userIsFeatureLocked($wopi->getEditorUid()),
+			'EnableRemoteLinkPicker' => (bool)$wopi->getCanwrite() && !$isPublic && !$wopi->getDirect(),
 		];
 
 		$zoteroAPIKey = $this->config->getUserValue($wopi->getEditorUid(), 'richdocuments', 'zoteroAPIKey', '');
--- a/src/document.js
+++ b/src/document.js
@@ -660,7 +660,16 @@ const documentsMain = {
 	async openLinkPicker() {
 		try {
 			const link = await getLinkWithPicker(null, true)
-			PostMessages.sendWOPIPostMessage('loolframe', 'Action_InsertLink', { url: link })
+			try {
+				const url = new URL(link)
+				if (url.protocol === 'http:' || url.protocol === 'https:') {
+					PostMessages.sendWOPIPostMessage('loolframe', 'Action_InsertLink', { url: link })
+					return
+				}
+			} catch (e) {
+				console.debug('error when parsing the link picker result')
+			}
+			PostMessages.sendWOPIPostMessage('loolframe', 'Action_Paste', { Mimetype: 'text/plain', Data: link })
 		} catch (e) {
 			showError(t('richdocuments', 'Failed to get a link with the picker'))
 			console.error('Link picker promise rejected :', e)
--- a/src/document.js
+++ b/src/document.js
@@ -697,7 +697,7 @@ const documentsMain = {
 					console.error('Error loading the reference image', e)
 				}
 			}
-			PostMessages.sendWOPIPostMessage('loolframe', ' Action_GetLinkPreview_Resp', { url, title, image: b64Image })
+			PostMessages.sendWOPIPostMessage('loolframe', 'Action_GetLinkPreview_Resp', { url, title, image: b64Image })
 		} catch (e) {
 			showError(t('richdocuments', 'Failed to get the link preview'))
 			console.error('Error resolving a reference', e)
--- a/src/document.js
+++ b/src/document.js
@@ -686,18 +686,23 @@ const documentsMain = {
 			const resolvedLink = result.data.ocs.data.references[url]
 			const title = resolvedLink?.openGraphObject?.name
 			const thumbnailUrl = resolvedLink?.openGraphObject?.thumb
-			let b64Image = null
 			if (thumbnailUrl) {
 				try {
-					const imageResponse = await axios.get(thumbnailUrl)
+					const imageResponse = await axios.get(thumbnailUrl, { responseType: 'blob' })
 					if (imageResponse?.status === 200 && imageResponse?.data) {
-						b64Image = btoa(unescape(encodeURIComponent(imageResponse.data)))
+						const reader = new FileReader()
+						reader.addEventListener('loadend', (e) => {
+							const b64Image = e.target.result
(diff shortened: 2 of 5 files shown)


lib/Controller/WopiController.php
```diff
@@ -223,6 +223,7 @@ public function checkFileInfo($fileId, $access_token) {
 			'DownloadAsPostMessage' => $wopi->getDirect(),
 			'SupportsLocks' => $this->lockManager->isLockProviderAvailable(),
 			'IsUserLocked' => $this->permissionManager->userIsFeatureLocked($wopi->getEditorUid()),
+			'EnableRemoteLinkPicker' => (bool)$wopi->getCanwrite(),
```
juliushaertl, line 226: I think the link picker should for now also be disabled for public links and direct editing (from the mobile clients)

src/document.js
```diff
@@ -648,6 +656,43 @@ const documentsMain = {
...
+					if (imageResponse?.status === 200 && imageResponse?.data) {
+						b64Image = btoa(unescape(encodeURIComponent(imageResponse.data)))
+					}
+				} catch (e) {
+					console.error('Error loading the reference image', e)
+				}
+			}
+			PostMessages.sendWOPIPostMessage('loolframe', ' Action_GetLinkPreview_Resp', { url, title, image: b64Image })
+		} catch (e) {
+			console.error('Error resolving a reference', e)
```
juliushaertl, line 692: Can we show a toast also here so the user knows that something failed?

src/document.js
```diff
@@ -648,6 +656,43 @@ const documentsMain = {
 		PostMessages.sendWOPIPostMessage('loolframe', 'Action_Mention', { list })
 	},
 
+	openLinkPicker() {
+		getLinkWithPicker(null, true)
```
juliushaertl, line 660: I'd also switch to async await here ;)

lib/Listener/LoadViewerListener.php
```diff
@@ -56,6 +63,7 @@ public function handle(Event $event): void {
 		}
 		if ($this->permissionManager->isEnabledForUser() && $this->userId !== null) {
 			$this->initialStateService->provideCapabilities();
+			$this->eventDispatcher->dispatchTyped(new RenderReferenceEvent());
```
juliushaertl, line 66: document.js is loaded in an iframe already, so triggering the event rather ehe should work: https://github.com/nextcloud/richdocuments/blob/e3105a17e6b9f4baa4182f97d05f7b6f39e4d7a9/lib/Controller/DocumentTrait.php#L15

This is how the frames are currently structured, I've been thinking for a while if we shouldn't refactor that at some point but that may be larger work:

- Nextcloud loads viewer.js
  - The Office.vue component loads an iframe rendering a simple template through the DocumentController with document.js
    - document.js submits the wopi data to Collabora within another iframe

juliushaertl:
> * Should we check the mimetype of the thumbnail request response? If so, we need to adjust `/core/references/preview/HASHED_REF` to provide a mimetype, for now, `application/octet-stream` is returned.

I'd say this is fine as we already validate before storing the image in https://github.com/nextcloud/server/blob/f0589523c80749b1bdfbbfea34b141cd562e216b/lib/private/Collaboration/Reference/LinkReferenceProvider.php#L163

lib/Controller/WopiController.php
julien-nc: True, not sure about the new conditions.

src/document.js
julien-nc: :+1:

src/document.js
julien-nc: :+1:

lib/Listener/LoadViewerListener.php
julien-nc: :heart:

julien-nc:
@juliushaertl Do you mean detecting this without changing the link picker implementation? Just by checking if the picker result is a URL or not? Seems good to me.

julien-nc:
@juliushaertl Done, it was fun to try it with the GPT-3 provider.

juliushaertl:
Nice, will do a full test later with the collabora branch :)

src/document.js
```diff
@@ -648,6 +657,53 @@ const documentsMain = {
...
+				try {
+					const imageResponse = await axios.get(thumbnailUrl)
+					if (imageResponse?.status === 200 && imageResponse?.data) {
+						b64Image = btoa(unescape(encodeURIComponent(imageResponse.data)))
+					}
+				} catch (e) {
+					console.error('Error loading the reference image', e)
+				}
+			}
+			PostMessages.sendWOPIPostMessage('loolframe', ' Action_GetLinkPreview_Resp', { url, title, image: b64Image })
```
eszkadev, line 700: There is a type here ' Action...' <- space as first character

eszkadev:
what is expected thumbnail value format you send?
for me best would be something I can put in img.src without processing.
Like: `img.src = 'data:' + receivedImage`

julien-nc:
@eszkadev Thanks for the review. There you go. It is now correctly requesting the image and getting a valid base64 string out of it.

juliushaertl (approved):
Tested and works. Thanks a lot @julien-nc

eszkadev:
Thanks a lot! Tested and works with website URLs, but with picked link to the board from deck app I got `data:application/octet-stream;base64,PHN2ZyB4b...` and it cannot be displayed by the browser

juliushaertl:
We'll have to address this separately in server

julien-nc:
@juliushaertl Actually I found a fix for that. It's true our reference preview endpoint sometimes does not return the correct mimetype. But for images like described by @eszkadev, the response mimetype is actually correct. The issue is the way we build the Blob to produce the base64 string. I'll make another PR.

pedropintosilva:
one more think, I noticed that typing a url with the protocol (http or https) doesn't work. Can we, when that is not entered by the user assume the protocol ?

PR #2815, opened 2023-02-20
ci: Pin phpunit version

juliushaertl:
Make sure CI runs properly again


marcelklehr:
What about main?

juliushaertl:
Main uses the template phpunit workflows which installs it through composer. There is passes and didn't automatically use phpunit 10: https://github.com/nextcloud/richdocuments/actions/runs/4202886596/jobs/7291598842

PR #2826, opened 2023-03-01
Fix base64 image generation

julien-nc:
Axios directly returns a blob that can be directly passed to the `FileReader`. This way the mimetype does not get lost in the base64 result.

I didn't spot this because having a `application/octet-stream` mimetype for PNG or JPEG images works fine when used as src of ``. But for SVG it only works if the correct mimetype is specified in the `data:` string.


eszkadev (approved):
Yes, now it works in both cases :)

juliushaertl (approved):

PR #2831, opened 2023-03-03
fix: Avoid error when federation is disabled

juliushaertl:
Moving code to only be called if federation is enabled. This is already fixed on newer stable branches/master during refactoring.


juliushaertl:
Node failure unrelated.

juliushaertl:
Already got confirmed that this works on the reporting instance.

PR #2845, opened 2023-03-09
fix(auto_logout): Register auto_logout listeners to prevent auto_logout while editing

marcelklehr:
* Resolves: -
* Target version: main

### Summary
When the collabora frame is visible the core auto_logout mechanism in session-heartbeat.js isn't tickled anymore and logs the user out after session_lifetime even if there has been activity.

### Checklist


juliushaertl (approved):

juliushaertl:
Makes sense, thanks for that :)

PR #2846, opened 2023-03-10
fix: Trigger initial setup earlier

juliushaertl:
### Summary

There is no need to wait for the full ready event and DOMContentLoaded helps to not run into problems when a file opening is triggered before.


marcelklehr (approved):
:elephant:

PR #2860, opened 2023-03-13
Add modal to set zotero API key during editing

### LINKED ISSUE #2764: Zotero improvements
- Send empty api key if none is configured so collabora knows that we support it https://github.com/nextcloud-releases/richdocuments/blob/db1582f937e2cde968e17eecfa1f435c3811f371/lib/Controller/WopiController.php#L228
- Listen to the post message for showing a dialog
  - Collabora will send a post message `UI_ZoteroKeyMissing` as per https://github.com/CollaboraOnline/online/pull/5703 if there was no key provided and the user clicked the button
    - We should catch this post message and show a modal as shown below

juliushaertl:
- feat: Add zotero admin setting and enable integration by default
- feat(Zotero): Add modal to set API key

Fixes #2764 

Cannot fully test right now as I'm on the train with outdated CODE container, but should work.

@pedropintosilva Currently we reload the iframe after setting the API key, but would probably be a good enhancement to also have a post message to pass back the api key to the current editing session, to make sure the user can continue just where they left.

### DIFF
--- a/src/components/Modal/ZoteroHint.vue
+++ b/src/components/Modal/ZoteroHint.vue
@@ -20,7 +20,7 @@
   -->
 
 <template>
-	<NcModal>
+	<NcModal :show="show" @close="close">
 		<div class="zotero-hint">
 			<h2>{{ t('richdocument', 'Link to your Zotero library') }}</h2>
 			<BookOpenPageVariantOutline :size="96" />
@@ -58,6 +58,12 @@ export default {
 		NcButton,
 		NcTextField,
 	},
+	props: {
+		show: {
+			type: Boolean,
+			default: false,
+		},
+	},
 	emits: ['submit'],
 	data() {
 		return {
@@ -75,6 +81,10 @@ export default {
 				showError(t('richdocuments', 'Failed to set Zotero API key'))
 			}
 			this.$emit('submit')
+			this.close()
+		},
+		close() {
+			this.$emit('update:show', false)
 		},
 	},
 }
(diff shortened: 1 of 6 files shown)


pedropintosilva (changes requested):
Awesome!

we need to tweak our onUpdateViews function (@eszkadev is on it)

Nevertheless this popup ended up having a direct input field to set api key (and not just a button to settings office page) which makes sense but that string never gets to our side , not until someone reloads the page. So probably the best would to reload the page on your side once the user sets the apikey in that popup. What do you think?

pedropintosilva:
apart from the need to refresh, this PR works as expected. Thanks

eszkadev:
I've noticed that when closing popup using 'x' icon - dark overlay still is visible and blocks user from any interaction with Collabora.

eszkadev:
PR on the Collabora side related to this: https://github.com/CollaboraOnline/online/pull/6033

juliushaertl:
Reloading should now also work. We reload the full iframe for now. I could imagine having a way to just refetch the checkFileInfo response would also be suitable here (and could also have other use cases, e.g. when permissions change) but this is as good as it gets for now I'd say.

src/components/Modal/ZoteroHint.vue
```diff
@@ -0,0 +1,105 @@
...
+		}
+	},
+	methods: {
+		async submit() {
+			try {
+				await savePersonalSetting({
+					zoteroAPIKeyInput: this.apiKey,
+				})
+			} catch (e) {
+				console.error('Failed to set zotero api key', e)
```
luka-nextcloud, line 79: I think we need to show alert here to notify user in case save error?

juliushaertl: Fixed

juliushaertl:
WOPI failures are unrelated https://github.com/nextcloud/richdocuments/issues/2896

luka-nextcloud (approved):

PR #2877, opened 2023-03-27
Fix file templates on public links

juliushaertl:
### Summary

Bring back the template picker on public share links of folders for file creation when the admin has setup global templates.

Furthermore this adds a app config value to disable this in case templates shall not be leaked to anonymous users and fixes missing draw file creation for share links as well.


juliushaertl:
Added a cypress test as well

luka-nextcloud (approved):

PR #2881, opened 2023-03-30
fix(files): Avoid reloading the file list and use update methods instead

juliushaertl:
### Summary

Using a file list reload causes quite some troubles like changing the URL and also has a performance impact of loading all file metadata again. We can actually just update the affected file rows during rename, save and save as.

### TODO

- [x] Add cypress tests: Time change related topics are hard to test on e2e reliably, so let's skip this for now
- [x] File bug at collabora that on rename the old filename shows up again and there is no loading indication about the operation. At some point the filename then changes to the expected new one. https://github.com/CollaboraOnline/online/issues/6131

### Checklist

### DIFF
--- a/src/view/FilesAppIntegration.js
+++ b/src/view/FilesAppIntegration.js
@@ -91,10 +91,8 @@ export default {
 			return
 		}
 
-		if (this.getFileList()) {
-			this.getFileList().setViewerMode && this.getFileList().setViewerMode(false)
-			this.getFileList().reload && this.getFileList().reload()
-		}
+		this.updateFileInfo(undefined, Date.now())
+
 		this.fileModel = null
 		if (!isPublic) {
 			this.removeVersionSidebarEvents()
@@ -103,6 +101,8 @@ export default {
 	},
 
 	saveAs(newName) {
+		const oldFile = this.getFileModel()
+
 		if (this.handlers.saveAs && this.handlers.saveAs(this)) {
 			return
 		}
@@ -112,8 +112,12 @@ export default {
 		}
 
 		if (this.getFileList()) {
+			const newFileModel = oldFile.clone()
+			newFileModel.set('name', newName)
+			newFileModel.set('mtime', Date.now())
 			this.getFileList()
-				.reload()
+				.add(newFileModel.toJSON())
+
 			OC.Apps.hideAppSidebar()
 		}
 	},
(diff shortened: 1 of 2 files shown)


juliushaertl:
@skjnldsv Just for some awareness for the files2vue project, we have this use case where the viewer interaction would want to update the underlying file list on save/save as/rename. Is that something considered already? I guess we could use events for that in the future.

skjnldsv:
> @skjnldsv Just for some awareness for the files2vue project, we have this use case where the viewer interaction would want to update the underlying file list on save/save as/rename. Is that something considered already? I guess we could use events for that in the future.

Yes, I started doing that here: https://github.com/nextcloud/server/blob/5024f295dcfdf850d2d95f419bcf122bf08be5c8/apps/files/src/components/NavigationQuota.vue#L81-L91
You can use the following:

```js
subscribe('files:file:created', () => {})
subscribe('files:file:deleted',  () => {})
subscribe('files:file:moved',  () => {})
subscribe('files:file:updated',  () => {})

subscribe('files:folder:created',  () => {})
subscribe('files:folder:deleted',  () => {})
subscribe('files:folder:moved',  () => {})
subscribe('files:folder:updated',  () => {})
```

src/view/FilesAppIntegration.js
```diff
@@ -91,10 +91,8 @@ export default {
 			return
 		}
 
-		if (this.getFileList()) {
-			this.getFileList().setViewerMode && this.getFileList().setViewerMode(false)
-			this.getFileList().reload && this.getFileList().reload()
-		}
+		this.updateFileInfo(undefined, Math.floor(Date.now()))
```
luka-nextcloud, line 94: How about `Date.now()` instead of `1Math.floor(Date.now())`? They return same value (integer).

src/view/FilesAppIntegration.js
```diff
@@ -576,4 +582,20 @@ export default {
...
+		}
+
+		if (name) {
+			fileInfo.set('name', name)
+		}
+
+		if (mtime) {
+			fileInfo.set('mtime', mtime)
+		}
+		this.getFileModel().trigger('change', this.getFileModel())
```
luka-nextcloud, line 598: I think we can use `fileInfo.trigger('change', fileInfo)` if I am not mistaken.

juliushaertl:
WOPI failures are unrelated https://github.com/nextcloud/richdocuments/issues/2896

PR #2948, opened 2023-05-12
Use the color-primary-element* variables

szaimen:
Explanation: the color-primary variables are not to be used in components because the introduce problems with high-contrast primary colors. Fix this by using the primary-element variables instead.

### DIFF
--- a/src/helpers/coolParameters.js
+++ b/src/helpers/coolParameters.js
@@ -53,10 +53,9 @@ const getCollaboraTheme = () => {
 const generateCSSVarTokens = () => {
 	/* NC versus COOL */
 	const cssVarMap = {
-		'--color-primary-text': '--co-primary-text',
+		'--color-primary-element-text': '--co-primary-text',
 		'--color-primary-element': '--co-primary-element:--co-text-accent',
-		'--color-primary-light': '--co-primary-light',
-		'--color-primary-element-light': '--co-primary-element-light',
+		'--color-primary-element-light': '--co-primary-light:--co-primary-element-light',
 		'--color-error': '--co-color-error',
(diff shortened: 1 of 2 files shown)


src/helpers/coolParameters.js
```diff
@@ -46,9 +46,9 @@ const getCollaboraTheme = () => {
 const generateCSSVarTokens = () => {
 	/* NC versus COOL */
 	const cssVarMap = {
-		'--color-primary-text': '--co-primary-text',
+		'--color-primary-element-text': '--co-primary-text',
 		'--color-primary-element': '--co-primary-element:--co-text-accent',
-		'--color-primary-light': '--co-primary-light',
+		'--color-primary-element-light': '--co-primary-light',
```
szaimen, line 51: not sure about these

juliushaertl: Will need to double check where they are actually used within Collabora

szaimen: Did you already check with them? :)

juliushaertl: Not yet, but with 27 and Collabora 25.05 we actually pass over the values directly and make use of the element ones mostly in the branding directly. Still on my todo list for the older versions.

juliushaertl (approved):

src/helpers/coolParameters.js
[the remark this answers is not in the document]
szaimen:
```suggestion
```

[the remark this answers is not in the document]
juliushaertl: Yep, just fixed :)

szaimen:
So merge or is the failure related?

PR #2963, opened 2023-05-21
Fix plugin watermark for php 8.1

AIlkiv:
* Target version: main

### Summary

From PHP 8.1 PDO DB integer field return as PHP integer type. https://www.php.net/manual/en/migration81.incompatible.php#migration81.incompatible.pdo.mysql

This caused a bug in "in_array($tagId, $tags, true)" in PermissionManager::shouldWatermark.

Fixed and added tests.

### Checklist

- [] Code is properly formatted
- [] Sign-off message is added to all commits
- [] Documentation (manuals or wiki) has been updated or is not required


blizzz:
@AIlkiv thank you for your contribution! Could you check the failing ci tests? The linter should be satisifed by "composer run cs:fix" and committing the results. Also the static analyyzer disagrees with the return type declaration, which is strictly speaking true, while not intended. The method could be called with a key that does not resolve to an array  setting and return a string.

AIlkiv:
> @AIlkiv thank you for your contribution! Could you check the failing ci tests? The linter should be satisifed by "composer run cs:fix" and committing the results. Also the static analyyzer disagrees with the return type declaration, which is strictly speaking true, while not intended. The method could be called with a key that does not resolve to an array setting and return a string.

Thanks. Fixed.

PR #2964, opened 2023-05-25
fix: Pass over dark/light variants of our css variables

juliushaertl:
Pass over plain versions of our css variables in dark/light variant so they can be reused for Collabora dark mode to adapt to our way of color tinting.

Required for dark mode compatibility with Collabora 23.05


luka-nextcloud (approved):

max-nextcloud (approved):
I had a look at the code and saw nothing obvious. Have not tried it locally though.

I don't quite understand the logic for the invert hack... but I trust you on that one. Please let me know if you need a more in depth review.

pedropintosilva (approved):

PR #2977, opened 2023-06-01
ci: Switch to ghcr.io images

juliushaertl:
- [x] Wait for dispatch of phpunit- actions from templates
- [x] collabora/code image is not on ghcr.io, asked if they would be open for having that in their repo
  - [x] For now switched to https://github.com/juliushaertl/nextcloud-docker-dev/pkgs/container/nextcloud-dev-code which is nightly built and a copy of collabora/code
- [x] cypress uses alpine for getting the docker ip

### Checklist

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -98,6 +98,7 @@ jobs:
           php occ user:add --password-from-env user1
           php occ user:add --password-from-env user2
           php occ config:system:set force_language --value en
+          php occ app:enable --force testing
           php occ app:enable --force viewer
           php occ app:enable --force richdocuments
           php occ app:list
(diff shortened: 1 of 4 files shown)


.github/workflows/cypress.yml
```diff
@@ -34,7 +34,7 @@ jobs:
 
     services:
       collabora:
-        image: collabora/code
+        image: ghcr.io/juliushaertl/nextcloud-dev-code:latest
```
blizzz, line 37: does it stay in your namespace or will it hop over eventually?

blizzz (approved):

.github/workflows/cypress.yml
juliushaertl: As mentioned in the pr description ideally this would be a official Collabora image on ghcr but for now this is a mirror with nightly updates

Issue #2933, opened 2023-04-30
Opening a recommended document with Nextcloud Office and closing it leaves the screen blank


RuudschMaHinda:
Since I don't know if it is an issue with Recommendations, or Nextcloud Office implementation I am duplicating this bug report here:
https://github.com/nextcloud/recommendations/issues/604
(I copy pasted the title from there.)

In short:
Open an Office Document from the Recommendations bit up top, close it, and it only shows the background and does not return back to file view.
When opening any Office Document normally from file view, this issue does not happen and everything works expected.

If this is not an issue with the Recommendations App, then it might have something to do with how closing a document reports back to the browser?

Screenshot and two different Webservers (nginx and apache) and two different PHP-Versions (8.1 and 8.2) both show the same issue.

Edit:
Opening from Dashboard from the recommended files works without issue. Maybe it's a Files issue?

SergeyMosin:
Same here , NC v25.0.6 ,  Nextcloud Office v7.1.4 , Collabora Office v22.05.15.2 

Step to reproduce:
1. Open an office document from **RECOMMENDED** area
2. Edit the doc
3. Close the document
4. Blank screen :boom: :fire: :bug: :lady_beetle: 


Error: when opening a doc from recommended:
```
Error during post message handler 
Object { msgId: "App_LoadingStatus", args: {…}, deprecated: false }
 TypeError: model is null
    _updateDetailsView https://my.domain.com/js/files/merged-index.js?v=e1b8d193-0:4827
    getFileModel https://my.domain.com/apps/richdocuments/js/richdocuments-viewer.js?v=e1b8d193-0:2
    initAfterReady https://my.domain.com/apps/richdocuments/js/richdocuments-viewer.js?v=e1b8d193-0:2
    postMessageHandler https://my.domain.com/apps/richdocuments/js/richdocuments-viewer.js?v=e1b8d193-0:2
    handlePostMessage https://my.domain.com/apps/richdocuments/js/richdocuments-viewer.js?v=e1b8d193-0:2
    handlePostMessage https://my.domain.com/apps/richdocuments/js/richdocuments-viewer.js?v=e1b8d193-0:2
    e https://my.domain.com/apps/richdocuments/js/richdocuments-viewer.js?v=e1b8d193-0:2
    e https://my.domain.com/apps/richdocuments/js/richdocuments-viewer.js?v=e1b8d193-0:2
[...]

PR #3010, opened 2023-06-11
fix: Catch and skip errors for adjusting the files view when the file is not present in the list

juliushaertl:
* Resolves: #2933
* Target version: main

### Summary

When opening a file from the recommended files there is no need to update it in the file list so we can safely ignore the error in this case.

Handling file updates is quite hacky but hopefully gets a better way to handle once the files to vue migration moves on.


datenangebot (approved):
Looks reasonable

PR #3015, opened 2023-06-12
Set the default language of the template to English (USA).

gokaysatir:
Signed-off-by: Gökay Şatır <gokaysatir@gmail.com>

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


gokaysatir:
@juliushaertl hello, below commit has wrong default language for the template files. I set the default to English (USA) in this commit.
https://github.com/nextcloud/richdocuments/pull/3000

Sorry for the forgotten bit.

juliushaertl:
Weren't the templates language agnostic so they could pick up the users language in the past? Is this now forcing to be English (USA)?

gokaysatir:
> Weren't the templates language agnostic so they could pick up the users language in the past? Is this now forcing to be English (USA)?

Yes that's right. But there is a minor thing to improve there. Nextcloud seems to use the document.odt file (not a template) as a template. There is also a template file - document.ott inside the "emptyTemplates" folder but it doesn't seem to be used.

So i updated the used one (with the *.odt extension) and forgot to update the language. A better fix would be using the "document.ott" template file.

juliushaertl:
Ah, that makes sense indeed. I'll check what changed there that the ott is no used.

juliushaertl (approved):

Issue #2996, opened 2023-06-07
Cannot open uploaded docx file when using object storage and encryption


max-nextcloud:
**Describe the bug**
Opening an uploaded docx file fails. Error message says it could not be decrypted and I should request the share to be reshared.

**To Reproduce**
1. Use latest Nextcloud
2. Enable object storage and encryption
3. Upload a docx file
4. Try to open the file
5. See error

**Expected behavior**
It should open the file fine.

**Screenshots**

**Client details:**
 - Browser [e.g. chrome, safari] Chrome
 - Device: [e.g. iPhone6, desktop]

## Server details

**Operating system**: nextcloud-docker-dev

juliushaertl:
- WOPI getFile fails with `Cannot decrypt this file, probably this is a shared file. Please ask the file owner to reshare the file with you.`
- Works fine on stable24-stable26
- Looks like it could be caused by server changes as reverting the one possible related line in WopiController didn't fix

mschoeneberger:
We lost access to several files before we realized the impact of this bug. It takes just the attempt of the file owner to open a Office file to corrupt it in a way which leaves it in a undecryptable state. We tried to recover the encryptionkeys from a backup, which would make the file readable by other users again (the owner still breaks it). After implementing the change proposed in #3028, the owner could open their files again. But if another user now made a change to these files these files were corrupted again now without being recoverable  by restoring their encryptionkeys (to my current knowledge).

Is currently a workaround available or a way to restore files lost by this ?

EukalyptusX:
Somehow, the files I tried to open, will be destroyed after "opening". I can't even download the files anymore - just nothing happens. But just the files, I tried to open.
I guess it's somehow related. Will I be able to download them after the planned merge? Or are the files irreversible destroyed?

karlemilnikka:
Now that Nextcloud AIO automatically upgrades the Nextcloud container to 27, more users are experiencing this issue.

https://help.nextcloud.com/t/cannot-decrypt-this-file-probably-this-is-a-shared-file-please-ask-the-file-owner-to-reshare-the-file-with-you/166856/

FlattusBlastus:
Maybe related to this?
Clean install, all IPv6 removed, latest:
2023-08-20 13:02:12 wsd-00001-00001 2023-08-20 13:02:12.881243 -0700 [ coolwsd ] TRC  Have 1 new children.| wsd/COOLWSD.cpp:5722
[...]

PR #3028, opened 2023-06-20
fix: Remove broken way of updating the encryption access list

juliushaertl:
Fix server side encryption as we actually access the file in the user scope with the system key these days.

This broke with 27 as there has been some changes to how the file keys are managed in https://github.com/nextcloud/server/pull/37243

* Resolves: https://github.com/nextcloud/richdocuments/issues/2996
* Target version: main

### Checklist

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -199,16 +199,6 @@ class DocumentController extends Controller {
 				}
 			}
 
-			$encryptionManager = \OC::$server->getEncryptionManager();
-			if ($encryptionManager->isEnabled()) {
-				// Update the current file to be accessible with system public shared key
-				$owner = $item->getOwner()->getUID();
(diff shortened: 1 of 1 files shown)


blizzz:
> Fix server side encryption as we actually access the file in the user scope with the system key these days.

Was there are an enforced migration towards systems keys?

juliushaertl:
Need to double check after https://github.com/nextcloud/server/pull/39447 was merged now

EukalyptusX:
Ist there a workaround I can implement by myself? It's a serious issue, that you can't even download the files.

juliushaertl:
Closing as during my testing I could no longer reproduce this and https://github.com/nextcloud/server/pull/39447 fixes the issue for me. 

If this is still an issue for anyone passing by please continue reading at https://github.com/nextcloud/richdocuments/issues/2996#issuecomment-1702546036

PR #3041, opened 2023-06-26
Let tests pass again reliably

juliushaertl:
We should clean this up but for now this makes them pass again


blizzz (approved):
:magic_wand: MySQL

PR #3042, opened 2023-06-28
fix(Wopi): Mark sensitive parameter as such

nickvergessen:
Ref https://github.com/nextcloud/server/pull/39040
This is only effective in PHP 8.2+ but it's the better and our future approach so doing both in parallel


blizzz (changes requested):
Also register via `IRegistrationContext::registerSensitiveMethods()` for PHP < 8.2.

nickvergessen:
Hmm so that means https://github.com/nextcloud/server/pull/39040 is actually not needed anymore?

blizzz:
> Hmm so that means [nextcloud/server#39040](https://github.com/nextcloud/server/pull/39040) is actually not needed anymore?

aye

blizzz (approved):

Issue #3029, opened 2023-06-20
Check why new empty files are not being created from templates


juliushaertl:
> Weren't the templates language agnostic so they could pick up the users language in the past? Is this now forcing to be English (USA)?

Yes that's right. But there is a minor thing to improve there. Nextcloud seems to use the document.odt file (not a template) as a template. There is also a template file - document.ott inside the "emptyTemplates" folder but it doesn't seem to be used.

So i updated the used one (with the *.odt extension) and forgot to update the language. A better fix would be using the "document.ott" template file.

_Originally posted by @gokaysatir in https://github.com/nextcloud/richdocuments/issues/3015#issuecomment-1590711002_

Issue #2983, opened 2023-06-03
Default blank document file is in US letter, all other ones in DIN A4


JMoVS:
**Describe the bug**
The default blank template .docx is in US letter format, but the default .ott and .odt files are all in DIN A4 - I think the format should be DIN A4 as default or at least allow an easy switch

**To Reproduce**
Create a new document, observe layout to be US letter.

**Expected behavior**
One has to manually adjust the template in the app directory to change that to DIN A4 - I hope it survives an update but I'm not even sure about that.

**Screenshots**

**Client details:**
 - Browser Firefox
 - Device: [e.g. iPhone6, desktop]

## Server details

**Operating system**:
Ubuntu
**Web server:**

**Database:**

**PHP version:**

**Nextcloud version:**
26.0.2
**Version of the richdocuments app**
8.0.2
**Version of Collabora Online**
latest CODE

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

JMoVS:
@juliushaertl Do you want me to create a pull request to fix this?

juliushaertl:
@pedropintosilva Can you tell if there is a way to make the template dynamically adjust the page size maybe depending on what is most common in the users region?

JMoVS:
happy to provide a pull request to make all consistent - I don't know how to solve the dynamic case though

juliushaertl:
This should also be addressed by https://github.com/nextcloud/richdocuments/pull/3059 @JMoVS If you are up for testing the patch, feedback is much appreciated :)

JMoVS:
earliest on the weekend - how could I test this? And can I test this on nextcloud 26?

bbx-github:
> The default blank template .docx is in US letter format

True - and if it is in DIN A4 it would solve my problem as well.

But `.docx` is not a template format. A template has `.dotx`. Such a file is missing in `richdocuments/emptyTemplates`. The naming scheme suggests a name `document.dotx` Using this template format may set the page format according to the locale? And is needed to make https://github.com/nextcloud/richdocuments/pull/3059 work for ooxml?

JMoVS:
@juliushaertl Not sure but right now, new documents are created in US letter again now that I upgraded to the latest version. I think this should be reopened

PR #3059, opened 2023-07-11
fix: Create empty files from template if available

juliushaertl:
Before this change empty files were always just created with a copy of the empty documents, which enforced US spell checking. With this we create a file from the present empty template files.

Fixes https://github.com/nextcloud/richdocuments/issues/3029
Fixes https://github.com/nextcloud/richdocuments/issues/2983

Steps to reproduce:
- Change user language to german
- Create a new empty file
- See that the spell checking language of the new document properly adapts

### Checklist


juliushaertl:
Arg, merged the wrong tab 🙈 @blizzz Up for a quick retroactive review?

thebearon:
@juliushaertl I assume this fix is also relevant to version 25, is it so?

PR #3103, opened 2023-08-09
CollaboraOnline#6546 enable automatic color in default paragraph style

timar:
use the xml token "use-window-font-color" to toggle auto color on.

https://github.com/CollaboraOnline/online/issues/6546

* Resolves: regression from PR #3000 
* Target version: main, stable

### Summary
PR #3000 introduced a regression, the template.odt was no longer locale-agnostic, it was Turkish. We used minimal ODF files, created manually, and when these files were opened by COOL, the language dependent settings were set according to the current locale of the user (document language, paper size, currency etc.).
@caolanm  gave a good solution with 80e85b33b004657a38b8279541e2e7a5637034c7 but the wrong, unused (?) file was patched, the document.ott.
So I will use @caolan's solution for template.odt, too.

### Checklist

- [ x] Code is properly formatted
- [ x] Sign-off message is added to all commits
- [ x] Documentation (manuals or wiki) has been updated or is not required


juliushaertl:
@timar Are you aware of any good tool to review such document changes? Is there any cli tool that could be used to diff the two?

timar:
> @timar Are you aware of any good tool to review such document changes? Is there any cli tool that could be used to diff the two?

@juliushaertl I usually unzip the ODF files and use a GUI tool (Beyond Compare). Actually I detected this issue when I wanted to solve the automatic paragraph color in Android app and searche our bug tracker to check if it was reported somewhere else. 
Basically here I reverted PR #3000 and added the styles.xml from 80e85b33b004657a38b8279541e2e7a5637034c7
Nothing new.

juliushaertl:
All right, thanks for the hint.

thebearon:
Manual backports:
#3147
#3148
#3149

PR #3107, opened 2023-08-14
fix: Log invalid wopi tokens at info level instead of error

juliushaertl:
This is a quite common case when inactive browser tabs are reconnecting so we should rather log at info than error.

### Checklist


max-nextcloud (approved):
Good idea!

luka-nextcloud (approved):

PR #3112, opened 2023-08-16
fix: Disable pdf preview gnerator through Collabora if server already has support for it

juliushaertl:
If the core preview provider for PDF is enabled we don't need to register the one from Collabora which involves additional network transfer and is likely slower due to that.

### Checklist


max-nextcloud (approved):

szaimen:
https://github.com/nextcloud/richdocuments/pull/3146

juliushaertl:
stable26 in https://github.com/nextcloud/richdocuments/pull/3176

Issue #3102, opened 2023-08-08
Version preview support in 27.1


juliushaertl:
Ref https://github.com/nextcloud/viewer/pull/1835

PR #3152, opened 2023-08-31
Version support and iframe refactoring

### LINKED ISSUE #3011: Refactor frontend to remove second iframe
I did a small PoC in https://github.com/nextcloud/richdocuments/compare/chore/refactor-iframe to remove the need for the separate document frame that is being loaded and instead load the Collabora page directly in the first iframe. Instead of loading the intermediate frame it just queries and endpoint to obtain the WOPI token and can then directly pass it to Collabora.

This has several benefits:
- Reduces code complexity as we no longer need to proxy messages through two frames
- No need to keep document state in two different script contexts
- Speeds up loading time by a full page load with Nextcloud javascript

In general this works nicely, but needs a bit of work in order to make sure all previous functionality is still available:
- [x] Basic PoC for viewer usage
  - [ ] File creation from templates
- [ ] Public share link (folder)
  - [ ] File creation from templates
- [ ] Public share link (single file)
- [ ] Direct editing logged in
  - [ ] File creation from templates
- [ ] Direct editing share link
- [ ] Federated editing
- [ ] Make sure all post messages are handled as before

juliushaertl:
Fixes https://github.com/nextcloud/richdocuments/issues/3102
Fixes #3011 

Best to be reviewed as [without whitespaces](https://github.com/nextcloud/richdocuments/pull/3152/files?w=1)

Requires https://github.com/nextcloud/server/pull/40184 for a smooth restore

Sorry I went a bit nuts on refactoring bits in the document controller, so this is a larger changeset, but the relevant changes in summary:

- Introduce a token endpoint to generate a wopi token
- Rework the viewer vue component to request the token from the new endpoint and directly load Collabora in an iframe instead of loading an intermediate page (DocumentController::index) which used to generate the token and load another iframe within
- Remove transition in vue component to avoid flickering
- Move missing post message logic into vue component (as the existing implementation from document.js is no longer used now in the viewer)
- Add cypress tests for important integration parts where the post message handling is happening
- Simplify and merge common logic to get a the file and share for token generation
- With that we can now also load versions through the token endpoint which enables us to make use of the version preview and side by side comparison with 27.1

### Summary

- [x] Refactor iframe loading for viewer to only load one instead of two nested iframes
- [x] Generate a token before through an API call
- [x] Implement support for version preview and comparison with viewer
- [x] Remove micro transition for animating the fade in as it caused more troubles then it helped

### TODO

- [x] Double check all cases for iframe loading
- [ ] Document follow ups to see where we still need to replace the double iframe
  - [ ] Public single file shares
  - [ ] Direct editing
  - [ ] Federated editing
- [x] Cleanup post message handling
- [x] Tests
- [ ] Follow up issue for text to handle restore gracefully

### Checklist


lib/Controller/DocumentController.php
```diff
@@ -184,7 +184,7 @@ public function index($fileId, ?string $path = null) {
 			$params = [
 				'permissions' => $item->getPermissions(),
 				'title' => $item->getName(),
-				'fileId' => $item->getId() . '_' . $this->config->getSystemValue('instanceid'),
+				'fileId' => $item->getId() . '_' . $this->config->getSystemValue('instanceid') . ($version !== null ? '_' . $version : ''),
```
juliushaertl, line 187: This is no longer needed i think

lib/Controller/DocumentController.php
```diff
@@ -489,4 +489,17 @@ public function editOnlineTarget(int $fileId, ?string $target = null) {
 
 		return $this->renderErrorPage('File not found', Http::STATUS_NOT_FOUND);
 	}
+
+
+
+	#[Http\Attribute\NoAdminRequired]
+	#[Http\Attribute\NoCSRFRequired]
+	public function token(int $fileId, ?string $shareToken = null) {
+		[$urlSrc, $token, $wopi] = $this->tokenManager->getToken((string)$fileId, $shareToken, $this->uid, false);
```
juliushaertl, line 498: Double check if this needs further safeguards

juliushaertl:
Quick profile of what we save between old and new appraoch

From Viewer loading to actual collabora frame being loaded

## Old
Full additional page load with core javascript initialization, times hard to measure reliably but in the profile ~4s
Screenshot 2023-08-31 at 12 53 09

## New
Only one API call ~80ms
image

lib/TokenManager.php
```diff
@@ -205,7 +207,6 @@ public function getToken(string $fileId, $shareToken = null, $editoruid = null,
 
 		return [
 			$this->wopiParser->getUrlSrc($file->getMimeType())['urlsrc'], // url src might not be found ehre
```
juliushaertl, line 209: - [ ] remove url src from here and rename method to match new return value

lib/Controller/DocumentController.php
```diff
@@ -106,7 +79,7 @@ public function __construct(
 	 *
```
juliushaertl, line 79: Migrate all occurences of annotations to attributes

src/mixins/openLocal.js
```diff
@@ -0,0 +1,91 @@
...
+	data() {
+		return {
+			openingLocally: false,
+		}
+	},
+	methods: {
+		unlockAndOpenLocally() {
+			if (this.openingLocally) {
+				this.unlockFile()
+					.catch(_ => {}) // Unlocking failed, possibly because file was not locked, we want to proceed regardless.
```
max-nextcloud, line 38: Maybe check the server response more carefully?
This might be anything from dns failure to 500, 502. Not sure what the response would be in case the file was not locked. But I think the catch should be restricted to the case you are anticipating.

src/mixins/openLocal.js
```diff
@@ -0,0 +1,91 @@
...
+					OC.linkToOCS('apps/files/api/v1', 2) + 'openlocaleditor?format=json',
+					{ path: this.filename }
+				).then((result) => {
+					const url = 'nc://open/'
+						+ getCurrentUser()?.uid + '@' + getNextcloudUrl()
+						+ OC.encodePath(this.filename)
+						+ '?token=' + result.data.ocs.data.token
+
+					this.showOpenLocalConfirmation(url, window.top)
+					window.location.href = url
```
max-nextcloud, line 86: If i read the code correctly this will also happen if the confirmation is declined.
I assume that will just cause the file to be opened online and is intended. Just want to make sure.

max-nextcloud (approved):
Just had a closer look at the mixins. They look good except for one catch that i think is too broad.

src/mixins/openLocal.js
juliushaertl: I refactored this a bit to actually show a slightly different dialog to indicate why opening might not have worked and allow to retry or switch back to online editing.

PR #3141, opened 2023-08-25
feat: add config option mobile_editing

luka-nextcloud:
* Resolves: https://support.nextcloud.com/#ticket/zoom/49745
* Target version: main

### TODO

- [ ] ...

### Checklist

- [ ] Code is properly formatted
- [ ] Sign-off message is added to all commits
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #3184, opened 2023-09-17
fix: Mobile support shall be enabled by default

juliushaertl:
Seems we broke the default with #3141 

Now it compares against the default of 'yes' which does not evaluate to false by default.

To disable one can use:
```
occ config:app:set richdocuments mobile_editing --value="no"
```

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -149,7 +149,7 @@ class Capabilities implements ICapability {
 					'mimetypesNoDefaultOpen' => array_values($optionalMimetypes),
 					'mimetypesSecureView' => $this->config->useSecureViewAdditionalMimes() ? self::SECURE_VIEW_ADDITIONAL_MIMES : [],
 					'collabora' => $collaboraCapabilities,
-					'direct_editing' => ($collaboraCapabilities['hasMobileSupport'] ?? false) && $this->config->getAppValue('mobile_editing') ?: false,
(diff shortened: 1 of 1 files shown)


max-nextcloud (approved):
code looks good.

Alternativend:
Is this the reason why office documents on Android are not working? 
iOS and Browser are fine here, but all Androids are broken.

jakobroehrl:
> Is this the reason why office documents on Android are not working? iOS and Browser are fine here, but all Androids are broken.

https://github.com/nextcloud/richdocuments/pull/3141#issuecomment-1728892509

juliushaertl:
https://github.com/nextcloud/richdocuments/pull/3220
https://github.com/nextcloud/richdocuments/pull/3221
https://github.com/nextcloud/richdocuments/pull/3222
https://github.com/nextcloud/richdocuments/pull/3223

Alternativend:
Still the same problems after updating Nextcloud Office to 8.2.1.

raystek:
The ios Nextcloud app works fine. It opens up the document for viewing/editing. The Android app just spins.

max-nextcloud:
@raystek could you open a new issue for this? Your comment might get lost here as the PR has already been merged.
Thanks!

Keessaus:
I am experiencing the same issues, but do not have much in terms of debugging data. If @raystek is still experiencing these issues, could you create an issue? If not, I would happily try to debug this if someone could point me to some Nextcloud Android app logs.

Keessaus:
For me, this seems to be related to a (closed) Android app issue: <https://github.com/nextcloud/android/issues/12338>.

PR #3206, opened 2023-09-28
Allow current user lock

ArtificialOwl:


lib/Controller/WopiController.php
```diff
@@ -710,6 +710,16 @@ private function lock(Wopi $wopi, string $lock): JSONResponse {
 		} catch (NoLockProviderException|PreConditionNotMetException $e) {
 			return new JSONResponse([], Http::STATUS_BAD_REQUEST);
 		} catch (OwnerLockedException $e) {
+
+			// This is a quick custom patch to allow richdocuments to accept lock from current user
+			/** @var OCP\IUserSession $userSession */
+			$userSession = \OC::$server->get(\OCP\IUserSession::class);
```
juliushaertl, line 716: We don't have a user session on those requests, we probably should just check against `$wopi->getEditorUid()`. However this would then still allow conflicts if the file gets locked by the desktop client and the same user opens it in Collabora. 

@ArtificialOwl Can you provide some more backgorund which case this change should cover?

ArtificialOwl: you can reproduce the issue:

- `userA` lock a shared file,
- `userA` open the freshly locked file with richdocument,
- the app will not open the said document

juliushaertl:
- User lock is accepted for opening with the same user (edit permissions)
- If a user lock is present we should open the file read only for other users

juliushaertl:
Pushed a few adjustments

juliushaertl (approved):

ArtificialOwl:
Tested ✔️

Issue #3187, opened 2023-09-20
File shared by link cannot be collaboratively edited


eszkadev:
**Describe the bug**
This is regression. When two users (author and second user who get public link) open file - they land in two copies of the same document. They don't see each other edits/cursors.

**To Reproduce**
1. User A shares a .odt file using a link
2. User A opens just shared file
3. User B (incognito tab, anonymous user) opens link shared by user A
4. User A or B  does some editing

**Expected behavior**
Both users see each other edits.

**Current result**
Users see two different copies of the same document, they can't collaborate. Notice they don't see other user avatars in the top bar.

**Client details:**
not important

## Server details

**Nextcloud version:** 28 / master, hash: fecae9422902fa1cbc6366b9379924f5586a1eda - doesn't matter

**Version of the richdocuments app** 8.2.0-dev / master, hash: 731c482833b9ca81daec6ed544b930396c38f5d2

**Version of Collabora Online** 23.05.4 / master - doesn't matter

## My research

This is regression bisected to https://github.com/nextcloud/richdocuments/commit/ba45233bff0c5356e8e926d4bffd8d34053b16cf
from PR: https://github.com/nextcloud/richdocuments/pull/3152

woiwoiwoiwoi:
I was just about to open an issue about this. Same here.

Nextcloud version: 27.1.0
COOLWSD version: 23.05.3.1
Nextcloud Office / richdocuments: 8.2.0

mmeeks:
Awesome - thanks Julius - great to have you back :-)

PR #3218, opened 2023-10-10
fix: Properly provide instance id

juliushaertl:
Fix diverging sessions between logged in and public users

Fix https://github.com/nextcloud/richdocuments/issues/3187

Being used in the frontend at https://github.com/nextcloud/richdocuments/blob/17389a75c8ffba3575b6759ea9b06e226aa3e4b1/src/view/Office.vue#L246


juliushaertl:
Cypress failure is known due to upstream issue in server which breaks viewer

blizzz (approved):

Issue #2618, opened 2022-11-14
Nextcloud session timing out while editing a document using Nextcloud Office


elhananjair:
**Describe the bug**
Hello there, 
I have set session_timeout to 5 Minutes on nextcloud config.php, the problem is when I am editing a document file on Nextcloud Office it will log out while I am typing or doing some activity. 
**To Reproduce**
1. Go to nextcloud config.php file, and add this config parameters, `'session_lifetime' => 300, 'session_keepalive' => false, 'auto_logout' => true`
2. Click on any document and keep editing for more than 5 Minutes
3. Nextcloud will log you out

**Expected behavior**
Nextcloud should have to see editing of documents as an activity and should extend the session lifetime. 

**Screenshots**

**Client details:**
 - OS: Fedora 
 - Browser: Firefox
 - Version: 106.0.4
 - Device: Laptop

## Server details

**Operating system**: Fedora Server

**Web server:** Apache

**Database:** PostgreSQL

**PHP version:** 8.0.25

**Nextcloud version:** 25.0.1

**Version of the richdocuments app**  7.0.1

**Version of Collabora Online: ** CODE 22.05.8.2

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
{"reqId":"Y3JNBRgQ1-c-yAL2i0olRQAAyhQ","level":3,"time":"2022-11-14T14:13:26+00:00","remoteAddr":"192.168.22.106","user":"--","app":"PHP","method":"GET","url":"/logout?requesttoken=EOtl1pZKLlZaRDm3VPRIXFxYDFsKS0SSiMu3LkTf1mw%3D%3Aetgz%2Fc44Yy89JWrnF456KnMCVAgzLS%2F%2F2a%2FtRh7thzg%3D","message":"session_start(): Ignoring session_start() because a session is already active at /var/www/html/nextcloud/lib/private/Session/Internal.php#219","userAgent":"Mozilla/5.0 (X11; Linux x86_64; rv:106.0) Gecko/20100101 Firefox/106.0","version":"25.0.1.1","data":{"app":"PHP"}}

```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

juliushaertl:
The document being edited outside of the Nextcloud session through Collabora directly so there is no way that we can detect that the session is still active when `session_keepalive` is disabled. This is something we don't consider implementing.

[...]

PR #3224, opened 2023-10-13
fix: Disable auto logout while editing

juliushaertl:
- [x] Requires discussion with Collabora about how we can safely detect activity/inactivity
- [x] https://github.com/CollaboraOnline/online/pull/7514
- [x] Integrate post messages

Fix https://github.com/nextcloud/richdocuments/issues/2618

### DIFF
--- a/src/mixins/autoLogout.js
+++ b/src/mixins/autoLogout.js
@@ -1,0 +1,101 @@
+/*
+ * @copyright Copyright (c) 2023 Julius Härtl <jus@bitgrid.net>
+ *
+ * @author Julius Härtl <jus@bitgrid.net>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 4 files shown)


src/mixins/autoLogout.js
```diff
@@ -0,0 +1,111 @@
...
+import { generateUrl } from '@nextcloud/router'
+import { getCurrentUser } from '@nextcloud/auth'
+
+const config = loadState('core', 'config', {})
+
+const getInterval = () => {
+	let interval = NaN
+	if (config.session_lifetime) {
+		interval = Math.floor(config.session_lifetime / 2)
+	}
```
mejo-, line 33:
```suggestion
	const interval = config.session_lifetime
		? Math.floor(config.session_lifetime / 2)
		: NaN
```

mejo-: And why not set `interval` to `900` directly here?

```js

	const interval = config.session_lifetime
		? Math.floor(config.session_lifetime / 2)
		: 900
	return Math.min(24 * 3600, Math.max(60, interval))
```

src/mixins/autoLogout.js
```diff
@@ -0,0 +1,111 @@
...
+			localStorage.setItem('lastActive', newValue)
+			const event = new StorageEvent('storage', {
+				key: 'lastActive',
+				oldValue,
+				newValue,
+			})
+			window.dispatchEvent(event)
+		},
+		registerCheck() {
+			if (!this.autoLogoutRegistered && this.postMessage) {
```
mejo-, line 82: I don't understand where `this.postMessage` and `this.postMessage.registerPostMessageHandler()` come from :thinking: 

Same for `this.sendPostMessage()` in the next if block.

mejo- (dismissed):

src/mixins/autoLogout.js
juliushaertl: They are methods of the Office.vue component, this mixin is only used there and just to group related code together

datenangebot (approved):
looks good, did not test it

Issue #3080, opened 2023-07-20
[Bug]: Code arm64 server stopped working after version 27 upgrade


astrometrics:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

Code arm64 server stopped working after version 27 upgrade... worked mostly fine before

### Steps to reproduce

1. Upgrade to NC 27
2. Try to use office document
3. fails

### Expected behavior

just open office document normally via NC

### Installation method

Community Manual installation with Archive

### Nextcloud Server version

27

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database engine version

PostgreSQL

### Is this bug present after an update or on a fresh install?

Upgraded to a MAJOR version (ex. 22 to 23)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
{
    "system": {
        "instanceid": "***REMOVED SENSITIVE VALUE***",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "xx.x"
        ],
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "dbtype": "pgsql",
        "version": "27.0.1.2",
        "overwrite.cli.url": "https:\/\/xx.x\/nc",
        "dbname": "***REMOVED SENSITIVE VALUE***",
        "dbhost": "***REMOVED SENSITIVE VALUE***",
        "dbport": "",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
[...]

PR #3252, opened 2023-10-23
Improve logging for WOPI Parser

R0Wi:
* Helps to find errors related to https://github.com/nextcloud/richdocuments/issues/3080 (see also https://github.com/CollaboraOnline/richdocumentscode/issues/119#issuecomment-1774850066)
* Target version: main

### Summary

Improve logging to be able to troubleshoot problems with WOPI discovery. If an admin sees a message like this in the logs:

```
[richdocuments] Error: Exception: Could not find urlsrc in WOPI
```

this usually means, that the WOPI discovery response was faulty.

A valid discovery response, which is received [here](https://github.com/nextcloud/richdocuments/blob/11a0387687dc0c8bd6d3c129641a3a56a5101800/lib/WOPI/Parser.php#L41), looks like this:

```xml
<wopi-discovery>
    <net-zone name="external-http">
        <app name="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet">
            <action default="true" ext="" name="edit" urlsrc="https://<NC_URL>/custom_apps/richdocumentscode/proxy.php?req=/browser/<TOKEN>/cool.html?"/>
        </app>
        <!-- Many more app entries ...-->
    </net-zone>
</wopi-discovery>
```

Im my case (because of some backend server error) the response was:

```html
<html><body>
<h1>Socket proxy error</h1>
<p>Error: Timed out opening local socket: 99 - Cannot assign requested address</p>
</body></html>
```

but I wasn't able to see this message until I added the logging lines of this PR.

### Checklist

### DIFF
--- a/lib/WOPI/Parser.php
+++ b/lib/WOPI/Parser.php
@@ -21,15 +21,22 @@
 
 namespace OCA\Richdocuments\WOPI;
 
+use Psr\Log\LoggerInterface;
+
 class Parser {
 	/** @var DiscoveryManager */
 	private $discoveryManager;
 
+	/** @var LoggerInterface */
+	private $logger;
+
 	/**
 	 * @param DiscoveryManager $discoveryManager
+	 * @param LoggerInterface $logger
 	 */
-	public function __construct(DiscoveryManager $discoveryManager) {
+	public function __construct(DiscoveryManager $discoveryManager, LoggerInterface $logger) {
 		$this->discoveryManager = $discoveryManager;
+		$this->logger = $logger;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


lib/WOPI/Parser.php
```diff
@@ -56,6 +64,7 @@ public function getUrlSrc($mimetype) {
 			];
 		}
 
+		$this->logger->warning('Didn\'t find urlsrc for mimetype {mimetype} in this WOPI discovery response: {discovery}', ['mimetype' => $mimetype, 'discovery' => $discovery]);
```
juliushaertl, line 67:
```suggestion
		$this->logger->error('Didn\'t find urlsrc for mimetype {mimetype} in this WOPI discovery response: {discovery}', ['mimetype' => $mimetype, 'discovery' => $discovery]);
```

I think this could even be logged at error level to be sure to catch it in any case.

juliushaertl:
That makes a lot of sense. I even think we can increase the level to error for the one message.

juliushaertl (approved):

Issue #3203, opened 2023-09-27
The Nextcloud sidebar shows up automatically and I don't understand its pattern


pedropintosilva:
The sidebar box that shows who has shared the file/folder with me appears robbing me of value space on from document and the document's interface. This is particularity dramatic when working with finance related documents (spreadsheets).

Result: I have developed the habit of instantly move the mice to the to top right corner to close that box a soon as it appears
  * Problem: it doesn't always appear and I don't understand the pattern. Thus, sometimes I close the whole document by mistaken because in that very instance the other sidebar didn't appear and by newly learned skill/habit got the best of me.   

----

Versions 

Reported by Julie

* Nextcloud
  * This Nextcloud is on version 27.0.2 Enterprise
  * OC.getCapabilities().richdocuments.version
    * 10:18:03.123 "8.2.0" 
* Collabora Online
  * COOLWSD version: 23.05.4.1snapshot (git hash: d39d2cd (E))
  * LOKit version: Collabora Office 23.05.4.20230917 (git hash: 34a9524)
  * Served by: Debian GNU/Linux 12 (bookworm)
  * Server ID: 5da3ee77

jancborchardt:
Sounds good to me to not automatically show our sidebar. I think we initially did this to make sharing more present, but indeed the document should take precedence.

cc @juliushaertl @AndyScherzinger

AndyScherzinger:
I agree 

> document should take precedence.

juliushaertl:
I don't think we explicitly open it in richdocuments anywhere, so might be worth to check other places like the files app or viewer if there is some logic that may open the sidebar automatically when opening.

juliushaertl:
Thanks for that hint. Then it is the files app that shows the sidebar.

@jancborchardt I see two options here:

- We do not automatically show the sidebar for opening any file type (pdf, images, office, text, ...)
- We close it again if a office fileis opened (if we want to limit this behaviour to office itself)

From my perspective we should just not show it for any file type as we open the viewer directly and should probably give the content as much space as possible there.

juliushaertl:
Note this behaviour is currently broken on master due to the files2vue migration which does not support to open a file directly yet, but reproducible with 27 and lower

jancborchardt:
@juliushaertl I'd say making an exception for Office would be good, because we do want to show the sidebar features presently. It's just that Office already has so many things that it's overwhelming in addition.

Darshan-upadhyay1110:
[...]

PR #3259, opened 2023-10-26
Close NC sidebar on file open

Darshan-upadhyay1110:
* Resolves: #3203
* Target version: main

### Summary
    - For internal shared links of Office documents sidebar gets open by default
    - sidebar should not open at all when we click on document file or file gets opened by shared link

CC: @juliushaertl


Darshan-upadhyay1110:
@juliushaertl do we need backport to 27 and 26 as well ?

juliushaertl:
Yes, but we can do this likely automatically once this one is merged. No need to manually prepare those for now.

juliushaertl:
@Darshan-upadhyay1110 I also invited you to the repo so you can create branches directly here and CI will run automatically for future prs ;)

Darshan-upadhyay1110:
@juliushaertl can we add comment in the main issue and do not close that ?

as this is a temp fix so it is good to just let that issue open for more generic fix

juliushaertl:
@Darshan-upadhyay1110 When reviewing i figured out a better place to put this which actually resolves the flickering as well.

- We close the sidebar now as early as the component renders
- It is only closed when opening a normal file, not when you open a version viewer from the sidebar as in this case we want to keep the version listing open

juliushaertl (approved):

PR #3260, opened 2023-10-26
feat: Add headers for wasm support

juliushaertl:

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Thanks @juliushaertl.

Looks good to me. @caolanm, any concerns?

caolanm:
no concerns, this seems to give us what we want without disrupting existing setups

mejo- (approved):
No idea about wasm, but code changes look sensible.

juliushaertl:
Amended a small change to only set the allowEvalWasm if wasm support is indicated by the capabilities.

Yetangitu:
These headers are added to _all_ Nextcloud apps by _richdocuments_ which causes the _News_ app to [fail to show feed icons](https://github.com/nextcloud/news/issues/2703#issue-2366708989). I solved this by commenting out that part of _richdocuments_ but that's a crude hack. These headers should _only_ be inserted when the _richdocuments_ app is active and in use in the current session, not when other apps are in the foreground. As to how to achieve this I have not found a good solution yet but maybe I'm overlooking something?

juliushaertl:
Can you please file a new issue about this? It could be that we apply those too widely.

Just a sidenote: Wouldn't it be better to fetch the icons on the backend and cache them on the Nextcloud side? That way the request would be also hiding the individual user ips and avoid loading remote content directly.

PR #3277, opened 2023-11-06
fix: Properly update files app state when using saveAs

juliushaertl:
When using saveAs the document in use changes to a new one so we need to replace the files route in the background to properly handle a reload of the browser or copying the URL.


mejo- (approved):
Code changes look good to me :+1:

PR #3303, opened 2023-11-17
feat: add preview provider for emf files

kesselb:
* Target version: main

### Summary

For https://github.com/nextcloud/files_emfviewer/issues/1
In addition to https://github.com/nextcloud/server/pull/41395

We added support for image/emf via LibreOffice for 28. 

For setups already using Richdocuments/Collabora it's nicer to let the previews generate via collabora web and no local LibreOffice installation is required. 

### TODO

- [x] CI
- [ ] Test
- [ ] Documentation

### Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -37,6 +37,7 @@ use OCA\Richdocuments\Listener\ReferenceListener;
 use OCA\Richdocuments\Listener\ShareLinkListener;
 use OCA\Richdocuments\Middleware\WOPIMiddleware;
 use OCA\Richdocuments\PermissionManager;
+use OCA\Richdocuments\Preview\EMF;
 use OCA\Richdocuments\Preview\MSExcel;
 use OCA\Richdocuments\Preview\MSWord;
 use OCA\Richdocuments\Preview\OOXML;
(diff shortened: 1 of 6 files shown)


lib/AppInfo/Application.php
```diff
@@ -88,6 +89,7 @@ public function register(IRegistrationContext $context): void {
 			'getPathForToken',
 			'getWopiForToken',
 		]);
+		$context->registerPreviewProvider(EMF::class, '/image\/emf/');
```
juliushaertl, line 92: Did you give that a test with a Collabora intance setup? We have not migrated our preview class to the IProviderV2 interface (which we really should 🙈 ) and as far as I see it the context registration requires that.

kesselb: > Did you give that a test with a Collabora intance setup?

Yep, it works like a charm ;)

> the context registration requires that.

True, the psalm-param type hints IProviderV2.

I guess it's better to use the deprecated way that expects a IProvider instance for now.

juliushaertl: Yep, let's stick to that then for now

juliushaertl (approved):

kesselb:
Documentation: https://github.com/nextcloud/documentation/pull/11292

Issue #3139, opened 2023-08-24
Office: Fonts improvements: Previews, popular from Google Fonts, checking with LO desktop fonts, deduplicating Noto Sans & Serif variants


juliushaertl:
- [x] Font preview ongoing in COOL
  - [x] https://github.com/CollaboraOnline/online/pull/7131
- [ ] Ship popular fonts
	- [ ] Clarify licenses if we can ship it, need to attribute (and how)
	- [ ] Test if variable fonts work as expected (e.g. `RobotoMono-VariableFont_wght.ttf`)
	- [ ] Pick font files to ship depending on if variable sizing is working to not increase the richdocuments bundle size too much
	- [ ] Implementation https://github.com/nextcloud/richdocuments/pull/3304 @luka-nextcloud 
  	  - [ ] Font files get shipped as part of richdocuments in an extra assets/fonts/ directory
	  - [ ] New fonts shall be copied from the assets/fonts folder and added through the existing [FontService method](https://github.com/nextcloud/richdocuments/blob/05657472698a27f8afc57072e777e83240280bab/lib/Service/FontService.php#L155)
	  - [ ] Add a [repair job](https://docs.nextcloud.com/server/19/developer_manual/app/repair.html) to add new fonts once to the installation ([repair step version check example](https://github.com/nextcloud/text/blob/886012022aac3e4a525f9a7e6af9ecbf3464211f/lib/Migration/ResetSessionsBeforeYjs.php#L39))
	  - [ ] Add an occ command to install shipped fonts again if manually removed as the admin
	  - [ ] Cypress test that ensures the fonts are listed in the admin settings
- [ ] Possible follow up: Deduplicate/group fonts in the UI -> Collabora

juliushaertl:
@jancborchardt ping for a possible list of fonts. We could start shipping those with richdocuments and then try to improve the auto setup to pick up fonts automatically from there.

jancborchardt:
Here are the ones Google Docs provides and I would say we make sure to add those. Fonts in (parentheses) are proprietary though:


- Amatic https://fonts.google.com/specimen/Amatic+SC
- (Arial)
- Caveat https://fonts.google.com/specimen/Caveat
- Comfortaa https://fonts.google.com/specimen/Comfortaa
- (Courier New)
- EB Garamond https://fonts.google.com/specimen/EB+Garamond
- Lexend https://fonts.google.com/specimen/Lexend
- Lobster https://fonts.google.com/specimen/Lobster
[...]

PR #3304, opened 2023-11-17
Feature/fonts improvement

luka-nextcloud:
* Resolves: #3139
* Target version: main

### Summary

- Add repair step to install default fonts on installation
- Add command richdocuments:install-fonts  to install default fonts

### TODO

- Font files get shipped as part of richdocuments in an extra assets/fonts/ directory
- New fonts shall be copied from the assets/fonts folder
- Add a repair job to add new fonts once to the installation
- Add an occ command richdocuments:install-fonts to install shipped fonts again if manually removed as the admin
- Add cypress test that ensures the fonts are listed in the admin settings


kesselb:
Tip: If composer dump-autoload does not update the autoloader then `cd composer`. I just ran into the same failing workflow today and was super confused by it :see_no_evil:

lib/Command/InstallDefaultFonts.php
```diff
@@ -0,0 +1,35 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class InstallDefaultFonts extends Command {
+	private FontService $fontService;
+
+	public function __construct(FontService $fontService) {
+		parent::__construct();
+		$this->fontService = $fontService;
+	}
```
juliushaertl, line 16: We can make use of PHP 8 constructor property promotion:

https://php.watch/versions/8.0/constructor-property-promotion

```suggestion
	public function __construct(private FontService $fontService) {
		parent::__construct();
	}
```

lib/Migration/InstallDefaultFonts.php
```diff
@@ -0,0 +1,32 @@
...
+use OCP\Migration\IRepairStep;
+
+class InstallDefaultFonts implements IRepairStep {
+	private IConfig $config;
+	private FontService $fontService;
+
+	public function __construct(IConfig $config, FontService $fontService) {
+		$this->config = $config;
+		$this->fontService = $fontService;
+	}
```
juliushaertl, line 17:
```suggestion
	public function __construct(private IConfig $config, private FontService $fontService) {
	}
```

lib/Service/FontService.php
```diff
@@ -242,4 +243,41 @@ private function generateFontOverview(ISimpleFile $fontFile): void {
...
+				continue;
+			}
+
+			$filePath = $dirPath . '/' . $fileName;
+
+			if (!is_file($filePath)) {
+				continue;
+			}
+
+			$fileContent = file_get_contents($filePath);
```
juliushaertl, line 274: While this is fine for this case a general note that using file_get_contents would read the full file content into memory which might cause problems with larger files. Better approach would be to use fopen instead and pass a file pointer instead which is also supported by the `newFile` method that is called within `uploadFontFile`

juliushaertl (commented):
Haven't tested yet, but looks good. Only a few minor comments inline.

lib/Command/InstallDefaultFonts.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace OCA\Richdocuments\Command;
+
+use OCA\Richdocuments\Service\FontService;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class InstallDefaultFonts extends Command {
```
juliushaertl, line 10: Ah we still don't have a public API for commands 🙈 We can add the required classes as a stub to tests/stub.phpstub

juliushaertl:
@luka-nextcloud Please make sure to provide some context like a short pr description and the link to the issue

juliushaertl:
> Tip: If composer dump-autoload does not update the autoloader then cd composer. I just ran into the same failing workflow today and was super confused by it

Yeah we should probably document that better. I went for a separate composer.json for the autoloader only to avoid dev dependencies getting autoloaded, but not the most straight forward I agree

juliushaertl (approved):

juliushaertl:
Pushed a commit adding the fonts and licenses through a build script.

PR #3217, opened 2023-10-10
added new 'Error_Messages' postmessage

Rash419:
- so that collbora online can show a custom with nextcloud documentation link

### TODO

- [x] Requires https://github.com/CollaboraOnline/online/pull/7375
- [ ] https://github.com/CollaboraOnline/online/issues/7709
- [ ] Only display additional docs hint for admins
- [ ] Test different scenarios where connection does not work (see admin messages in https://github.com/nextcloud/richdocuments/pull/3315)

### Checklist

- [x] Code is properly formatted
- [x] Sign-off message is added to all commits
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #3311, opened 2023-11-20
feat: add loadingMessage() to show different loadingMsg

Rash419:
* Resolves: # 
* Target version: main

### Summary
Basically when using indirection it might happen that there cluster is scaling and there is request to open the document 
COOL will retry after a interval to get the new routeToken but until then we update the user with the loading message that "Cluster is scaling" or something even better
corresponding COOL pr: https://github.com/CollaboraOnline/online/pull/7683
### Checklist

### DIFF
--- a/src/view/Office.vue
+++ b/src/view/Office.vue
@@ -25,7 +25,7 @@
 		<div v-if="showLoadingIndicator"
 			class="office-viewer__loading-overlay"
 			:class="{ debug: debug }">
-			<NcEmptyContent v-if="!error" :name="t('richdocuments', 'Loading {filename} …', { filename: basename }, 1, {escape: false})">
+			<NcEmptyContent v-if="!error" :name="loadingMessage">
 				<template #icon>
 					<NcLoadingIcon />
 				</template>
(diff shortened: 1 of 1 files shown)


src/view/Office.vue
```diff
@@ -186,6 +187,13 @@ export default {
 				return this.error
 			}
 		},
+		loadingMessage() {
+			if (this.loadingMsg) {
+				return t('richdocuments', this.loadingMsg)
```
juliushaertl, line 192: This won't work for translating. Either you will need to pass over an already translated message or pass over a static load message code so we can handle translations within richdocuments for each specific message.

src/view/Office.vue
```diff
@@ -320,8 +328,13 @@ export default {
 				if (args.success) {
 					this.documentReady()
 				} else {
-					this.error = args.errorMsg
-					this.loading = LOADING_STATE.FAILED
+					if (args.errorType === 'clusterscaling') {
+						this.loadingMsg = args.errorMsg
```
juliushaertl, line 332: Ah the error type is already used, so we could just have something like
```suggestion
						this.loadingMsg = t('richdocuments', 'Cluster is scaling')
```

However it would be good to find a nice message that is not technical as this message is presented to end users not only admins.

juliushaertl:
I pushed some changes in also include handling of https://github.com/CollaboraOnline/online/commit/d4918c79981058dce6d9071d2d64ce8cdb8f8c29 so this also includes what is needed for #3217

Issue #3086, opened 2023-07-25
Missing IDN conversion in Content-Security-Policy builder


Sprinterfreak:
**Describe the bug**
richdocuments breaks entire nextcloud instance if installed on an instance with unicode domain.  
Adds raw unicode domain to the Content-Security-Policy-Header which makes browsers ignore those and therefore prevent any assets from being loaded. Nextcloud without javascript, images and css doesn't look that good.

**To Reproduce**
Install richdocuments on an instance with idn domain.

**Expected behavior**
FQDN in the Content-Security-Policy-Header must be idn_ascii converted

See https://github.com/nextcloud/server/issues/39555 for details (including patch)

@kesselb has allready developed a patch for that https://github.com/nextcloud/server/issues/39555#issuecomment-1648414770, which I can confirm working.

Sprinterfreak:
Still affects richdocuments 8.1.1

Sprinterfreak:
Still affects richdocuments 8.2.3 although the responsible code has been rewritten

juliushaertl:
Pasting this here for reference:

https://www.php.net/manual/de/function.idn-to-ascii.php

```
Index: lib/Listener/CSPListener.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lib/Listener/CSPListener.php b/lib/Listener/CSPListener.php
--- a/lib/Listener/CSPListener.php	(revision 1b609dac1bf216c4913188ab3861f37fabb49147)
+++ b/lib/Listener/CSPListener.php	(date 1690223340061)
@@ -66,6 +66,7 @@
 		);
 
 		$urls = array_filter($urls);
+		$urls = array_map(fn($url) => idn_to_ascii($url), $urls);
 
 		$policy = new EmptyContentSecurityPolicy();
 		$policy->addAllowedFrameDomain("'self'");
```

juliushaertl:
Added a commit to #3315

PR #3315, opened 2023-11-23
Improve setup checks and url handling with separate callback url

juliushaertl:
- fix: Properly handle internal urls
- chore: Update psalm baseline with missing symfony dependencies
- feat: Add browser connectivity check

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


juliushaertl:
@joshtrichards I'm unsure if you have a development setup, but I'd be curious about your feedback around this given that you did quite some issue triaging efforts. This PR aims to introduce one additional setting but the core part is to improve error reporting for admins. Maybe you have a way to give it a try? Or just give some feedback on further improving the error messages?

lib/Middleware/WOPIMiddleware.php
```diff
@@ -97,6 +97,12 @@ public function afterException($controller, $methodName, \Exception $exception):
 			return new JSONResponse([], Http::STATUS_FORBIDDEN);
 		}
 
+		if ($exception instanceof \Exception) {
```
luka-nextcloud, line 100: As I see, this condition is always true. So, I guess we just need to return here.

lib/WOPI/Parser.php
```diff
@@ -21,40 +21,32 @@
...
 		$this->logger->debug('WOPI::getUrlSrc discovery: {discovery}', ['discovery' => $discovery]);
-		if (\PHP_VERSION_ID < 80000) {
-			$loadEntities = libxml_disable_entity_loader(true);
-			$discoveryParsed = simplexml_load_string($discovery);
-			libxml_disable_entity_loader($loadEntities);
-		} else {
-			$discoveryParsed = simplexml_load_string($discovery);
-		}
-
+		$discoveryParsed = simplexml_load_string($discovery);
```
luka-nextcloud, line 49: I think we should require `ext-simplexml` in composer.json since we are using function `simplexml_load_string()`.

lib/Middleware/WOPIMiddleware.php
juliushaertl: Good catch, I changed the condition slightly to only catch and return json if we actually want that for the WopiController

lib/WOPI/Parser.php
juliushaertl: Thanks, added 👍

Sprinterfreak (approved):
Tested on unicode domain using server 27.1.4. With this commit applied, richdocuments doesn't break nextcloud anymore.

Fixes #3086 again

luka-nextcloud (approved):

PR #3324, opened 2023-11-27
Do not hide speadsheet statusbar by default

pedropintosilva:
Status bar is crucial for productivity tool (specially when it comes
to office suites and in specific spreadsheets).

Context, this is a regression from:
- https://github.com/nextcloud/richdocuments/pull/2059
- fabed162ea07ac27c031f288f0fab6f027badf78

Even if,  and as noted in
https://github.com/nextcloud/richdocuments/pull/2053#issuecomment-1044371555,
the status bar was still being hidden for safari I think it's best to
don't break the spreadsheet experience for every browser. On top of
that I think the richdocuments frame suffer some changes and that
safari issue might not be present anymore.

Signed-off-by: Pedro Pinto Silva <pedro.silva@collabora.com>

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/helpers/coolParameters.js
+++ b/src/helpers/coolParameters.js
@@ -39,7 +39,7 @@ const getUIDefaults = () => {
 	let uiDefaults = 'TextRuler=' + textRuler + ';'
 	uiDefaults += 'TextSidebar=' + sidebar + ';TextStatusbar=' + statusBar + ';'
 	uiDefaults += 'PresentationSidebar=' + sidebar + ';PresentationStatusbar=' + statusBar + ';'
-	uiDefaults += 'SpreadsheetSidebar=' + sidebar + ';SpreadsheetStatusbar=' + statusBar + ';'
+	uiDefaults += 'SpreadsheetSidebar=' + sidebar + ';SpreadsheetStatusbar=true;'
 	uiDefaults += 'UIMode=' + uiMode + ';'
 	uiDefaults += 'UITheme=' + uiTheme + ';'
 	uiDefaults += 'SaveAsMode=' + saveAsMode + ';'


Darshan-upadhyay1110:
Tested working good. ByDefault statusbar is visible in spreadsheet

image

CC: @juliushaertl @pedropintosilva

juliushaertl (approved):

thebearon:
> Status bar is crucial for productivity tool (specially when it comes to office suites and in specific spreadsheets).
> 
> Context, this is a regression from:
> 
>     * [Revert "Do not hide speadsheet statusbar by default" #2059](https://github.com/nextcloud/richdocuments/pull/2059)
> 
>     * [fabed16](https://github.com/nextcloud/richdocuments/commit/fabed162ea07ac27c031f288f0fab6f027badf78)
> 
> 
> Even if, and as noted in [#2053 (comment)](https://github.com/nextcloud/richdocuments/pull/2053#issuecomment-1044371555), the status bar was still being hidden for safari I think it's best to don't break the spreadsheet experience for every browser. On top of that I think the richdocuments frame suffer some changes and that safari issue might not be present anymore.

The showing of status bar and subsequent revert was done due to the opposite reason:
In Safari, the window height was shorter than it should've been due to a bug, and setting the status bar to show made it so at least the sheet list would be visible (but not the status bar still). Ie. showing the status bar temporarily set to true to avoid this problem in Safari, and then was reverted as the issue got fixed in COOL and the workaround was unneeded.
So, it wasn't a regression, but phasing out a temporary workaround.

@pedropintosilva Is there a reason not to enable the status bar in all apps by default? Only doing it for spreadsheets seems like an inconsistency to me.

juliushaertl:
This was mainly due to designers input to have a more simplified interface with Nextcloud office. cc @jancborchardt

pedropintosilva:
> @pedropintosilva Is there a reason not to enable the status bar in all apps by default? Only doing it for spreadsheets seems like an inconsistency to me.

@thebearon : [Juliu's comment](https://github.com/nextcloud/richdocuments/pull/3324#issuecomment-1829532298) is a direct answer to your question

juliushaertl:
I think it makes sense to do some comparison here and pick what most users are used to on competitor (market leading) platforms. Zooming might for example also deserve a more prominent place. Or the status bar needs some revisiting to be less crowded (just from comparing with ms excel)

thebearon:
> @thebearon : [Juliu's comment](https://github.com/nextcloud/richdocuments/pull/3324#issuecomment-1829532298) is a direct answer to your question

My question was about the inconsistency that it is now enabled in one app, but not in the others, while @juliushaertl 's answer was about why it had been disabled in all apps in the first place. My concern here is that the different defaults will lead to confusing UIs to unfamiliar users (like the one you mentioned, looking for the zoom option where the status bar would be, except now it will be there in Calc, but not in the other apps).

Nevertheless, I am glad to hear this is being revisited. Should we open a GH issue on it?

juliushaertl:
Filed as https://github.com/nextcloud/richdocuments/issues/3327 so we don't forget

PR #3341, opened 2023-12-06
Fix opening file after rename in COOL

eszkadev:
There is a bug which happens after renaming a file inside Collabora Online (using fiels on the headerbar).

1. Open any document
2. Rename it so it will contain spaces
3. Close document
4. Try to open renamed document from the list Result: error, cannot find file

Notice that name was encoded so space become %20.
If you look at requests - the name of a file become encoded second time in one of them so it failed.
eg. file '1 1.odt' after rename does request:
http://192.168.0.108/remote.php/dav/files/admin/1%25201.odt

In Nextcloud 27 visible name doesn't contain %20 but still doesn't work.

### DIFF
--- a/src/view/FilesAppIntegration.js
+++ b/src/view/FilesAppIntegration.js
@@ -140,9 +140,9 @@ export default {
 	},
 
 	rename(newName) {
-		this.updateFileInfo(newName, Date.now())
+		this.updateFileInfo(decodeURIComponent(newName), Date.now())
 
-		this.fileName = newName
+		this.fileName = decodeURIComponent(newName)
 
 		if (this.handlers.rename && this.handlers.rename(this)) {
 			return


juliushaertl (approved):

eszkadev:
should I prepare PR for stable27 and master too?

juliushaertl:
> In Nextcloud 27 visible name doesn't contain %20 but still doesn't work.
> should I prepare PR for stable27 and master too?

That would be appreciated if it also works there

The code paths on master might be different as we rewrote the files app in parts with new frontend code in the server itself. Feel free to also just open an issue then I can have a closer look there.

eszkadev:
added port for stable27: https://github.com/nextcloud/richdocuments/pull/3343
For stable28 and master my fix doesn't work. It doesn't update file name on the file list at all...
Created issue: https://github.com/nextcloud/richdocuments/issues/3344

PR #3342, opened 2023-12-06
feat: Add headers for wasm support if CO has hasWASMSupport enabled

caolanm:
which was added in: https://github.com/CollaboraOnline/online/pull/7784

So we can conditionally add these headers iff WASM is enabled in coolwsd.xml

### TODO

- [ ] ...

### Checklist


juliushaertl:
Thanks, let me merge this into the branch of the other PR then and I'll take care of the rest.

Issue #3344, opened 2023-12-06
Opening a file after rename in COOL doesn't work


eszkadev:
**Describe the bug**
There is a bug which happens after renaming a file inside Collabora Online (using field on the headerbar).

**To Reproduce**
1. Open any document
2. Rename it so it will contain spaces
3. Close document using X
4. Try to open renamed document from the list
Result: error: cannot find file, notice that file didn't change the name.

**Expected behavior**
File list should be updated, so renamed file should be listed. Also after clicked it should open the file.

**Details:**
 - Nextcloud 28 and master/main are affected
- Fixes for 26 and 27:
https://github.com/nextcloud/richdocuments/pull/3341
https://github.com/nextcloud/richdocuments/pull/3343

PR #3345, opened 2023-12-07
fix: Make rename update the file list properly on files2vue

juliushaertl:
* Resolves: https://github.com/nextcloud/richdocuments/issues/3344
* Target version: main

### Summary

@eszkadev I've wondered a bit why the post message content would actually be url encoded, I don't think there is a real need, on the other hand it is proably easier adjusting here then in online which could be considered a breaking change.

Also the File_Rename post message does not seem to be documented at https://sdk.collaboraonline.com/docs/postmessage_api.html

### DIFF
--- a/src/view/FilesAppIntegration.js
+++ b/src/view/FilesAppIntegration.js
@@ -506,6 +506,10 @@ export default {
 		const node = await this.getFileNode()
 
 		if (node) {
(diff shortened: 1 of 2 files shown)


eszkadev (approved):
works as expected

PR #3356, opened 2023-12-12
fix: mention popup reappears after inserting the mention

Rash419:
- when COOL inserts the mention it sends 'UI_Mention' with 'type: selected', we don't want to send 'Action_Mention' for 'selected' type

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/view/Office.vue
+++ b/src/view/Office.vue
@@ -350,7 +350,9 @@ export default {
 				})
 				break
 			case 'UI_Mention':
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

juliushaertl:
@Rash419 could you rebase instead of merging for update? Otherwise the automated backports will fail

PR #3369, opened 2023-12-19
fix: Avoid 3px offset on public share links

juliushaertl:
Avoid that the background page is shining through at the bottom of public share links

### DIFF
--- a/css/files.scss
+++ b/css/files.scss
@@ -21,10 +21,10 @@
 #richdocumentsframe {
 	background-color: #fff;
 	width:100%;
-	height: calc(100vh - 50px + 3px);
+	height: 100vh;
 	display:block;
 	position:absolute;
-	top: -3px;
+	top: 0px;
 	left: 0;
 	z-index:110;
 	&.full {


juliushaertl:
> @juliushaertl does this have anything to do with the color accent bar on the top at the compact view? Cause that does not show for the share link in compact view (does show for compact view when logged in).

Before this patch the accent bar in compact mode was hidden due to this offset.

Now with this patch the behaviour is the following (which I think is the expected):
- Compact mode shows accent bar
- Tabbed mode shows no accent bar

jancborchardt (approved):

pedropintosilva:
Thanks @juliushaertl I'm a bit afraid of this. By he looks of it, it seems will affect the component height and position no matter if it's share link or not (unless that is only used in those cases).  I also would like to understand why we had before the additional calculation in the height and the top position. I wonder if it was fixing some sizing thingy and I', afraid it can become regression for some devices.

Here is what I have found out

* It seems that height was introduced with `Fix scrolling behaviour on Nextcloud >= 14 ` https://github.com/nextcloud/richdocuments/commit/907708f5ee167d2be99e6648527c7a3390a6c136
* The additional 3px  (height) and the -3px (top) were added to  `Improve loading state handling`  https://github.com/nextcloud/richdocuments/commit/c9cb1bd98224a53bd7db73c62ae4c81c6ea51f00

pedropintosilva:
my conclusion: good to merge but we need to make sure if doesn't get backported to a version where it will brake other things

Issue #3363, opened 2023-12-16
Error being thrown to log Error Call to undefined method OCA\Richdocuments\Service\DiscoveryService::refetch() Could not boot richdocuments: Call to undefined method OCA\Richdocuments\Service\DiscoveryService::refetch()


Jeefongithub:
**Describe the bug**

Getting the following error thrown to the log every few seconds. It doesn't seem to be affecting any of my workloads but I'd like to try and identify how to solve it

 Error Call to undefined method OCA\Richdocuments\Service\DiscoveryService::refetch()
Could not boot richdocuments: Call to undefined method OCA\Richdocuments\Service\DiscoveryService::refetch()

**To Reproduce**
1. Go to admin settings
2. scroll down to logging
3. error is presented

**Expected behavior**

unsure

**Screenshots**

![image](https://github.com/nextcloud/richdocuments/assets/53583661/65625d00-1dae-4d38-b5a7-5a0501df5ae2)

**Client details:**
 - OS: Debian 12
 - Browser Floorp (firefox fork)
 - Version 11.6.1
 - Device: Laptop

## Server details

**Operating system**: Debian 12

**Web server:** Apache 2.4.56

**Database:** MariaDB 10.5.21

**PHP version:** 8.2.13

**Nextcloud version:** 28.0.0

**Version of the richdocuments app** 8.3.0

**Version of Collabora Online** 25.5.602

#### Nextcloud log (data/nextcloud.log)

```
{
  "reqId": "aotXjCZpSlFyOnqxkmgS",
  "level": 4,
  "time": "2023-12-16T13:13:08+00:00",
  "remoteAddr": "192.168.55.4",
  "user": "Jeef",
  "app": "no app in context",
  "method": "GET",
  "url": "/index.php/apps/logreader/api/poll?lastReqId=",
  "message": "Could not boot richdocuments: Call to undefined method OCA\\Richdocuments\\Service\\DiscoveryService::refetch()",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/115.0",
  "version": "28.0.1.0",
  "exception": {
    "Exception": "Error",
    "Message": "Call to undefined method OCA\\Richdocuments\\Service\\DiscoveryService::refetch()",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/html/nextcloud/apps/richdocuments/lib/AppInfo/Application.php",
        "line": 154,
[...]

PR #3380, opened 2023-12-25
fix(Application.php): Update methods post-refactor

joshtrichards:
* Resolves: #3363
* Target version: main

### Summary

The refactor in #3315 swapped around a few methods and this updates Application.php to use the appropriate ones.

### Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -226,9 +226,9 @@ class Application extends App implements IBootstrap {
 			$discoveryService = $this->getContainer()->get(DiscoveryService::class);
 			$capabilitiesService = $this->getContainer()->get(CapabilitiesService::class);
 
-			$discoveryService->refetch();
-			$capabilitiesService->clear();
-			$capabilitiesService->refetch();
+			$discoveryService->resetCache();
+			$capabilitiesService->resetCache();
+			$capabilitiesService->fetchFromRemote();
 		}
 	}
 }


manuelkamp (approved):
this solved the issue

juliushaertl (approved):

juliushaertl:
Thanks @joshtrichards 

I'm puzzled why psalm has not thrown an error for this case, but that is not a blocker for getting this in of course.

gehkah:
The fix works for eliminating the errors. But It now hangs on uploading large files, like ISOs onto Nextcloud. Some instances crash. others just dont finalize the upload. I got NC instances where I did not apply this fix, there its working. on 3 instances, where I applied this fix, I got this problem. waiting on a new release.

joshtrichards:
@gehkah Please open a bug report so that we can get the info needed to solve that. I'm afraid a mere comment on the original PR isn't sufficient to reproduce + fix the problem you're reporting (if it is, indeed, even related)

gehkah:
> @gehkah Please open a bug report so that we can get the info needed to solve that. I'm afraid a mere comment on the original PR isn't sufficient to reproduce + fix the problem you're reporting (if it is, indeed, even related)

Dear Josh

I'm sorry for the circumstances. I upgraded to php8.3 and the problems are gone on my instances. I applied your fix to other NC instances and reproduces the problem unter php8.2 (VM, debian bullseye, apache2) but when I upgraded these to php 8.3, the problem no longer exists. No more 100% cpu usage or vm crash, I'm able to upload ISOs again. I'm fine with it. thank you very much!!

prismopensource:
Same problem over here on PHP 8.2 problem was fixed by migrating to PHP 8.3.

PR #3448, opened 2024-02-02
fix: Run move in lock scope of the user

juliushaertl:
Otherwise the move could fail if the files_lock app is enabled


elzody (approved):
Looks very sensible after reading the PHP API :)

Issue #2762, opened 2023-01-17
Make use of Reset_Access_Token post message


juliushaertl:
https://github.com/CollaboraOnline/online/pull/4745

- Make sure that it works on the web
- Emit a post message to mobile to reload the direct editing token and refresh the page
- Discuss implementation with the client teams

PR #3462, opened 2024-02-06
fix: Handle Reset_Access_Token to generate a new token and reload

juliushaertl:
Fix https://github.com/nextcloud/richdocuments/issues/2762


juliushaertl:
I actually misunderstood this message, it is supposed to be sent to the other direct where the client would ask Collabora to use a new token: https://github.com/CollaboraOnline/online/pull/4745

PR #3469, opened 2024-02-13
fix: Drop use of deprecated jquery load

juliushaertl:
Should make cypress pass again on main due to https://github.com/nextcloud/server/pull/43469

We should get rid of all of that jquery stuff of course but this fixes the regression breaking some file opening scenarios and makes CI passing again.


enjeck (approved):
Looks good to me :+1:

PR #3475, opened 2024-02-16
fix: emit allow attribute on iframe for the clipboard (fixes #3474)

### LINKED ISSUE #3474: iframe lacks allow attribute to use navigator.clipboard in Chrome
**Describe the bug**

Recent online.git tries to use navigator.clipboard in Chrome for better paste and paste special when you use the notebookbar buttons and not the keyboard. This works when COOL is running standalone, but not when nextcloud is served from one domain and COOL is from an other domain, due to how Chrome restricts cross-origin iframes.

**To Reproduce**
1. Open Writer in richdocuments
2. Type a character
3. Ctrl-A to select it
4. Ctrl-C to copy it
5. Press the paste dropdown on the notebookbar
6. Pick the paste menu item from the dropdown

**Expected behavior**
A Chrome popup asks if reading from the  clipboard should be allowed.

**Actual behavior**

Nothing happens, unless nextcloud and COOL is served from the same domain.

**Screenshots**

The developer console shows this:

"navigator.clipboard.read() failed: The Clipboard API has been blocked because of a permissions policy applied to the current document. See https://goo.gl/EuHzyv for more details."

**Other details**

Probably the fix is to emit an `allow="clipboard-read *; clipboard-write *"` attribute on the iframe element in src/view/Office.vue, will submit a PR to do that. At least that makes the clipboard popup show and after granting the permission, paste works in Chrome.

vmiklos:
As described at
<https://sites.google.com/a/chromium.org/dev/Home/chromium-security/deprecating-permissions-in-cross-origin-iframes> newer Chrome requires explicit markup for code in an iframe to execute JS that requires permissions, like clipboard.

If this markup is missing, then the user won't be even asked.  Use the wildcard syntax, because the COOL JS code in the iframe is not the initial src attribute value of the iframe, it gets changed later.

With this, a permission popup on paste shows up in Chrome even if the paste is perssed on the notebookbar, even if nextcloud is served from one domain and COOL is served from an other domain.

This fixes the document edit case; possibly it should be also added at all other places where the allowfullscreen attribute is used, which is not done in this commit.

* Resolves: #3474

### DIFF
--- a/src/view/Office.vue
+++ b/src/view/Office.vue
@@ -78,6 +78,7 @@
 			data-cy="coolframe"
 			scrolling="no"
 			allowfullscreen
+			allow="clipboard-read *; clipboard-write *"
 			class="office-viewer__iframe"
 			:style="{visibility: showIframe ? 'visible' : 'hidden' }"
 			:src="iframeSrc" />


vmiklos:
To clarify, this patch fixes the main, editing use-case. There are other places where allowfullscreen is used, and possibly this new allow attribute should be added there as well. One place is the read-only view of past revisions. I think we currently don't support copy for read-only views, so it's not useful to have this attribute there (although it doesn't hurt, either).

juliushaertl (approved):

vmiklos:
> Yes, would be good to just apply that consistently to any iframe:

Will do that in a next PR.

PR #3480, opened 2024-02-20
fix: emit allow attribute on all iframes for the clipboard (related t…

vmiklos:
…o #3474)

The unhandled cases were:

- read-only view

- view a past revision

- for completeness, also adapt the iframe in files.js

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -206,7 +206,7 @@ const documentsMain = {
 				+ '</form>'
 
 			// iframe that contains the Collabora Online Viewer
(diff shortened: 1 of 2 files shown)


vmiklos:
@juliushaertl could you please review this? Thanks.

This adds the 'allow' attribute at the remaining places. I'm not sure what the files.js case is about, so I just did that for completeness. If you have a guess, I would be happy to learn. :-)

Other than that, please also merge the PR if you approve, I'm not a maintainer in this repo.

juliushaertl:
> This adds the 'allow' attribute at the remaining places. I'm not sure what the files.js case is about, so I just did that for completeness. If you have a guess, I would be happy to learn. :-)

That is still some legacy code as the public share links of files are not migrated to Vue.js yet, so we have two different implementations for launching the iframe. Hopefully gets obsolete at some point and can just use Viewer.vue

juliushaertl (approved):

Issue #3487, opened 2024-02-20
"Open file locally" doesn't work


lpranam:
**Is your feature request related to a problem? Please describe.**
Clicking on 'Open file locally' doesn't do anything and gives an internal server error 500. 

**Describe the solution you'd like**
Fixing this problem is great, but also it would be nice if there is a way to hide this button.

juliushaertl:
@lpranam I could not reproduce the 500 error but https://github.com/nextcloud/richdocuments/pull/3489 should address both the config option as well as a fix for the actual opening that I've seen

Issue #3299, opened 2023-11-16
Edit locally does not work with WOPI allow list


juliushaertl:
Edit locally tries to unlock the file using the WOPI endpoints which does not work if you have a wopi allow list configured.

juliushaertl:
Having a separate endpoint for frontend unlock sounds the most sane to me.

One unresolved question, how does the unlock behave when other users are also joined the editing session

pedropintosilva:
Something else I have discovered just outside of collaborative editing:

1. In the files view `…` -> `Edit locally`
2. Browser's native dialog appears 
3. Either choose cancel Or open (but make sure you nextcloud client doesn't have that instance configured)
4. Open file locally dialog appears and it's impossible to close the dialog
5. Press close button or `Continue editing online` -> see the web console errors
![image](https://github.com/nextcloud/richdocuments/assets/65948705/0a66d1ae-3aa1-44c8-92ad-95c505308465)


```javascript
15:21:27.767 Uncaught Error: Opening the viewer with a single string parameter is deprecated. Please use a destructuring object instead
    NextJS 79
[...]
15:21:37.913 Uncaught Error: Opening the viewer with a single string parameter is deprecated. Please use a destructuring object instead
    NextJS 77
```

<details>
15:21:08.909 [DEBUG] text: sending steps  
Object { app: "text", uid: "pedro", level: 0, steps: [], awareness: "AWABwqLizQEGWHsidXNlciI6eyJuYW1lIjoicGVkcm8iLCJjbGllbnRJZCI6NDMxNTI2MjEwLCJjb2xvciI6IiNiOGJlNjgiLCJsYXN0VXBkYXRlIjoxNzA0MjA1MTg1fX0=", version: 17845 }
text-init.js:2:13255
15:21:08.919 XHRPOST
https://tech-preview.nextcloud.com/apps/text/session/538194/push
[HTTP/1.1 200 OK 67ms]

15:21:09.009 [DEBUG] text: synced  
Object { app: "text", uid: "pedro", level: 0, version: 17845, steps: [] }
text-init.js:2:13255
15:21:09.825 [DEBUG] text: [PollingBackend] Fetching steps 
Object { app: "text", uid: "pedro", level: 0 }
text-init.js:2:13255
15:21:09.832 XHRPOST
https://tech-preview.nextcloud.com/apps/text/session/538194/sync
[HTTP/1.1 200 OK 85ms]

15:21:09.923 [DEBUG] text: synced  
[...]

PR #3489, opened 2024-02-21
Fix open locally with files lock and wopi allow list

juliushaertl:
- fix: Add config option to disable edit locally
- fix: Allow to unlock through separate endpoint for edit locally

Fix https://github.com/nextcloud/richdocuments/issues/3487
Fix https://github.com/nextcloud/richdocuments/issues/3299

### Summary

We add a config value for disabling but also address that unlocking the file was no longer working with a WOPI allow list. We now move this to a separate endpoint to handle any prepare steps we need to do before moving to the local editor.

In a second iteration we can further extend this to update the other views to read only or notify them through Collabora.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -77,6 +77,7 @@ return [
 	'ocs' => [
 		// Public pages: new file creation
 		['name' => 'documentAPI#create', 'url' => '/api/v1/file', 'verb' => 'POST'],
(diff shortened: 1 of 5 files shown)


max-nextcloud (approved):
Code changes look good. did not test it though.

lpranam:
is it going to backport to 26? should I backport it manually? @juliushaertl

juliushaertl:
Sorry I've missed that, the automatic one failed so a manual one is needed. I can check, but if you want to help preparing that, just opening a pr is of course much appreciated.

Issue #3434, opened 2024-01-21
Preview generation not working when wopi_url has a trailing slash


krzys-h:
**Describe the bug**
When configuring the app with an external Collabora server (hosted on the same machine using Docker, both behind a Traefik reverse proxy), I entered the URL such as `https://office.example.com/` (with a valid domain ofc, redacted for privacy). Everything worked, but document previews in the file listing did not generate. After changing the URL to `https://office.example.com`, the issue went away.

The issue seems to come from here:
https://github.com/nextcloud/richdocuments/blob/97ec712668d7b125fa33ccc9c39e593da057ceb6/lib/Preview/Office.php#L94
When I had the trailing slash in the config, the URL generated was `https://office.example.com//lool/convert-to/png`, and the double slash in the URL not correctly handled by Collabora, resulting in a `400 Bad Request` response. Other places in the code seem to use the newer `/cool` endpoint, so perhaps this is something specific to the older `/lool` endpoint handling?

**To Reproduce**
1. Configure a Collabora server with a trailing slash in the path
2. Create a new .docx file, or upload one
3. Look at the directory listing - the preview is missing

**Expected behavior**
The preview generation should work

## Server details

**Operating system**: Ubuntu 22.04.3 LTS, running the official `nextcloud:apache` docker image

**Web server:** Apache from the docker image, behind a Traefik reverse proxy

**Database:** mysql

**PHP version:** 8.2.14

**Nextcloud version:** 28.0.1

**Version of the richdocuments app** 8.3.1

**Version of Collabora Online** 23.05.7.2

**Configuration of the richdocuments app**
```
{
    "apps": {
        "richdocuments": {
            "disable_certificate_verification": "",
            "doc_format": "ooxml",
            "enabled": "yes",
            "installed_version": "8.3.1",
            "public_wopi_url": "https:\/\/office.example.com/",
            "types": "prevent_group_restriction",
            "wopi_url": "https:\/\/office.example.com/"
        }
    }
}
```

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
[...]

PR #3491, opened 2024-02-21
Fix preview URLs and migrate providers to new API

juliushaertl:
- fix: Preview url generation and cleanup
  - We use a trimmed url now (fix https://github.com/nextcloud/richdocuments/issues/3434)
  - Move from old url pattern `lool` to `cool`
  - Move to LoggerInterface and constructor property promition
- chore: Move to preview provider v2
  - Move to new preview provider API to register lazy in Application bootstrap


blizzz (approved):
Looks good, did not test

PR #3493, opened 2024-02-22
Fixes #3492 joining after document is renamed inside editor

### LINKED ISSUE #3492: Joining document after it was renamed inside editor causes document conflict and potential data loss
**Describe the bug**
User can land in two instances of the same document when using rename inside Collabora Online.

**To Reproduce**
1. Open document in Collabora Online
2. Rename it inside editor using title bar input field
3. Open new tab in the browser and open the same document second time from NC file picker
Result: Tow sessions doesn't see each other, after saving conflict dialog appears

**Expected behavior**
Two users join the same session

eszkadev:
Bug:
1. Open document
2. Rename in Collabora Online using input in the title bar
3. In the new tab open the same document from file picker in NC Result: 2 sessions are in different instances of the same document

getCallbackBaseUrl() puts "/" at the end, but WOPISrc was adding additional one:
const wopiurl = getCallbackBaseUrl() + '/ ...

Then reconnected session after rename had single "/index.php" but new session used "//index.php" what caused that COOL server didn't consider both as the same identifier

### DIFF
--- a/src/helpers/url.js
+++ b/src/helpers/url.js
@@ -34,7 +34,7 @@ const getSearchParam = (name) => {
 
 const getCallbackBaseUrl = () => {
 	const callbackUrl = Config.get('wopi_callback_url')
-	return callbackUrl || window.location.protocol + '//' + window.location.host + getRootUrl() + '/'
+	return callbackUrl || window.location.protocol + '//' + window.location.host + getRootUrl()
 }
 
 const getWopiSrc = (fileId) => {


src/helpers/url.js
```diff
@@ -40,7 +40,7 @@ const getCallbackBaseUrl = () => {
 const getWopiSrc = (fileId) => {
 	// WOPISrc - URL that loolwsd will access (ie. pointing to ownCloud)
 	// index.php is forced here to avoid different wopi srcs for the same document
-	const wopiurl = getCallbackBaseUrl() + '/index.php/apps/richdocuments/wopi/files/' + fileId
+	const wopiurl = getCallbackBaseUrl() + 'index.php/apps/richdocuments/wopi/files/' + fileId
```
thebearon, line 43: Wouldn't it be better to ensure that `getCallbackBaseUrl()` returns URL without trailing `/`? I remember similar problems back in the day when the Collabora Online URL was misbehaving depending on whether the admins added trailing `/` or not, and IIRC it ended up being always removed (but I could be wrong).

If that's the case, perhaps not adding the trailing `/` in `getCallbackBaseUrl() ` would be consistent (but it also has to be ensured that `wopi_callback_url` is stored without a trailing `/`):
`return callbackUrl || window.location.protocol + '//' + window.location.host + getRootUrl() + '/'`
https://github.com/nextcloud/richdocuments/blob/b784eeb24dd1dfa32fdade11bdc814b1d312229e/src/helpers/url.js#L37

@juliushaertl Can you please share your thoughts?

juliushaertl: Yeah I think it would be more consistent if we trim any trailing slash from the base. There are other places where this is probably wise to do

eszkadev: changed base then

juliushaertl (approved):

PR #3505, opened 2024-02-27
feat/range requests

juliushaertl:
- chore: Move WopiController to constructor property promotion
- feat: rudimentary support for range requests when fetching the file content

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


pedropintosilva:
I think this was brought up by @mmeeks sometime ago (that we could try to split the document and downloaded its parts concurrently). @Ashod you are probably the right the person to look at this (to check if it brings a performance benefit when integrating with Collabora Online).  What do you think?

Notes:
* It doesn’t parse the Range header fully standard compatible, only one range is supported for now, however should be enough to proof that this brings a performance benefit when integrating with Collabora Online
* There is already ongoing work to load the file earlier (at the request of the iframe, @Ashod was working on this, I think)

PR #3513, opened 2024-03-07
fix: Avoid absolute positioning on interactive widgets

juliushaertl:
Fix positioning with https://github.com/nextcloud/server/pull/44012


skjnldsv (approved):
:see_no_evil:

luka-nextcloud (approved):

PR #3515, opened 2024-03-08
Fix open locally with files lock and wopi allow list (backport stable26)

lpranam:
- fix: Add config option to disable edit locally
- fix: Allow to unlock through separate endpoint for edit locally

Fix https://github.com/nextcloud/richdocuments/issues/3487
Fix https://github.com/nextcloud/richdocuments/issues/3299

Summary
We add a config value for disabling but also address that unlocking the file was no longer working with a WOPI allow list. We now move this to a separate endpoint to handle any prepare steps we need to do before moving to the local editor.

In a second iteration we can further extend this to update the other views to read only or notify them through Collabora.

Configuration option to disable local editing
```
occ config:app:set richdocuments open_local_editor --value="no"
```

* Target version: stable26

backport from #3489
### Checklist

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -72,6 +72,7 @@ return [
 	'ocs' => [
 		// Public pages: new file creation
 		['name' => 'documentAPI#create', 'url' => '/api/v1/file', 'verb' => 'POST'],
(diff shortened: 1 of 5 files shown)


lpranam:
@juliushaertl can you please take on this?  😓

juliushaertl (approved):

PR #3445, opened 2024-02-01
fix #3209 collabora open when nextcloud is in a iframe

jnoel974:
* Resolves: #3209
* Target version: main

### Summary
collabora open when nextcloud is in a iframe

### Checklist

- [x] Code is properly formatted
- [ ] Sign-off message is added to all commits
- [x] Documentation (manuals or wiki) has been updated or is not required

PR #3539, opened 2024-03-21
match dark theme of system if no nextcloud theme specified

lpranam:
problem example:
a new user is created who uses system dark mode, so nextcloud will be in dark mode but collabora would not be turned to dark mode as all the themes(ie:themeDark and themeDarkHighcontrast) value are undefined

* Target version: main

### Checklist

### DIFF
--- a/src/helpers/coolParameters.js
+++ b/src/helpers/coolParameters.js
@@ -28,12 +28,12 @@ const getUIDefaults = () => {
 	const textRuler = 'false'
 	const sidebar = 'false'
 	const saveAsMode = 'group'
-	const uiMode = defaults.UIMode ?? 'notebookbar' // or notebookbar
+	const uiMode = defaults.UIMode ?? 'notebookbar'
 
 	const systemDarkMode = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
-	// FIXME For now we need to access parent here as the public page loaded for collabora doesn't have the user theme
-	const nextcloudDarkMode = parent.document.body.dataset.themeDark === '' || parent.document.body.dataset.themeDarkHighcontrast === ''
-	const matchedDarkMode = parent.document.body.dataset.themeDefault === '' ? systemDarkMode : nextcloudDarkMode
+	const dataset = (document.body.dataset.themes ? document.body.dataset : parent?.document.body.dataset) ?? {}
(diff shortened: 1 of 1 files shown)


src/helpers/coolParameters.js
```diff
@@ -32,7 +32,7 @@ const getUIDefaults = () => {
 
 	const systemDarkMode = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
 	// FIXME For now we need to access parent here as the public page loaded for collabora doesn't have the user theme
-	const nextcloudDarkMode = parent.document.body.dataset.themeDark === '' || parent.document.body.dataset.themeDarkHighcontrast === ''
+	const nextcloudDarkMode = parent.document.body.dataset.themeDark === '' || parent.document.body.dataset.themeDarkHighcontrast === '' || (parent.document.body.dataset.themes === '' && systemDarkMode)
```
juliushaertl, line 35:
```suggestion
	... (parent.document.body.dataset.themes === '' && systemDarkMode)
```

I cannot see why this would be empty, can you provide some more clear steps to replicate this issue? Is this something on an older Nextcloud release?

lpranam: How I reproduced it:
1. My system is in dark mode
2. created a new user
3. new user should have dark nextcloud theme by default
4. just now open any doc

juliushaertl:
Thanks a lot @lpranam 

There is indeed some inconsistency between the default for a new user or when manually setting the default, I adapted your change a bit to also properly work with public share links and to take the change from https://github.com/nextcloud/richdocuments/pull/3445/files into account

juliushaertl (approved):

Ornanovitch:
Interesting, thanks for this enhancement :)

However, while there is no doubt this will be great [in the near future](https://github.com/CollaboraOnline/online/issues/6449), Collabora still has [a lot of problems with its "full dark" mode](https://github.com/CollaboraOnline/online/labels/dark-mode) and a lot of NC users might complain soon. How could an admin "force" the light mode no in the meantime?

cc @juliushaertl @lpranam

HolgerHees:
also after the latest fix, the error still happens. I'm running latest richdocument app (8.3.4)

I have embedded nextcloud inside an iframe like

outer frame is mydomain.de
nextcloud frame is nextcloud.mydomain.de

and I get the error "DOMException: Permission denied to access property "document" on cross-origin object"

I thin you should catch this kind of errors and just ignore, because it is not allowed to access the outer frame

PR #3558, opened 2024-04-03
fix(WopiController): In `putFile()` set editor user as user scope

mejo-:
This fixes saves in shares being attributed to the share owner. With this change, they're attributed to 'remote user' in the activity stream.

### Checklist


juliushaertl (approved):

juliushaertl:
As already expected, this causes the scenario from https://github.com/nextcloud/server/pull/44294 to fail in richdocuments as well, but the server PR fixes it as well.

Issue #3281, opened 2023-11-10
File gets truncated on save when quota is exceeded


thebearon:
- Set up a certain quota for a user,
- Add files to almost fill the quota,
- Inside a document opened in Collabora Online, embed a large image that gets the user beyond the quota limit,
- Save document.

-> Error message: `Document cannot be saved, please check your permissions`, and `Document has been changed` popup right afterwards.
More importantly, the file itself is truncated in storage.

Expectation:
- the file should be untouched in storage (Nextcloud),
- the user should be informed the document cannot be saved due to exceeded quota, offering them the option to download instead (COOL).

juliushaertl:
Thanks for reporting. I would not expect the Nextcloud storage layers to truncated the file when the storage throws an out of quota exception but I might be wrong there. I'll check that on our side.

For the error, we can definitely report back such a case on the WOPI request. It actually would be good to know the size upfront to we do not even attempt to write larger files if they don't fit.

- [ ] Check if Content-Length is already passed by Collabora and if we can validate it against the free space
- [ ] Check why 

Regarding reporting back we might need to agree on a response status and format to give back a custom error message as WOPI itself only defines some basic status codes and no format for a response body https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/putfile#status-codes However I think error message hand over to also show something to the user would be nice in any failure case

juliushaertl:
We can check for the quota with something like this:

Server storage layer does not check the quota upfront as is just does a stream copy without being aware of the size.

```
diff --git a/lib/Controller/WopiController.php b/lib/Controller/WopiController.php
index a9cf99ac..5e31e318 100644
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -70,6 +70,7 @@ use OCP\Share\IShare;
 use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\NotFoundExceptionInterface;
 use Psr\Log\LoggerInterface;
+use Sabre\DAV\Exception\InsufficientStorage;

 class WopiController extends Controller {
        /** @var IRootFolder */
@@ -107,6 +108,7 @@ class WopiController extends Controller {

        // Signifies LOOL that document has been changed externally in this storage
        public const LOOL_STATUS_DOC_CHANGED = 1010;
+       public const LOOL_STATUS_NO_SPACE = 3000;

        public const WOPI_AVATAR_SIZE = 64;

[...]

PR #3562, opened 2024-04-04
don't alter file when there is not enough free space

grnd-alt:
* Resolves: #3281 
* Target version: main

### Summary
When Updating files via WOPI the file used to be written from the beginning just to realize mid writing that there's not enough space and throw an error. Therefore the entire doc gets broken and mostly unusable.

Now: When a File is to be updated it gets checked if it would exceed the free space on storage/quota, if not an error is thrown and the file is left as is.

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -485,7 +485,13 @@ class WopiController extends Controller {
 
 			$content = fopen('php://input', 'rb');
 
+			$freespace = $file->getParent()->getFreeSpace();
+			$contentLength = (int)$this->request->getHeader('Content-Length');
+
 			try {
+				if ($freespace >= 0 && $contentLength > $freespace) {
+					throw new \Exception('Not enough storage');
+				}
 				$this->wrappedFilesystemOperation($wopi, function () use ($file, $content) {
 					return $file->putContent($content);
 				});


lib/Controller/WopiController.php
```diff
@@ -485,10 +485,21 @@ public function putFile($fileId,
 
 			$content = fopen('php://input', 'rb');
 
+			// Get expected FileSize after writing
+			$storage = \OC_Helper::getStorageInfo($path);
```
juliushaertl, line 489: Unless I oversee something, it should be possible to use the file parent to get the free space through the public (OCP) API:

Something like:
```
$freeSpace = $file->getParent()->getFreeSpace();
```

lib/Controller/WopiController.php
```diff
@@ -485,10 +485,21 @@ public function putFile($fileId,
...
-					return $file->putContent($content);
-				});
+				if ($finalSize <= $storage['free']){
+					$this->wrappedFilesystemOperation($wopi, function () use ($file, $content) {
+						return $file->putContent($content);
+					});
+				}
+				else {
+					throw new \Exception('Not enough storage');
+				}
```
juliushaertl, line 502: I personally find early return or throw a bit more readable:

```suggestion
				if ($finalSize > $storage['free']){
					throw new \Exception('Not enough storage');
				}
				
				$this->wrappedFilesystemOperation($wopi, function () use ($file, $content) {
					return $file->putContent($content);
				});
```

lib/Controller/WopiController.php
```diff
@@ -485,10 +485,21 @@ public function putFile($fileId,
 
 			$content = fopen('php://input', 'rb');
 
+			// Get expected FileSize after writing
+			$storage = \OC_Helper::getStorageInfo($path);
+			fseek($content,0,SEEK_END);
+			$finalSize = ftell($content) + $file->getSize();
+			rewind( $content );
```
juliushaertl: I'm not fully sure but remember something around the input stream being only readable once, however I could only find a reference to this prior to PHP 5.6.

https://github.com/php/doc-en/commit/cdaea0421544885f02ff3d36bd203dc01b78299e#diff-f476ea6ab31592691b4878cd08dd817003bef1fa186317914fc0e94d51bca52fL56-L62

@icewind1991 Do you happen to know if this works fine?

grnd-alt: @icewind1991 so we should find another way to read out the size, right? is there any better way to do it? Maybe read out the stream once and than pass it on as a string instead of a stream?

juliushaertl: Passing into a string will potentially cause issues if the file size exceeds the memory limit of the php process.

We could write it to a temporary file and use the size of that but that has some overhead in copying and writing/reading that file of course.

As an alternative we could check if Collabora actually sends a Content-Length header that we can just use, otherwise we may want to copy to a temporary stream and use that.

grnd-alt: CONTENT_LENGTH header is indeed set, changed it to use that.

lib/Controller/WopiController.php
```diff
@@ -485,7 +485,13 @@ public function putFile($fileId,
 
 			$content = fopen('php://input', 'rb');
 
+			$freespace = $file->getParent()->getFreeSpace();
+			$contentLength = $this->request->getHeader('CONTENT_LENGTH');
```
juliushaertl, line 489:
```suggestion
			$contentLength = (int)$this->request->getHeader('Content-Length');
```

lib/Controller/WopiController.php
```diff
@@ -485,7 +485,13 @@ public function putFile($fileId,
...
+
 			try {
+				if ($contentLength > $freespace) {
```
juliushaertl, line 492:
```suggestion
				if ($freespace >= 0 && $contentLength > $freespace) {
```

Otherwise if there is no quota for various reasons we would always fail as those are indicated by negative numbers https://github.com/nextcloud/server/blob/e62e9e3dbf1a2573554b1a9eabbf5b59b652dae6/lib/public/Files/FileInfo.php#L53-L63

lib/Controller/WopiController.php
juliushaertl: The naming is mostly to match the actual http header, uppercase and underscore are then converted in our getHeader method in the server to match the PHP pattern

juliushaertl (changes requested):
Small comments, but otherwise tested and works

juliushaertl (approved):

juliushaertl:
Can you squash the commit to have a clean history?

Issue #3579, opened 2024-04-11
Wrong protocol for TemplateSource URL when using nextcloud behind a proxy


k-jell:
**Describe the bug**
I run nextcloud behind a proxy (traefik) which does SSL-termination and another proxy for authentification:

` browser ---> Traefik (https termination) ---> Auth Proxy ---> Nextcloud`

collabora is setup in a similar way. But internally collabora and nextcloud can communicate over http using the internal ip addresses.

I have set `overwriteprotocol` to `https`. Now whenever I create a new document using the "New" Button and try to open it for the first time I get an error: 
```
Failed to read document from storage, please try to load the document again.

Please check the Collabora Online server log for more details and make sure that Nextcloud can be reached from there.
```
The problem is that collabora gets the wrong URL from nextcloud (https instead of http which is used for the internal communication between collabora and nextcloud, I am using the sociallogin app):

```
wsd-00009-00075 2024-04-11 13:40:45.834391 +0000 [ docbroker_004 ] DBG  WOPI::CheckFileInfo: {"BaseFileName":"New document (1).odt","DisableCopy":false,"DisableExport":false,"DisablePrint":false,"DownloadAsPostMessage":false,"EnableInsertRemoteImage":true,"EnableRemoteLinkPicker":true,"EnableShare":true,"HideExportOption":false,"HidePrintOption":false,"HideUserList":"","IsUserLocked":false,"LastModifiedTime":"2024-04-11T13:40:34.000000Z","OwnerId":"admin.example","PostMessageOrigin":"https://nextcloud28.example.org/","Size":1268,"SupportsLocks":false,"SupportsRename":true,"TemplateSource":"https://10.66.60.1:9968/index.php/apps/richdocuments/wopi/template/112?access_token=FtHZIfsoGHjVmfG6EJ8uLmagI7CAkxJY","UserCanNotWriteRelative":false,"UserCanRename":true,"UserCanWrite":true,"UserExtraInfo":{"avatar":"https://10.66.60.1:9968/avatar/admin.example/64","is_admin":true},"UserFriendlyName":"admin.example","UserId":"admin.example","UserPrivateInfo":{"ZoteroAPIKey":""},"Version":"0"}| wsd/Storage.cpp:835
```
Then collabora fails to open the document (see logs).

If I disable `overwriteprotocol` it does work, but this brings other problems (the authproxy redirect having the wrong scheme).

All other communication with collabora works fine. Is there any way to fix this?

**Expected behavior**
[...]

PR #3575, opened 2024-04-10
[stable27] fix: Allow configuring the wopi_callback_url

juliushaertl:
This is a minimal invasive backport of the parts of https://github.com/nextcloud/richdocuments/pull/3315 that allow configuring a separate callback URL for WOPI that can be used to use a different URL for requests Collabora is performing towards Nextcloud

```
occ config:app:set richdocuments wopi_callback_url --value="http://stable27"
```

PR #3580, opened 2024-04-12
fix: Use wopi callback url for all urls handed over to Collabora

juliushaertl:
Fix https://github.com/nextcloud/richdocuments/issues/3579

Some URLs were not respecting the wopi_callback_url setting that we have in place now to configure the URL that Collabora will use to talk to Nextcloud, so this PR adapts those.

### DIFF
--- a/src/helpers/url.js
+++ b/src/helpers/url.js
@@ -38,7 +38,7 @@ const getCallbackBaseUrl = () => {
 }
 
 const getWopiSrc = (fileId) => {
-	// WOPISrc - URL that loolwsd will access (ie. pointing to ownCloud)
+	// WOPISrc - URL that Collabora will use to access Nextcloud
 	// index.php is forced here to avoid different wopi srcs for the same document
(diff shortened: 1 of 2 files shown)


src/helpers/url.js
```diff
@@ -38,7 +38,7 @@ const getCallbackBaseUrl = () => {
 }
 
 const getWopiSrc = (fileId) => {
-	// WOPISrc - URL that loolwsd will access (ie. pointing to ownCloud)
+	// WOPISrc - URL that Collabora will use to access Nextcloud
```
juliushaertl, line 41: @blizzz Sneaked this in from your comment in https://github.com/nextcloud/richdocuments/pull/3575 😁

blizzz (approved):
nitpick: URL generation can be put in an new short method, so the next time it needs just only one adjustment. Not a blocker.

juliushaertl:
Cypress failure seems related, will check

PR #3597, opened 2024-04-15
[stable26] fix: Use Close_Session post message to properly end the Collabora editing before opening locally

juliushaertl:
If we edit locally we want to ensure that the document is saved and all sessions are removed before handing over the file to the desktop client. The previous approach was somewhat flaky and could lead to opening the file locally too early. The Close_Session post message will ensure that the file is properly terminated on the Collabora side and the lock on it is removed. 

Further we can use the Session_Closed post message on other editing sessions to show a hint to the user why the document was auto-closed.


max-nextcloud (approved):
Just did a quick pass over the code. Let me know if this needs more testing etc.

Issue #3599, opened 2024-04-15
refactor to use office.vue component


elzody:
Currently the richdocuments app uses [document.js](https://github.com/nextcloud/richdocuments/blob/main/src/document.js) to handle loading individual documents, which contains legacy code. This should be cleaned up and moved to use the [Office.vue](https://github.com/nextcloud/richdocuments/blob/main/src/view/Office.vue) component which is more suitable and makes things more modular.

Issue #3557, opened 2024-04-02
Accessing a document from a public shared folder does not ask for guest username


elzody:
**Describe the bug**
When a document is inside of a publicly shared folder, guest users who access the document are not asked to input a guest username. It simply defaults to "Anonymous Guest." On the other hand, when accessing a document which was publicly shared individually, it asks for a guest username, as expected.

**To Reproduce**
1. Create a folder in the Files app and insert a blank document into it
2. Create a public share link for the folder
3. Open the shared folder (not logged in)
4. Open the document inside
5. It will open the document directly

**Expected behavior**
The expect behavior is that, when opening a document as a guest, no matter where the document is (within a shared folder, or shared individually), the user will be asked to input a guest username. This makes identifying people much easier, as when there are several people editing, you don't know who it is.

**Client details:**
 - Browser: Firefox, Chrome
 - Version: Affects v27 - v30
 - Device: Desktop

**Notes**
Probably related to [this pull request](https://github.com/nextcloud/richdocuments/pull/3152). Also worth to mention [this pull request](https://github.com/nextcloud/richdocuments/issues/3368)

PR #3600, opened 2024-04-15
refactor: move to office.vue component

elzody:
* Resolves: #3599

### Summary
Removes `document.js` which contains legacy code and is difficult to maintain, replacing it with the `Office.vue` component.

This will help complete #3557 

### Checklist

### DIFF
--- a/cypress/e2e/share-link.js
+++ b/cypress/e2e/share-link.js
@@ -51,11 +51,11 @@ describe('Public sharing of office documents', function() {
 							cy.spy(win, 'postMessage').as('postMessage')
 						},
 					})
-					cy.waitForCollabora(true)
+					cy.waitForCollabora()
 					cy.get('@loleafletframe').within(() => {
 						cy.get('#closebutton').click()
 					})
-					cy.get('#viewer', { timeout: 5000 }).should('not.exist')
+					cy.get('#viewer', { timeout: 10000 }).should('not.exist')
 				})
 			})
 		}
--- a/cypress/e2e/share-link.js
+++ b/cypress/e2e/share-link.js
@@ -55,7 +55,7 @@ describe('Public sharing of office documents', function() {
 					cy.get('@loleafletframe').within(() => {
 						cy.get('#closebutton').click()
 					})
-					cy.get('#viewer', { timeout: 10000 }).should('not.exist')
+					cy.get('#viewer', { timeout: 5000 }).should('not.exist')
 				})
 			})
 		}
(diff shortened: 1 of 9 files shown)


lib/Listener/LoadPublicViewerListener.php
```diff
@@ -0,0 +1,50 @@
...
+	public function handle(Event $event): void {
+		if (!$event instanceof BeforeTemplateRenderedEvent) {
+			return;
+		}
+
+		if ($event->getResponse()->getRenderAs() !== TemplateResponse::RENDER_AS_PUBLIC) {
+			return;
+		}
+
+		Util::addScript(Application::APPNAME, 'richdocuments-public');
```
juliushaertl, line 48: We can probably merge this into the ShareLinkListener which already handles the case of public links

juliushaertl:
The failure of direct.spec.js is expected as the full document.js was removed. This and federated editing are two other leftover cases where it is in use which are probably a bit more tricky to migrate. Let's maybe chat tomorrow about those, i need to think a bit about how we can move those over as well

juliushaertl:
@elzody I pushed a small revert for `Cypress.config` to `Cypress.env` in 

We use an environment variable to pass in the collabora url for CI and therefore it was failing with just using the cypress config option:
http://github.com/nextcloud/richdocuments/blob/main/.github/workflows/cypress.yml#L17

Then the failure of the settings test also caused a failure on the other one as the collabora url has no longer been properly set

For adjusting URLs on local runs I usually just use:
```
CYPRESS_baseUrl=https://nextcloud.local/index.php CYPRESS_collaboraUrl=https://collabora.local npx cypress run --e2e --spec cypress/e2e/share-internal.spec.js
```

cypress/e2e/share-link.js
```diff
@@ -51,7 +51,18 @@ describe('Public sharing of office documents', function() {
...
+					cy.waitForCollabora()
+					cy.get('@postMessage', { timeout: 20000 }).should(spy => {
+						const calls = spy.getCalls()
+						const findMatchingCall = calls.find(call => call.args[0].indexOf('"MessageId":"App_LoadingStatus"') !== -1)
+						if (!findMatchingCall) {
+							return expect(findMatchingCall).to.not.be.undefined
+						}
+						const object = JSON.parse(findMatchingCall.args[0])
+						expect(object.Values).to.have.property('Status', 'Initialized')
+					})
```
juliushaertl, line 64: We might want to move this into a cypress command that can be reused in the future. So we can assert post messages in other tests as wel. For now fine to just keep it as it is i think.

juliushaertl (approved):

PR #3601, opened 2024-04-15
fix: Use Close_Session post message to properly end the Collabora editing before opening locally

juliushaertl:
Forward port of https://github.com/nextcloud/richdocuments/pull/3597

If we edit locally we want to ensure that the document is saved and all sessions are removed before handing over the file to the desktop client. The previous approach was somewhat flaky and could lead to opening the file locally too early. The Close_Session post message will ensure that the file is properly terminated on the Collabora side and the lock on it is removed. 

Further we can use the Session_Closed post message on other editing sessions to show a hint to the user why the document was auto-closed.

Steps to reproduce (with some back luck in timing):
1. Open a document in Collabora in multiple browser tabs
2. One tab clicks the open locally button so that the desktop client will take over editing

Now before this change there was a high chance that if the save and termination of individual sessions took longer, Collabora hadn't fully closed the document and may have ended in a inconsistent state (either the client failed to lock or Collabora hasn't finished writing the document). With the new post message handling the Collabora side should be fully terminated before we handover.

### DIFF
--- a/cypress/e2e/integration.spec.js
+++ b/cypress/e2e/integration.spec.js
@@ -101,20 +101,28 @@ describe('Nextcloud integration', function() {
 		cy.openFile('document.rtf')
 	})
 
-	it.skip('Open locally', function() {
+	it('Open locally', function() {
 		cy.get('@loleafletframe').within(() => {
 			cy.get('#Open_Local_Editor').click()
 		})
 
-		cy.get('.oc-dialog').should('be.visible')
-		cy.get('.oc-dialog .oc-dialog-title')
-			.should('contain', 'Open file locally ')
+		cy.get('.confirmation-dialog').should('be.visible')
+		cy.get('.confirmation-dialog h1')
+			.should('contain', 'Open file locally')
 
-		cy.on('url:changed', (newUrl) => {
-			expect(newUrl).to.contain('nc://')
-		})
-		cy.get('.oc-dialog button.primary').click()
-
+		cy.intercept({
+			method: 'POST',
+			url: '**/openlocaleditor',
+		}).as('getLocalToken')
+		cy.window()
+			.then((window) => {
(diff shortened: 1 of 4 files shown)


src/mixins/openLocal.js
```diff
@@ -66,11 +52,22 @@ export default {
 						return
 					}
 					this.openingLocally = true
-					this.sendPostMessage('Get_Views')
+					this.postMessage.registerPostMessageHandler(this.handleCloseSesssion)
```
max-nextcloud, line 55:
```suggestion
					this.postMessage.registerPostMessageHandler(this.handleCloseSession)
```

src/mixins/openLocal.js
```diff
@@ -66,11 +52,22 @@ export default {
...
+						DontSaveIfUnmodified: false,
+						Notify: false,
+					})
+					this.sendPostMessage('Close_Session')
 				},
 			)
 		},
 
+		handleCloseSesssion() {
+			this.postMessage.unregisterPostMessageHandler(this.handleCloseSesssion)
```
max-nextcloud, line 67:
```suggestion
		handleCloseSession() {
			this.postMessage.unregisterPostMessageHandler(this.handleCloseSession)
```

max-nextcloud (approved):
Found a typo that should have no effect as it's consistent. Other than that seems good.

cypress/e2e/integration.spec.js
```diff
@@ -101,20 +101,21 @@ describe('Nextcloud integration', function() {
 		cy.openFile('document.rtf')
 	})
 
-	it.skip('Open locally', function() {
+	it.only('Open locally', function() {
```
max-nextcloud, line 104:
```suggestion
	it('Open locally', function() {
```

Issue #3368, opened 2023-12-19
Redesign guest name input


juliushaertl:
As brought up by @jancborchardt 

- reference for asking guest name from Talk: https://github.com/nextcloud/spreed/pull/10467
- We need to show the input before loading the document as we need to pass the name along to Collabora
- We can take the opportunity to get rid of quite some legacy code there then and do it properly with new vue components

jancborchardt:
FYI @szaimen assigned you for the design part since you did the same for Talk, but since it’s already done for Talk I don’t think much more is needed, right @juliushaertl? Just so it is known who can answer follow-up questions.

juliushaertl:
Just to confirm, clarify on how I would imagine the adapted talk mockup to public file shares, however if a quick mockup can be provided that is probably the best:

- Opening a single public share link shows the public share link page in the background and opens the modal directly to enter the name, once confirmed the name the document will open
- On a shared folder we will show the modal once the user opens a file, once confirmed the name the document will open
- The modal would show:
  - "Enter your name to join the "My document.odt" as a guest."
  - Input to enter the name, prefilled with a placeholder "Anonymous guest"
  - Button to "Submit name and join"
  - Talk shows the call icon and name, should we show something similar as the header of the modal?

juliushaertl:
@elzody Since you are checking the other ticket about the guest name picker, this might be valuable input to consider when recreating the picker in vue

PR #3606, opened 2024-04-19
Fix/guest username

elzody:
* Resolves: #3557 
* Resolves: #3368 
* Target version: main

### Summary

This pull request uses `Office.vue` to ask for guest username, which solves the problem of a guest not being asked for a username when accessing a public share located within a shared folder (instead of a direct share link to a file).

The pull request is not ready for review yet, just creating it here for reference.

### Checklist


szaimen (approved):
Screenshot looks good from design perspective :)

src/helpers/getLoggedInUser.js
```diff
@@ -0,0 +1,30 @@
+import axios from '@nextcloud/axios'
+import { generateUrl } from '@nextcloud/router'
+
+/**
+ * Gets the current user's display name if logged in.
+ *
+ * @return null | string
+ */
+async function getLoggedInUser() {
+	try {
```
juliushaertl, line 10: Just some background on why this worked before and why it now doesn't:

The files_sharing app [sets the incognito mode](https://github.com/nextcloud/server/blob/e7a7b4a40184dc3da2c83e858c820625f660e48e/apps/files_sharing/lib/Controller/ShareController.php#L327) which will basically [hide the user id](https://github.com/nextcloud/server/blob/ec5133b739eabc76271789504b4dbb91a534f552/lib/private/User/Session.php#L231) in any code that tries to fetch it, so also in the template rendering logic that would normally add the `data-user` attributes to the html attribute.

With the previous logic of document.js being nested in another iframe this frame document didn't apply the incognito mode so within the iframe we knew that the user was still logged in. 

The incognito mode was introduced in https://github.com/owncloud/core/pull/6003 however I think has quite some design flaws with hiding the user id.

I think your approach works fine, we could do it slightly different if we want to save the extra request:
- If we don't need the user id in Office.vue (the token api request will still have it) we can just expose an initial state of the incognito mode and skip the guest name picker then
```diff
diff --git a/lib/Service/InitialStateService.php b/lib/Service/InitialStateService.php
index 6726a09d..ee653ac2 100644
--- a/lib/Service/InitialStateService.php
+++ b/lib/Service/InitialStateService.php
@@ -94,6 +94,7 @@ class InitialStateService {
        }

        private function provideOptions(): void {
+               $this->initialState->provideInitialState('incognitoMode', \OC_User::isIncognitoMode());
                $this->initialState->provideInitialState('theme', $this->config->getAppValue(Application::APPNAME, 'theme', 'nextcloud'));
                $this->initialState->provideInitialState('uiDefaults', [
                        'UIMode' => $this->config->getAppValue(Application::APPNAME, 'uiDefaults-UIMode', 'notebookbar')
```
- If we need the user information we can do something like this: https://github.com/nextcloud/richdocuments/blob/5bd5e09473ccb94a524d5b581845a1558183e575/lib/PermissionManager.php#L65-L73

juliushaertl:
Rebased on top of #3634

juliushaertl:
Seems I broke it when rebasing, but should be a quick fix now

juliushaertl (approved):

PR #3619, opened 2024-04-24
fix: Make edit mode in interactive widgets opt-in

juliushaertl:

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


src/view/Office.vue
```diff
@@ -83,11 +83,18 @@
 			:style="{visibility: showIframe ? 'visible' : 'hidden' }"
 			:src="iframeSrc" />
 
+		<NcButton v-if="isEmbedded && !hasToggledInteractiveEmbedding" class="toggle-interactive" @click="toggleEdit">
+			{{ t('text', 'Edit') }}
```
max-nextcloud, line 87:
```suggestion
			{{ t('richdocuments', 'Edit') }}
```

src/view/Office.vue
```diff
@@ -456,6 +476,10 @@ export default {
 			this.close()
 		},
 
+		toggleEdit() {
+			this.hasToggledInteractiveEmbedding = true
+		},
+
```
max-nextcloud, line 482: Minor naming nitpicks: 
* This does not really toggle. It only sets to true.
* the names do not match well - `Edit` vs. `InteractiveEmbedding`.

How about this?
```suggestion
		enableEditing() {
			this.hasEditingEnabled = true
		},

```

max-nextcloud (approved):
Code looks good. Added two minor comments.

src/view/Office.vue
juliushaertl: 🙈

src/view/Office.vue
juliushaertl: Good call, used hasWidgetEditingEnabled now to be more explicit about where its used.

juliushaertl:
Cypress fails due to the new 24.04 release of Collabora with changed selectors, will address separately

PR #3634, opened 2024-04-29
test(cypress): Adapt selectors to Collabora 24.04

juliushaertl:
Let's get main branch back to green with Collabora 24.04 that was recently released and the docker image used by CI was updated

- [x] Check SaveAs feature upstream as when testing locally the button seems to be missing
- [x] https://github.com/CollaboraOnline/online/pull/8926

### DIFF
--- a/cypress/e2e/integration.spec.js
+++ b/cypress/e2e/integration.spec.js
@@ -47,7 +47,7 @@ describe('Nextcloud integration', function() {
 	it('Sharing sidebar', function() {
 		cy.get('@loleafletframe').within(() => {
 			cy.get('#File-tab-label').click()
-			cy.get('#ShareAs').click()
+			cy.get('#ShareAs-button').click()
 		})
 
 		cy.get('#app-sidebar-vue')
@@ -62,7 +62,7 @@ describe('Nextcloud integration', function() {
 	it('Versions sidebar', function() {
 		cy.get('@loleafletframe').within(() => {
 			cy.get('#File-tab-label').click()
-			cy.get('#Rev-History').click()
+			cy.get('#Rev-History-button').click()
 		})
 
 		cy.get('#app-sidebar-vue')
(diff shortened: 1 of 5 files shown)


cypress/e2e/integration.spec.js
```diff
@@ -128,7 +129,7 @@ describe('Nextcloud integration', function() {
 	it('Insert image', function() {
 		cy.get('@loleafletframe').within(() => {
 			cy.get('#Insert-tab-label').click()
-			cy.get('#insert-insert-graphic').click()
+			cy.get('#insert-insert-graphic-button').click()
 			cy.get('#w2ui-overlay-insert-graphic-menu .menu-text').eq(1).click()
```
pedropintosilva, line 133: It's now failing because of this. Either replace it with `#insert-insert-graphic-entry-0` or `#insert-insert-graphic-entry-1`

image

pedropintosilva (changes requested):

cypress/e2e/open.spec.js
```diff
@@ -51,12 +51,13 @@ describe('Open existing office files', function() {
 			cy.waitForViewer()
 			cy.waitForCollabora()
 
-			cy.screenshot('open-file_' + filename)
+			cy.waitForPostMessage('App_LoadingStatus', { Status: 'Document_Loaded' })
 
 			// Share action
+			cy.wait(2000)
```
juliushaertl, line 57: It looks like there is some UI reset once the document renders, but I cannot figure out a proper way to assert that it is actually ready. Would go with the wait for now to move on.

Cypress run recording for reference:

https://github.com/nextcloud/richdocuments/assets/3404133/41dd2339-4863-4d5f-bb42-a01b38c2eeca

FYI @pedropintosilva Nothing critical i think, though wondering why the UI is reset there

elzody (approved):

PR #3635, opened 2024-04-30
fix: Block incompatible operations with remote tokens

juliushaertl:
With federated tokens we do not allow storage operations that would be
relative to the file path so we should block them

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


elzody (approved):
Seems reasonable~

Issue #3621, opened 2024-04-26
No Edit in Nextcloud Office option for PDF files


thebearon:
Similar issue to #1894. In Nextcloud 28.0.5 and richdocuments 8.3.5, there is no Edit in Nextcloud Office option for PDF files in the ... menu.

This used to work in Nextcloud 27 (upgraded from 27.1.9).

juliushaertl:
I can reproduce. It looks like we missed to migrate the optional file action to the new API with 28:

https://github.com/nextcloud/richdocuments/blob/main/src/viewer.js#L45-L67

Needs to be moved to https://nextcloud-libraries.github.io/nextcloud-files/functions/registerFileAction.html

PR #3645, opened 2024-05-01
File action to open PDFs with richdocuments when files_pdfviewer is enabled

elzody:
* Resolves: #3621 
* Target version: main

### Summary
Since 28, when the files_pdfviewer app is enabled, there is no open to open a PDF using richdocuments. This PR fixes that, using the files API to register a file action. By default, when you click on the file, it will still use the files_pdfviewer when enabled.

### Checklist


img/app.svg
juliushaertl, line 1: We should rather add a separate app-dark.svg file. Unfortunately the apps management UI currently requires the main app icon to be white.

juliushaertl (commented):
Small comment on the app icon, otherwise very nice and good to get in 👍

img/app.svg
elzody: Ah, that explains why there are already two of them then! I will revert it :)

juliushaertl (approved):

juliushaertl:
@elzody Would be quite nice if we could follow up with a small cypress test for this to avoid further regressions

PR #3664, opened 2024-05-06
Add SPDX header

AndyScherzinger:
* Resolves: none
* Target version: main

### TODO

- [x] finish-up adding missing header
- [x] run spdx converter
- [x] add reuse GH action

### Checklist


AndyScherzinger:
Auto loader check fails: would need help to understand why and how to fix.
https://github.com/nextcloud/richdocuments/actions/runs/8968932531/job/24629343575?pr=3664

composer/autoload.php
```diff
@@ -1,6 +1,8 @@
 <?php
-
-// autoload.php @generated by Composer
+/**
```
juliushaertl, line 2: Can we rather specify this through the .reuse/dep5 file? Those files are autogenerated by composer to have a static autoloader and the CI checks if those match the generated ones.

AndyScherzinger: Yes, sure, especially when auto generated 👍

AndyScherzinger: @juliushaertl DONE 👍

juliushaertl:
Haven't seen this cypress failure before and main branch is passing, need to look into why that happens

juliushaertl (approved):

Issue #3605, opened 2024-04-18
Federated editing is broken after vue migration


juliushaertl:
This is no longer possible since https://github.com/nextcloud/richdocuments/pull/3152

- [ ] Come up with a fix
- [ ] Document federated editing process and share knowledge with the team
- [ ] Write a cypress test

PR #3669, opened 2024-05-07
fix: Bring back federated editing in viewer iframe

juliushaertl:
Quick attempt to fix https://github.com/nextcloud/richdocuments/issues/3605

Ideally we'd also remove the document.js usage there, but this at least brings back the functionality.

- [x] Try to come up with a cypress test

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -282,7 +282,9 @@ class DocumentController extends Controller {
 					'userId' => $remoteWopi->getEditorUid() ? ($remoteWopi->getEditorUid() . '@' . $remoteServer) : null,
 				];
 
-				return $this->documentTemplateResponse($wopi, $params);
+				$response = $this->documentTemplateResponse($wopi, $params);
+				$response->addHeader('X-Frame-Options', 'ALLOW');
+				return $response;
 			}
 		} catch (ShareNotFound $e) {
 			return new TemplateResponse('core', '404', [], 'guest');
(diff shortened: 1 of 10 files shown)


lib/Controller/DocumentController.php
```diff
@@ -294,6 +296,16 @@ public function remote(string $shareToken, string $remoteServer, string $remoteS
 		return new TemplateResponse('core', '403', [], 'guest');
 	}
 
+	/**
+	 * Open file on Source instance with token from Initiator instance
+	 */
+	#[PublicPage]
+	#[NoCSRFRequired]
+	public function remotePost(string $shareToken, string $remoteServer, string $remoteServerToken, ?string $filePath = null): TemplateResponse {
+		return $this->remote($shareToken, $remoteServer, $remoteServerToken, $filePath);
```
juliushaertl, line 305: This is needed as currently the routing cannot handle two routes with the same controller method name. We need one for POST and one for GET

src/helpers/coolParameters.js
```diff
@@ -14,7 +14,7 @@ const getUIDefaults = () => {
 	const uiMode = defaults.UIMode ?? 'notebookbar'
 
 	const systemDarkMode = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
-	const dataset = (document.body.dataset.themes ? document.body.dataset : parent?.document.body.dataset) ?? {}
+	const dataset = {} // (document.body.dataset.themes ? document.body.dataset : parent?.document.body.dataset) ?? {}
```
juliushaertl, line 17: Needs to be checked still

mejo- (approved):
Didn't test and don't know the code well, but at least didn't spot anything obvious.

PR #3676, opened 2024-05-08
fix: Do not set user id for guests

juliushaertl:
Should fix #3675 

Aparently I've not catched that during testing as it only happens on file save. We should extend our integration tests for the WOPI endpoints in that regard.

@elzody Do you recall what the line tried to achieve in your recent change?

Edit: This test could have catched this however the /token endpoint is not used in the behat tests. Will see if we can extend that

PR #3681, opened 2024-05-08
Test guest name setting

juliushaertl:
Adds some integration tests to check that the guest name is actually properly set, returned to the WOPI endpoint and saving a file works for public shares.

In additions it fixes my wrong fix in #3676 as just dropping the line also dropped using the guest name parameter.


elzody (approved):
This already seems like a good improvement from my perspective

PR #3726, opened 2024-06-03
Always pass is_admin property to userextrainfo

eszkadev:
This will be used by COOL for server audit dialog
- which should be shown only for admin users. In case of missing is_admin property - it will be shown too because it means integration don't use that property.

see https://github.com/CollaboraOnline/online/pull/9204


juliushaertl (approved):

eszkadev:
Hi, can we backport later to stable branches too?

PR #3749, opened 2024-06-11
fix: Avoid requesting remote endpoints during bootstrap

juliushaertl:
This will add some abstraction layer for cached request data as used by
capabilities and discovery endpoints. By default it will always return
the cached data which will never expire (backed by a file in app data in
case the memory cache vanishes).

In short the logic is:
- Fetch the endpoints when configuring Collabora
- Refetch them every hour with the background job
- When trying to get the data we will (in order)
  - Check the memory cache
  - Check the file in appdata (and set the memory cache again)

There is still a use of fetch for richdocumentscode but that needs some more refactoring to work differently. Does not affect instances where richdocumentscode is not enabled.

Also decreasing the timeout to 5 seconds which should be more then enough.

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -120,12 +120,18 @@ class Application extends App implements IBootstrap {
 			$appConfig->setAppValue('wopi_url', $new_wopi_url);
 			$appConfig->setAppValue('disable_certificate_verification', 'yes');
 
+			/** @var DiscoveryService $discoveryService */
 			$discoveryService = $this->getContainer()->get(DiscoveryService::class);
+			/** @var CapabilitiesService $capabilitiesService */
 			$capabilitiesService = $this->getContainer()->get(CapabilitiesService::class);
 
 			$discoveryService->resetCache();
 			$capabilitiesService->resetCache();
-			$capabilitiesService->fetchFromRemote();
+			try {
+				$capabilitiesService->fetch();
+				$discoveryService->fetch();
+			} catch (\Exception $e) {
+			}
 		}
 	}
 }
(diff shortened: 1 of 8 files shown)


lib/AppInfo/Application.php
```diff
@@ -120,12 +120,18 @@ public function checkAndEnableCODEServer() {
...
+			/** @var CapabilitiesService $capabilitiesService */
 			$capabilitiesService = $this->getContainer()->get(CapabilitiesService::class);
 
 			$discoveryService->resetCache();
 			$capabilitiesService->resetCache();
-			$capabilitiesService->fetchFromRemote();
+			try {
+				$capabilitiesService->fetch();
+				$discoveryService->fetch();
+			} catch (\Exception $e) {
```
elzody, line 133: Do we need to do anything here with the caught error? Is it OK to catch it and just throw it away?

lib/Service/CapabilitiesService.php
```diff
@@ -8,52 +8,42 @@
...
 		if (!$this->capabilities) {
-			$this->capabilities = $this->cache->get('capabilities');
-		}
-
-		$isARM64 = php_uname('m') === 'aarch64';
-		$CODEAppID = $isARM64 ? 'richdocumentscode_arm64' : 'richdocumentscode';
-		$isCODEInstalled = $this->appManager->isEnabledForUser($CODEAppID);
-		$isCODEEnabled = strpos($this->config->getAppValue('richdocuments', 'wopi_url'), 'proxy.php?req=') !== false;
-		$shouldRecheckCODECapabilities = $isCODEInstalled && $isCODEEnabled && ($this->capabilities === null || count($this->capabilities) === 0);
-		if ($this->capabilities === null || $shouldRecheckCODECapabilities) {
```
juliushaertl, line 55: This is what has actually caused the issue, as the cache was only there for an hour every request after it expired triggered a new request to the capabilities endpoint, which was unreachable

blizzz (approved):
Quick review, did not test

juliushaertl:
Cypress failure also on main, likely some upstream change

PR #3754, opened 2024-06-12
Use new property for guest indication

eszkadev:


eszkadev:
followup for https://github.com/nextcloud/richdocuments/pull/3748

Issue #3506, opened 2024-02-27
[Bug]: Unable to uncheck 'restrict usage/edit to specific groups' in Nextcloud Admin Settings


isdnfan:
**Describe the bug**
- once checked both values can not be unchecked.
error is logged in browser console

**To Reproduce**
1. Go to 'https://cloud.mydomain/settings/admin/richdocuments)'
2. Check the checkbox 'restrict usage to specific groups'
3. try to unCheck the checkbox 'restrict usage to specific groups'
4. in browser console see error 'TypeError: this.value.filter is not a function'

**Expected behavior**
settings could be checked/unchecked

**Screenshots**
![image](https://github.com/nextcloud/richdocuments/assets/18125597/7f9d5219-b7e1-4dcf-8b5c-ce31ec351d4c)

**Client details:**
 - OS: Windows
 - Browser Firefox

## Server details

**Operating system**: docker community edition
**Web server:** Apache
**Database:** MariaDB
**PHP version:** 8.2
**Nextcloud version:** 28.0.2
**Version of the richdocuments app** 8.3.1
**Version of Collabora Online** 23.05.6.4
**Configuration of the richdocuments app**
```
    "richdocuments": {
      "canonical_webroot": "",
      "doc_format": "",
      "edit_groups": "",
      "enabled": "yes",
      "external_apps": "",
      "installed_version": "8.3.1",
      "public_wopi_url": "https://collabora.mydomain.tld",
      "types": "prevent_group_restriction",
      "use_groups": "",
      "wopi_allowlist": "172.16.0.0/12,fd00:feed:beef::/48",
      "wopi_url": "https://collabora.mydomain.tld"
    }
```

initially reported at https://help.nextcloud.com/t/nextcloud-collabora-code-docker-compose-almost-everything-is-working-except-it-isnt/183450

isdnfan:
similar to #3377

lftsy:
FYI, here is the woraround I used

```
root@ifrit:[~]# sudo -u www-data php --define apc.enable_cli=1 /var/www/nextcloud/occ config:app:delete richdocuments use_groups
Config value use_groups of app richdocuments deleted

root@ifrit:[~]# sudo -u www-data php --define apc.enable_cli=1 /var/www/nextcloud/occ config:app:delete richdocuments edit_groups
Config value edit_groups of app richdocuments deleted
```

RealKoenisch:
[...]

PR #3758, opened 2024-06-12
fix: Uncheck group limit in admin settings

elzody:
* Resolves: #3506 
* Target version: main

### Summary
Previously, trying to uncheck the checkbox would simply re-enable it automatically. This PR fixes this issue. Also adds an `inputLabel` prop to the `NcSelect` so it doesn't error anymore (and accessibility reasons :D)

https://github.com/nextcloud/richdocuments/assets/23369449/3940d131-83a1-49e3-a8b3-af656070c70a

### Checklist

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -225,7 +225,7 @@
 				:disabled="updating"
 				@input="updateOoxml" />
 
-			<SettingsCheckbox :value="settings.use_groups !== null"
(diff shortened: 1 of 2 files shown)


juliushaertl:
Thanks for looking into this. We used to hide the select if the box was unchecked. Maybe we can bring back this behaviour as well?

jancborchardt (commented):
Looks good, have only the same feedback as Julius about hiding the dropdown. :)

juliushaertl (approved):

PR #3763, opened 2024-06-13
richdocuments: add host_session_id parameter

hcvcastro:
* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -21,6 +21,7 @@ import '../css/document.scss'
 import axios from '@nextcloud/axios'
 import { spawnDialog } from '@nextcloud/dialogs'
 import SaveAs from './components/Modal/SaveAs.vue'
+import { getCapabilities } from './services/capabilities.ts'
 
 const PostMessages = new PostMessageService({
 	parent: window.parent,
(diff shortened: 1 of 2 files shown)


juliushaertl:
Can you provide some context on what this is used?

In addition you can already get the richdocuments version in the frontend code through the capabilities so no backend changes should be required and please also add it to https://github.com/nextcloud/richdocuments/blob/e9ec3189c92c229475f6430b47e02d6463699c65/src/view/Office.vue#L51
```
import { getCapabilities } from '../services/capabilities.ts'

getCapabilities().version
```

juliushaertl (changes requested):
Needs mentioned adjustments

hcvcastro:
Thanks, it is mostly to implement Wopi HOST_SESSION_ID, to debugging log purposes and
to track wopi host  logs for troubleshooting purposes based on a host-specific session identifier

src/document.js
```diff
@@ -21,6 +21,7 @@ import '../css/document.scss'
 import axios from '@nextcloud/axios'
 import { spawnDialog } from '@nextcloud/dialogs'
 import SaveAs from './components/Modal/SaveAs.vue'
+import { getCapabilities } from '../services/capabilities.ts'
```
juliushaertl, line 24:
```suggestion
import { getCapabilities } from './services/capabilities.ts'
```

juliushaertl (changes requested):
Still needs to also be added in 
https://github.com/nextcloud/richdocuments/blob/e9ec3189c92c229475f6430b47e02d6463699c65/src/view/Office.vue#L51

PR #3797, opened 2024-06-24
Fix event name files:node:rename

artonge:
From the files action, it does not have an ending `d`: https://vscode.dev/github/nextcloud/server/blob/artonge/fix/update_favorite_navigation/apps/files/src/actions/renameAction.spec.ts#L90
Or maybe it should and the file action is wrong?

Other reference: https://github.com/nextcloud/text/blob/master/src/views/RichWorkspace.vue#L105


artonge:
Response from @skjnldsv, this is two different events. The one without the `d` is to trigger the rename.

Issue #3711, opened 2024-05-24
Template-based PDF generation via API 


blizzz:
**Is your feature request related to a problem? Please describe.**

Through an API endpoint, it shall be possible to create a PDF document based on a template with placeholders. 

**Describe the solution you'd like**

An OCS endpoint accepts
- a folder id to store the resulting document
- a template identifier
- a json parameter with key-value-pairs for the placeholders

The app will on-the-fly create a new document, do the replacements, and store it as a PDF document.

**Additional context**

This is required for the upcoming workflow engine efforts.

juliushaertl:
Would be good to evaluate together with Collabora if this could just be done by extending the convert to endpoint they have, as this would also allow doing conversions to various file types right away https://www.collaboraoffice.com/document-conversion/

juliushaertl:
> a template identifier

Has there been some discussions on how templates are defined? I assume they'd just be another Nextcloud file so a file ID is probably best for this one?

juliushaertl:
Could also be the same backend mechanism as for https://github.com/nextcloud-gmbh/office-team/issues/51

blizzz:
> > a template identifier
> 
> Has there been some discussions on how templates are defined? I assume they'd just be another Nextcloud file so a file ID is probably best for this one?

Not specified more closely, my assumption was to use the existing mechanism. If that works with file id, file id it is.

juliushaertl:
Basic implementation idea:
- [ ] Some service class in richdocuments with two methods (`extractFields(Node $file): ?array`/`fillFields(Node $file, array $fieldValues)`)
- [ ] OCS Controller to expose those as an API with two possible modes
  - [ ] Send file data directly
  - [ ] Send a file id for an existing file in Nextcloud
  - [ ] Have a parameter target to either write the target file to or return the file if left out

The service can then be reused for https://github.com/nextcloud-gmbh/office-team/issues/51

juliushaertl:
For reference the early draft from Collabora:


If you want to experiment early, we expect the API would look like this:

## 1) Extraction

- new /cool/extract-doc-structure?limit=content-control endpoint

 - we'll add more document structure here, emitting everything by
   default, but you want to limit to content controls probably

- takes a document

- outputs a JSON like this:

```
{
   "DocStructure": {
       "ContentControls.ByIndex.0": {
           "id": -1318486635,
           "tag": "machine-readable",
           "alias": "Human Readable",
           "content": "plain text value",
[...]

PR #3798, opened 2024-06-24
feat: PDF Template Workflow

elzody:
* Resolves: #3711 
* Target version: main

### Summary
API which handles the extraction and filling of form controls in template documents. This means that a user will be able to input form controls in the document via Collabora Online, and when creating a new document based on that template, have the option to fill out the fields and they will be populated when the document is created. This is required for the workflow and automation improvements.

Related server pull request: [nextcloud/server#46383](https://github.com/nextcloud/server/pull/46383)

### TODO

- [x] Also accept any document, not just templates -- user may want to get/fill fields from other documents as well
- [ ] Testing, would need help with this, was not working locally when I tried :)

### Checklist


juliushaertl:
Just noticed, the field classes seem to be missing from the commits ;)

lib/Service/TemplateFieldService.php
```diff
@@ -0,0 +1,129 @@
...
+			'headers' => ['Content-Type' => 'multipart/form-data'],
+		];
+
+		try {
+			$response = $httpClient->post(
+				$collaboraUrl . "/cool/extract-document-structure",
+				[
+					'query' => ['limit' => 'content-control'],
+					'multipart' => [$formData],
+				]
```
juliushaertl, line 56: Requests should also pass the default options, @mejo- just yesterday wrapped that nicely in https://github.com/nextcloud/richdocuments/pull/3831/files#diff-c2d693a39b27f535396a48fdaad4a41a517c489be3a492fb968ac91e5e0181efR14-R21

lib/Service/TemplateFieldService.php
```diff
@@ -0,0 +1,129 @@
...
+			$response = $httpClient->post(
+				$collaboraUrl . '/cool/transform-document-structure',
+				[
+					'multipart' => [$formData, $formFormat, $formContents]
+				]
+			);
+
+			return $response->getBody();
+		} catch (\Exception $e) {
+			return $e->getMessage();
```
juliushaertl, line 126: Probably good to log here and throw a generic one that just returns with a proper HTTP status code in the controller.

lib/Template/CollaboraTemplateProvider.php
```diff
@@ -25,7 +25,11 @@ class CollaboraTemplateProvider implements ICustomTemplateProvider {
 	/** @var ITemplateManager */
 	private $coreTemplateManager;
 
-	public function __construct(TemplateManager $templateManager, IURLGenerator $urlGenerator, ITemplateManager $coreTemplateManager) {
+	public function __construct(
+		TemplateManager $templateManager,
+		IURLGenerator $urlGenerator,
+		ITemplateManager $coreTemplateManager
+	) {
```
juliushaertl, line 32: We can shorten the boiler plate code with https://stitcher.io/blog/constructor-promotion-in-php-8

Then you can also drop the separate propery definition above

```suggestion
	public function __construct(
		private TemplateManager $templateManager,
		private IURLGenerator $urlGenerator,
		private ITemplateManager $coreTemplateManager
	) {
	}
```

lib/Service/TemplateFieldService.php
juliushaertl: Ah otherwise the error message could also get written to the file 🙈

lib/Service/TemplateFieldService.php
```diff
@@ -0,0 +1,129 @@
...
+
+	public function __construct(
+		IClientService $clientService,
+		AppConfig $appConfig,
+		IRootFolder $rootFolder
+	) {
+		$this->clientService = $clientService;
+		$this->appConfig = $appConfig;
+		$this->rootFolder = $rootFolder;
+	}
```
juliushaertl, line 30:
```suggestion
	public function __construct(
		private IClientService $clientService,
		private AppConfig $appConfig,
		private IRootFolder $rootFolder
	) {
	}
```

lib/Listener/BeforeGetTemplatesListener.php
```diff
@@ -0,0 +1,35 @@
...
+use OCP\EventDispatcher\IEventListener;
+use OCP\Files\Template\BeforeGetTemplatesEvent;
+
+/** @template-implements IEventListener<BeforeGetTemplatesEvent|Event> */
+class BeforeGetTemplatesListener implements IEventListener {
+	private TemplateFieldService $templateFieldService;
+
+	public function __construct(TemplateFieldService $templateFieldService) {
+		$this->templateFieldService = $templateFieldService;
+	}
```
juliushaertl, line 21:
```suggestion
	public function __construct(
		private TemplateFieldService $templateFieldService,
	) {
	}
```

lib/Controller/TemplateFieldController.php
```diff
@@ -0,0 +1,72 @@
...
+	 * @return DataResponse
+	 */
+	#[NoAdminRequired]
+	public function extractFields(int $fileId): DataResponse {
+		try {
+			$fields = $this->templateFieldService->extractFields($fileId);
+
+			return new DataResponse($fields, Http::STATUS_OK);
+		} catch (\Exception $e) {
+			return new DataResponse([$e->getMessage()], Http::STATUS_INTERNAL_SERVER_ERROR);
```
juliushaertl, line 53: Sorry should have been more clear, ideally just return a more generic error message instead of sending the exception message to the response as it might contain and leak internal server data.

juliushaertl (approved):
Just one small thing left, otherwise 👍

PR #3806, opened 2024-07-03
feat: implement PdfService using mikehaertl/php-pdftk

luka-nextcloud:
* Resolves: 
* Target version: main

### Follow-up
- [ ] Setup check to suggest installing pdftk
- [ ] Template creator for PDF files https://github.com/nextcloud/richdocuments/issues/3867
- [ ] Tests


lib/Service/PdfService.php
```diff
@@ -0,0 +1,60 @@
...
+class PdfService {
+	public function __construct(
+		private IConfig $config,
+		private LoggerInterface $logger
+	) {
+	}
+
+	public function extractFields(Node $file): array {
+		$dataDir = $this->config->getSystemValue('datadirectory');
+		$filePath = $dataDir . $file->getPath();
```
juliushaertl, line 24: This will not work with object storage or other external storage locations. Rather use something like this:


```suggestion
		$filePath = $file->getStorage()->getLocalFile($file->getInternalPath());
```

lib/Service/PdfService.php
```diff
@@ -0,0 +1,60 @@
...
+				$templateFields[] = [
+					'name' => $field['FieldName'],
+					'type' => $field['FieldType'],
+					'default' => $field['FieldValueDefault'] ?? '',
+					'options' => $field['FieldStateOption'] ?? [],
+				];
+			}
+			return $templateFields;
+		} catch (\Exception $e) {
+			// Log
```
juliushaertl, line 40:
```suggestion
```

lib/Service/PdfService.php
```diff
@@ -0,0 +1,58 @@
...
+			$pdf = new Pdf($filePath);
+			$fields = $pdf->getDataFields();
+			$templateFields = [];
+			foreach ($fields as $field) {
+				$templateFields[] = [
+					'name' => $field['FieldName'],
+					'type' => $field['FieldType'],
+					'default' => $field['FieldValueDefault'] ?? '',
+					'options' => $field['FieldStateOption'] ?? [],
+				];
```
juliushaertl, line 35: @luka-nextcloud This needs a few adjustments after rebasing to the latest state of the PR from @elzody 

There is now a dedicated Field class that can be used for this.

lib/Service/PdfService.php
```diff
@@ -0,0 +1,58 @@
...
+		$dataDir = $this->config->getSystemValue('datadirectory');
+		$filePath = $dataDir . $file->getPath();
+
+		try {
+			$pdf = new Pdf($filePath);
+			$pdf->fillForm($fieldValues);
+			$pdf->flatten();
+			$pdf->saveAs($filePath);
+		} catch (\Exception $e) {
+			$this->logger->debug('Failed to fill fields in PDF: {error}', ['error' => $e->getMessage()]);
```
juliushaertl, line 54:
```suggestion
			$this->logger->error('Failed to fill fields in PDF: {error}', ['error' => $e->getMessage()]);
```

juliushaertl: Seems important enough to log as error, also passing as exception helps to preserve details.

lib/Service/PdfService.php
```diff
@@ -0,0 +1,58 @@
...
+			return $templateFields;
+		} catch (\Exception $e) {
+			$this->logger->debug('Failed to extract fields from PDF: {error}', ['error' => $e->getMessage()]);
+			return [];
+		}
+	}
+
+	public function fillFields(Node $file, array $fieldValues): void {
+		$dataDir = $this->config->getSystemValue('datadirectory');
+		$filePath = $dataDir . $file->getPath();
```
juliushaertl, line 46: Also use the getLocalFile method here

juliushaertl (commented):
Some comments inline :)

lib/Service/PdfService.php
luka-nextcloud: @juliushaertl Which class do you mean?

juliushaertl: There is a Field class in server now that can be used for this. Checkout the implementation around https://github.com/nextcloud/richdocuments/blob/6f85ff4ff176870a0f086267b267c21a97544c74/lib/Service/TemplateFieldService.php#L63-L70 for office file types

juliushaertl:
@luka-nextcloud I pushed a few fixes while reviewing testing. Mainly to make the matching of fields work

@elzody I extended the OCS API also to be able to pass a destination along. However I noticed we have not really typed the fields array structure in the server API and here. Currently it seems to just use `array<int,array{content: string}>`. We should properly document that in the server API and all methods around here. What I currently lack to know is if the array key is actually using the Field class index value or is just a numeric index based on what the vue frontend in files loops over.

Sample curl commands
```

 curl -H 'OCS-APIRequest: true' https://admin:admin@nextcloud.local/ocs/v2.php/apps/richdocuments/api/v1/template/fields/extract/37526\?format\=json
 
 curl -X POST -H 'OCS-APIRequest: true' https://admin:admin@nextcloud.local/ocs/v2.php/apps/richdocuments/api/v1/template/fields/fill/37526\?format\=json -H "Content-Type: application/json" --data '{"fields":[{"content":"FooBar"}],"destination":"foobar.pdf"}'

```

elzody:
@juliushaertl The index should be the string index from the field class, afaik. The second curl command looks like it should have something like  `'{"fields":["someIndex": {"content":"FooBar"}],"destination":"foobar.pdf"}'`, else I'm not sure how it would know which form control the content being passed belongs to

juliushaertl:
All right, the extra object with a content property originates from the Collabora API?

elzody:
@juliushaertl I think so, if I remember correctly you pass the fields to the endpoint in richdocuments as an array, and it loops over the fields passed and extracts the form control index and its matching properties (content, alias, etc.) then passes an object looking something like this to the Collabora API in JSON format:

```json
{
  "ContentControls.ByIndex.0": {
    "content": "Something"
  },
  "ContentControls.ByIndex.1": {
    "alias": "Something else"
  }
}

And yeah because it allows changing multiple controls at a time the index is needed to specify what data each control should have via its index. Hopefully that makes sense, not sure if that's what you're asking
```

juliushaertl (approved):

PR #3832, opened 2024-07-25
fix: Do not ask for guest name on read only pdfs

juliushaertl:
Avoid asking for the guest name if a read only PDF is opened as there is nothing guests can do then


DanScharon:
is it possible to backport this fix also to 28 and 29?

juliushaertl:
Yes, we should indeed, thanks for bringing this up

PR #3868, opened 2024-08-01
docs: Add API docs for form extraction and filling API

juliushaertl:
Document what was there and added with #3806

### DIFF
--- a/docs/form_filling_api.md
+++ b/docs/form_filling_api.md
@@ -1,0 +1,96 @@
+<!--
+  - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+# Form extraction/filling API
+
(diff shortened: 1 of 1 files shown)


elzody (approved):
Yep looks good, sorry forgot to do it myself when introducing the changes

docs/form_filling_api.md
```diff
@@ -0,0 +1,96 @@
...
+```
+POST /ocs/v2.php/apps/richdocuments/api/v1/template/fields/fill/{fileId}
+```
+
+### Parameters
+
+| Parameter   | Format                                 | Description                                                               |
+|-------------|----------------------------------------|---------------------------------------------------------------------------|
+| fileId      | numeric                                | file id of the source file to fill out                                    |
+| fields      | array<string, array{content: string}>  | Fields to be filled out with string content, identified by a string index |
```
luka-nextcloud, line 86: I think the format should be `object<string, {content: string}>`

juliushaertl: Yes, for the json definition that makes more sense

PR #3890, opened 2024-08-07
ci: Properly split with multi-collabora matrix

juliushaertl:
It seems we distributed tests across all matrix cases so some were executed against nightly some against release containers of collabora 🙈


juliushaertl:
Last commit to be reverted once https://github.com/CollaboraOnline/online/issues/9727 is fixed and released

luka-nextcloud (approved):

Issue #3865, opened 2024-08-01
Guest name is not picked up the second time loading a share link


juliushaertl:
As reported in the Collabora call:

## Steps to reproduce

1. Open a public share link
2. Enter a name
3. Reload the page

## Actual behavior

No guest name picker, Showing up as anonymous guest, the cookie for guestName seems to be set

## Expected behavior 

No guest name picker, but the existing cookie is passed along properly to Collabora

juliushaertl:
We could just pass the guest name along as parameter and get rid of using a cookie for that these days.

Then we can also just use https://github.com/nextcloud-libraries/nextcloud-auth/pull/666 to store and obtain it

PR #3907, opened 2024-08-14
Fix/guest username

elzody:
* Resolves: #3865 
* Target version: main

### Summary
This PR updates the `@nextcloud/auth` version in use by richdocuments in order to take advantage of the `getGuestnickname` and `setGuestNickname` methods introduced, which use local browser storage rather than a cookie to set the guest nickname.

I think the issue previously was that the cookie with the guest name was set, but it wasn't read again and used, thus only being used on the first time the user is asked. But now it's fixed, and does not use a cookie anymore.

As far as I can tell, it does not require any back-end change.

### Checklist

- [ ] Tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "8.5.0-beta.1",
       "license": "AGPL-3.0-or-later",
       "dependencies": {
-        "@nextcloud/auth": "^2.3.0",
+        "@nextcloud/auth": "^2.4.0",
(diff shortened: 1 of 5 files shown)


juliushaertl (approved):
Looks good, yes backend wise passing the guestName over with the request to generate the token is enough as it gets set by that then

https://github.com/nextcloud/richdocuments/blob/2b02997fd1a6639d870b615cc5c211842f47b579/lib/Controller/DocumentController.php#L380

Issue #3892, opened 2024-08-07
Cache extracted template fields


juliushaertl:
When extracting fields from a template with Collabora we should cache the results so that not every attempt to create a new file needs to talk to the remote Collabora instance.

We can just wrap the remote API call with a cache that uses the file id and etag as cache key so it invalidates automatically whenever the file changes.

Some code sample https://github.com/nextcloud/richdocuments/blob/00f61ba7a5bf1232936ec7655292190fd6705184/lib/Service/CachedRequestService.php#L39-L52

PR #3885, opened 2024-08-06
feat(tests): Implement testing for new template workflow (filling templates)

elzody:
* Resolves: #3886 
* Target version: main

### Summary

This PR fixes the existing cypress tests in `templates.spec.js`, and introduces new tests to cover the additions to the template workflow; namely filling templates.

### Checklist

- [x] Code is properly formatted
- [x] Sign-off message is added to all commits
- [x] Documentation (manuals or wiki) has been updated or is not required

PR #3916, opened 2024-08-15
Cache `extractFields` response from Collabora

elzody:
* Resolves: #3892
* Target version: main

### Summary
I rebased this PR onto #3885 because it includes a fix for an error due to using `getFirstNodeById()` incorrectly.
Thus, should probably wait for #3885 to be merged before merging this one

The response from extracting fields via the Collabora endpoint should be cached as long as the file has not changed, since it prevents an extra hit to Collabora. When the file changes then, it should re-hit the Collabora endpoint for the updated template fields, then cache from there.

### Checklist

### DIFF
--- a/lib/Service/TemplateFieldService.php
+++ b/lib/Service/TemplateFieldService.php
@@ -14,15 +14,20 @@ use OCP\Files\NotFoundException;
 use OCP\Files\Template\Field;
 use OCP\Files\Template\FieldType;
 use OCP\Http\Client\IClientService;
+use OCP\ICache;
+use OCP\ICacheFactory;
 use Psr\Log\LoggerInterface;
 
 class TemplateFieldService {
+	private ICache $cache;
+
 	public function __construct(
 		private IClientService $clientService,
 		private CapabilitiesService $capabilitiesService,
 		private AppConfig $appConfig,
 		private IRootFolder $rootFolder,
-		private LoggerInterface $logger
+		private LoggerInterface $logger,
+		private ICacheFactory $cacheFactory
 	) {
 	}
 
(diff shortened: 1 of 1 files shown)


lib/Service/TemplateFieldService.php
```diff
@@ -44,6 +50,13 @@ public function extractFields(Node|int $file) {
 				throw new NotFoundException();
 			}
 
+			$cacheName = $file->getId() . "/" . $file->getEtag();
+			$cachedResponse = $this->cache->get($cacheName);
+
+			if ($cachedResponse) {
```
juliushaertl, line 56:
```suggestion
			if ($cachedResponse !== null) {
```

Otherwise an empty array would also evaluate to false and not use the cached result.

lib/Service/TemplateFieldService.php
```diff
@@ -14,16 +14,22 @@
...
 	public function __construct(
 		private IClientService $clientService,
 		private CapabilitiesService $capabilitiesService,
 		private AppConfig $appConfig,
 		private IRootFolder $rootFolder,
-		private LoggerInterface $logger
+		private LoggerInterface $logger,
+		private ICacheFactory $cacheFactory
 	) {
+		$this->cache = $cacheFactory->createLocal('richdocuments_templates/');
```
juliushaertl, line 32: For this case it does not matter as the createLocal method is rather lightweight, but generally it is a good pattern to avoid calling any code in constructors for classes that use dependency injection. That way this code will always be called as soon as TemplateFieldService initialized which can potentially cause performance issues that are rather hard to debug.

juliushaertl (approved):
Just one small issue with checking the cache, otherwise 👍

lib/Service/TemplateFieldService.php
elzody: Thanks for the note, is there a better pattern I should follow then in the future?

PR #3925, opened 2024-08-19
fix(branding): properly send custom logo to collabora

elzody:
* Resolves: #3913 
* Target version: main

### Summary
This PR properly formats the string which hands over the css values to the collabora frame. It was previously not accepting a `;` at the end, invalidating the string and collabora couldn't read the `--nc-custom-logo` value properly. It seems to work better with string interpolation.

### Checklist


juliushaertl (approved):
🙈

PR #3948, opened 2024-08-26
fix: Do not set a guest name on actual users

juliushaertl:
Should fix using the wrong user for filesystem access as described in https://github.com/nextcloud/server/issues/40090

Before this the node was obtained from the share owner which caused issues getting the relative path later on

Debugger screenshot for my own reference:

image

### Checklist

### DIFF
--- a/lib/TokenManager.php
+++ b/lib/TokenManager.php
@@ -185,7 +185,7 @@ class TokenManager {
 		);
 
 		$serverHost = $this->urlGenerator->getAbsoluteURL('/');
-		$guestName = $editoruid === null ? $this->prepareGuestName($this->helper->getGuestNameFromCookie()) : $editoruid;
+		$guestName = $editoruid === null ? $this->prepareGuestName($this->helper->getGuestNameFromCookie()) : null;
 		return $this->wopiMapper->generateFileToken($fileId, $owneruid, $editoruid, $version, $updatable, $serverHost, $guestName, $hideDownload, $direct, 0, $shareToken);
 	}
 


SystemKeeper (approved):
Works fine for me and solves https://github.com/nextcloud/server/issues/40090 👍 
Thanks!

juliushaertl:
@elzody Do you remember what the intention of the change in https://github.com/nextcloud/richdocuments/commit/63aece8ff3119d8aec9f1a998dd07ac4e5e18950#diff-7a3746877ab19393cc83236c399a1a05d24b6dd96ddc1bfa2985b74108a04ca2R202 was?

elzody (approved):
@juliushaertl I can't recall, I think it was confusing logic for me and I accidentally had the wrong approach there. It makes sense now with this, and if it indeed fixes the issue, looks good from my side

ariep:
Hi @juliushaertl , the above messages seem to indicate that this fix is not included in recent releases of `richdocuments` for nextcloud 27, 28 and 29, is that correct? Also this PR is not mentioned at all on the [releases](https://github.com/nextcloud/richdocuments/releases) page.

If so, is there any chance you could perform that manual backport, perhaps only for 29?

elzody:
@ariep I'm not Julius, but indeed it seems I missed these when doing the releases -- I think maybe you're right that the backports were missed somehow -- will look into it

ariep:
> @ariep I'm not Julius, but indeed it seems I missed these when doing the releases -- I think maybe you're right that the backports were missed somehow -- will look into it

Thanks @elzody for looking into it!

juliusknorr:
30 and below are EOL so we are not doing those backports anymore

PR #3951, opened 2024-08-27
feat: Add parameter to allow converting files after filling out fields

juliushaertl:
Need to review my own code again tomorrow but so far this should do the trick to help the integrations team with building nice workflows

cc @julien-nc 

I should also add a test

```
curl -v -X POST -H 'OCS-APIRequest: 1' -u admin:admin https://nextcloud.local/ocs/v2.php/apps/richdocuments/api/v1/template/fields/fill/406\?convert\=pdf  > test2.pdf
curl -v -X POST -H 'OCS-APIRequest: 1' -u admin:admin https://nextcloud.local/ocs/v2.php/apps/richdocuments/api/v1/template/fields/fill/406\?convert\=pdf\&destination\=myfile.pdf
```


julien-nc (approved):
Tested locally, works like a charm.
Code looks good.
Nice that it now prints the document content if destination is null!
Thanks a lot :blue_heart:

juliushaertl:
Pushed some integration tests for the endpoints

elzody (approved):
Gave it a thorough test, works nicely. Also nice code

PR #3958, opened 2024-08-29
fix: Check for mimetype before processing new file from template

juliushaertl:
Fix https://github.com/nextcloud/whiteboard/issues/113

Otherwise we will try to apply richdocuments logic to other template files


mejo- (approved):
Looks sensible

PR #3960, opened 2024-08-29
fix: Emit notification on mentions

juliushaertl:
It seems we totally missed this part when implementing the initial mention support back in the days.

This will emit a notification to mentioned users that can be clicked, just like it works in text.


lib/Controller/MentionController.php
```diff
@@ -0,0 +1,66 @@
...
+
+		$notification = $this->manager->createNotification();
+		$notification->setUser($mention)
+			->setApp(Application::APPNAME)
+			->setSubject(Notifier::TYPE_MENTIONED, [
+				Notifier::SUBJECT_MENTIONED_SOURCE_USER => $this->userId,
+				Notifier::SUBJECT_MENTIONED_TARGET_USER => $mention,
+			])
+			->setObject('file', (string)$fileId);
+		;
```
elzody, line 55: random semicolon maybe?

juliushaertl: Right, fixed 👍

elzody (approved):

PR #3967, opened 2024-09-02
fix: Catch parsing errors or null doc structure

juliushaertl:
Fixes errors like:

- [PHP] foreach() argument must be of type array|object, null given
- [PHP] Undefined array key "DocStructure"

### TODO

- [ ] ...

### Checklist


blizzz (approved):

juliushaertl:
CI fails due to collabora docker images

Issue #3990, opened 2024-09-05
Guest name dialog can't submit guest name with `Enter` key


elzody:
**Describe the bug**
When opening an office document as a guest, you are asked to input a guest name. The dialog which appears for this does not allow you to submit the name by pressing the `Enter` key; you are instead forced to use your mouse to click the "Submit name" button.

**To Reproduce**
1. Be logged out
2. Visit a public share link to an office document
3. Type in a guest username in the dialog
4. Pressing `Enter` does not submit the name

**Expected behavior**
The name will be submitted either by clicking the "Submit name" button, or alternatively by pressing `Enter` on the keyboard, which allows for quicker movement.

**Screenshot**
<details>
<summary>Example</summary>

![image](https://github.com/user-attachments/assets/a9f96c21-f500-4974-b5d0-4e411a4152c9)
</details>

PR #3992, opened 2024-09-05
fix: pressing `Enter` submits guest name input

elzody:
* Resolves: #3990 
* Target version: main

### Summary
With this PR the guest name input field is automatically focused, and you can submit it by pressing the `Enter` key. This makes it faster to input the name and get to the document, and follows the behavior users expect.

This is done by moving the dialog content into a `<form>` element, and using the form submit action. It also includes a few styling changes which I think make the dialog look better (but I'm not a designer :D)

### Example
<details>
<summary>Before</summary>

image
</details>

<details>
<summary>After (video)</summary>

[vokoscreenNG-2024-09-05_14-00-57.webm](https://github.com/user-attachments/assets/92482037-a959-4366-ab61-086a935d8656)
</details>

### Checklist

### DIFF
--- a/src/components/GuestNamePicker.vue
+++ b/src/components/GuestNamePicker.vue
@@ -24,7 +24,8 @@
 			</p>
 
 			<fieldset>
-				<NcTextField :value="guestName"
+				<NcTextField ref="guestNameInput"
+					:value="guestName"
 					data-cy="guestNameInput"
 					:label="t('richdocuments', 'Guest name')"
 					:placeholder="t('richdocuments', 'Anonymous guest')"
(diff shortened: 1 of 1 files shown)


blizzz (approved):

src/components/GuestNamePicker.vue
```diff
@@ -7,7 +7,9 @@
 		:out-transition="true"
 		size="small"
 		:show.sync="show">
-		<div class="modal__content" data-cy="guestNameModal">
+		<form class="modal__content"
```
susnux, line 10: Not needed, just migate this to `NcDialog` and set `is-form` prop :)

elzody: That sounds more reasonable, I will take a look :D

marcoambrosini (approved):

jancborchardt (approved):
Nice detail fix, also good on the layout update! :)

PR #3994, opened 2024-09-06
fix missing ':' prefix

hcvcastro:
* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


juliushaertl (approved):

timur-g:
@juliushaertl thanks. Can we have this also in stable30? That is more practical to test in various servers..

timur-g:
Excuse me, if this is in stable30, was it supposed to be in 8.5.0, as it is not listed in the Release note?

Issue #3957, opened 2024-08-28
Smart picker not working for document sections


pedropintosilva:
**Describe the bug**
I have noticed today that the https://sdk.collaboraonline.com/docs/linking_api.html is not working: https://paste.opensuse.org/pastes/3be00af7d685 . 

https://github.com/user-attachments/assets/a0de5e19-1f28-45c7-8c07-c246af608a73

I could reproduce this with current master and stable but also I have tried with a slightly older combination and I still get the same result:
* COOL 24.04.4.2 git hash: fbf97e9 (E)
* Richdocuments 8.4.3

juliushaertl:
Fails at https://github.com/nextcloud-libraries/nextcloud-dialogs/blame/144923cd026110e33e824b445dffad827b91a8b1/lib/utils/dialogs.ts#L19 as we pass in an element not a selector

juliushaertl:
We should actually use https://github.com/nextcloud-libraries/nextcloud-dialogs?tab=readme-ov-file#use-the-vue-component-directly instead

PR #3995, opened 2024-09-06
fix: use file picker component directly

elzody:
* Resolves: #3957 
* Target version: main

### Summary
The previous implementation stopped working because the file picker's `container` prop expects a selector, but an element was passed instead. This PR fixes that, and uses the file picker component directly.

### Checklist


src/view/DocumentTargetPicker.vue
```diff
@@ -5,7 +5,7 @@
 
 <template>
 	<div v-if="filePath === null" class="office-target-picker">
-		<div ref="picker" class="reference-file-picker" />
+		<div class="reference-file-picker" />
```
juliushaertl, line 8: We could actually use the vue component directly https://github.com/nextcloud-libraries/nextcloud-dialogs?tab=readme-ov-file#use-the-vue-component-directly

That way we would not have nested modals

elzody: Indeed, it has been improved, however the FilePickerVue component itself is actually wrapped in an NcDialog
https://github.com/nextcloud-libraries/nextcloud-dialogs/blob/016d046e79b949838bb78d01dca9bcb1e1db286a/lib/components/FilePicker/FilePicker.vue#L6

juliushaertl: All right, then fine either way 😄

juliushaertl:
Could you also add a cypress test for that?

juliusknorr:
@elzody took me quite a bit but found why this failed. On your local setup it likely worked because you had text enabled which dispatches an RenderReferenceEvent. This event is listened for to [provide the list of providers](https://github.com/nextcloud/server/blob/dae7c159f728a90ffa53247d6e033abdae5d2bd6/lib/private/Collaboration/Reference/RenderReferenceEventListener.php#L41). I left my commits with debugging attempts for you to follow along for now. Feel free to clean them up and get this merged.

After seeing that the initial state was empty if checked where in the server that was set. Then I locally set a breakpoint to see which event dispatching actually triggers the listener on my local setup. That was text, so disabling text showed the same beahavior as CI when running the cypress test.

elzody:
@juliusknorr Thanks for the help, seems like it required some really deep debugging. I think I managed to clean up the nonsense debug commits, and it seems like everything passes now!

juliusknorr (approved):

elzody:
@juliusknorr Yep, only back to stable30. I already have a PR open for it: https://github.com/nextcloud/richdocuments/pull/4028

I just need to replicate what you added here and get it cleaned up

Issue #3170, opened 2023-09-08
[F2V] Migrate new file menu to new API


juliushaertl:
- [ ] Check logged in file handling
- [ ] New file menu for public share links

juliushaertl:
For 28 internal menu is handled by the template creation, public pages are not migrated to vue yet

juliushaertl:
Some progress in https://github.com/nextcloud/server/pull/45652 probably targetting 31

PR #4014, opened 2024-09-10
Change some element selectors to fix a few Cypress tests

elzody:
### Summary
Some selectors changed for the files app, so we correct those here. There are other issues which still cause some tests to fail, but they are deeper issues which may not be entirely worth looking into immediately as things are still in flux in the server repo and things are changing.

### Checklist

### DIFF
--- a/cypress/e2e/share-link.js
+++ b/cypress/e2e/share-link.js
@@ -84,7 +84,8 @@ describe('Public sharing of office documents', function() {
 						},
 					})
 
-					cy.get('tr[data-file="document.odt"] a.name').click()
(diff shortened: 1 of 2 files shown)


juliushaertl (approved):
Looks good, I'd say we skip the remaining failing test and add a checkbox to https://github.com/nextcloud/richdocuments/issues/3170 to bring it back once we have the server template api working for public share links

juliushaertl:
Pushed a commit to skip the test

juliusknorr:
Seems some changes in selectors have been backported in server

PR #4030, opened 2024-09-12
fix: failed token generation on public shares

elzody:
* Target version: main

### Summary
This is part of an effort to improve public sharing on the main branch with all the server changes for NC 31.
The issue here was that the sharing token was grabbed from an invisible html element that contained the token, which is no longer present -- the token instead must be gotten via either initial state from files_sharing or via this method using the `@nextcloud/sharing` library which I think is clearer. It also will be helpful for other things relating to public shares

https://nextcloud-libraries.github.io/nextcloud-sharing/modules/public.html

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "@nextcloud/paths": "^2.2.1",
         "@nextcloud/router": "^3.0.1",
         "@nextcloud/vue": "^8.18.0",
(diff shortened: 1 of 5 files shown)


juliushaertl (approved):

juliushaertl:
There seem to be other occurences where shareToken is still obtained from an element by id, maybe worth to check those as well separately if they are still in use

PR #4043, opened 2024-09-17
feat(sharing): use `@nextcloud/sharing` to detect is public share

elzody:
### Summary
Changes occurrences where `isPublic` was trying to read from an invisible element on the page and now use the `@nextcloud/sharing` library which is cleaner and works now with the changes on server for 31.

### Checklist


juliushaertl:
Actually the cypress failures on templates seem odd as they should work as logged in users

>   3) Create new office files from templates

elzody:
@juliushaertl It's possible they are not yet using the correct selectors since the `fix/why-ci-not-green` branch has not been merged yet which fixes the selectors

PR #4050, opened 2024-09-19
fix: Properly render compare view

juliushaertl:
Requires https://github.com/nextcloud/viewer/pull/2549


elzody (approved):
Works well paired with the viewer PR [nextcloud/viewer#2549](https://github.com/nextcloud/viewer/pull/2549)

PR #4101, opened 2024-10-07
fix: Use proper file id for direct editing

juliusknorr:
As reported by @marinofaggiana there was an issue creating new files on iOS which I could reproduce with curl.

I assume this broke with some refactoring, but not sure where, requiring a test to avoid future regressions

Still to do:
- [ ] Test with iOS/Android
- [x] I think we should have tests for this, double check or add

### DIFF
--- a/cypress/e2e/direct.spec.js
+++ b/cypress/e2e/direct.spec.js
@@ -2,6 +2,22 @@
  * SPDX-FileCopyrightText: 2023 Julius Härtl <jus@bitgrid.net>
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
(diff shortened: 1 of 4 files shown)


juliusknorr:
Added a cypress test for this

juliusknorr:
@elzody I fail to validate this locally on iOS/Android due to issues with my build environments with that. I'd vote to review and merge, then we can test on the daily instance with real devices and we can backport once double checked.

elzody (approved):

juliusknorr:
All right, seems something is still odd, when testing on iOS i end up with 0kb files. Need to figure out what is different there compared to the cypress tests.

PR #4124, opened 2024-10-14
Fix: Fix Admin Settings page when Watermark label selected

Koc:
* Target version: main

### Summary
Reproducer:
1. Open settings
2. Enable watermark that will apply on files with some tags
3. Refresh page

Expected result:
It is possible view and change tag that used for watermark assigning.

Actual result:
There is an error `Error in render: "TypeError: option is undefined"`.
image

### TODO
- [ ] Backport to stable 29

### Checklist

### DIFF
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -48,6 +48,11 @@ class AppConfig {
 		'watermark_linkTagsList' => 'array'
 	];
 
+	private const INTEGER_LIST_KEYS = [
+		'watermark_allTagsList' => true,
+		'watermark_linkTagsList' => true,
+	];
+
 	public function __construct(
 		private IConfig $config,
 		private IAppManager $appManager,
@@ -117,7 +122,11 @@ class AppConfig {
 				$value = $value === 'yes' ? true : $value;
 				$result[$key] = $value === 'no' ? false : $value;
 			}
+			if (!empty(self::INTEGER_LIST_KEYS[$key])) {
+				$result[$key] = array_map('intval', $result[$key] ?? []);
+			}
 		}
+
 		return $result;
 	}
 


lib/AppConfig.php
```diff
@@ -118,6 +123,10 @@ public function getAppSettings() {
 				$result[$key] = $value === 'no' ? false : $value;
 			}
 		}
+		if (!empty(self::INTEGER_KEYS[$key])) {
+			$result[$key] = array_map('intval', $result[$key]);
+		}
```
elzody, line 128: This does indeed fix it, the only issue is that the block you added here needs to be added *within* the `foreach` loop, so it can access `$key`. Tried it on my end, and it does resolve it this way.

```suggestion
foreach ($keys as $key) {
    if (str_starts_with($key, 'watermark_')) {
        $value = $this->getAppValueArray($key);
        $value = $value === 'yes' ? true : $value;
        $result[$key] = $value === 'no' ? false : $value;
    }
    // This needs to be inside the foreach loop to get $key
    if (!empty(self::INTEGER_KEYS[$key])) {
        $result[$key] = array_map('intval', $result[$key]);
    }
}
// It used to be here... but then it can't get the $key
```

elzody (changes requested):
Looks okay, only one suggestion

lib/AppConfig.php
Koc: Good catch, thank you. I haven't tested this code yet :point_right: :point_left:

Koc: tested manually :heavy_check_mark: 
image

elzody (approved):
The failing tests are unrelated, looks good to me. Could not reproduce the original issue on stable28, so only backporting to stable29.

PR #4125, opened 2024-10-14
Move template settings to vue component

elzody:
* Target version: main

### Summary
This resolves the issue on main where we are still using a legacy file upload plugin, causing an error and failing tests. I took the chance to migrate the global template settings to a Vue component, a start which I hope will lead to a cleanup of the admin settings in general, since everything is shoved into one large `AdminSettings.vue` component, and it's messy and difficult to reason about.

### TODO

- [x] add cypress tests

### Checklist


cypress/e2e/templates.spec.js
```diff
@@ -3,7 +3,9 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-describe('Create new office files from templates', function() {
+import {User} from "@nextcloud/cypress";
+
+describe.only('Global templates', function() {
```
juliusknorr, line 8:
```suggestion
describe('Global templates', function() {
```

juliusknorr (approved):
Tested and works

cypress/e2e/templates.spec.js
juliusknorr: Pushed a fix

juliusknorr (approved):

Issue #4123, opened 2024-10-14
Digitally sign documents via software certificates.


vmiklos:
**Is your feature request related to a problem? Please describe.**

Collabora Online is adding digital signature support, see https://github.com/CollaboraOnline/online/issues/9992 for the details. This is something that richdocuments could use.

**Describe the solution you'd like**

The solution is meant to provide signing support for ODF (e.g. ODT), OOXML (e.g. DOCX) and PDF files.

We discussed this with @juliushaertl about a month ago, but back then the Collabora Online side was far from ready. The suggested plan is:

- Step 1: Exposing signing via software certificates
  -  Key store in richdocuments
  - Signing can happen in Collabora Online
  - One key per user (in PEM format, this means a key, a certificate and optionally a CA chain)
  - this issue.
- Step 2: build electronic signature support on top of this, via some external provider
  - this is interesting, but let's get back to this once step 1 is done.

**Describe alternatives you've considered**

It's always possible to export as PDF and sign that PDF outside Collabora Online, if you don't mind the export / download, sign, upload steps, which is a bit complicated.

**Additional context**

I intend to work on this, just creating the issue to track the richdocuments progress at a single place.

PR #4129, opened 2024-10-16
Digitally sign documents via software certificates

vmiklos:
* Resolves: https://github.com/nextcloud/richdocuments/issues/4123
* Target version: main

### Summary

This implements step 1 of digital signature signing, where the keystore is in richdocuments and the signing is performed in Collabora Online, not via some external signing service. This supports ODF, OOXML and PDF formats.

### Checklist

- [x] Code is properly formatted: `npm run stylelint` passes
- [ ] Documentation (manuals or wiki) has been updated or is not required: this is not done in this PR, maybe we should link to the Collabora Online SDK to explain how to generate signing keys?

### How to test

1) Make sure to have a new enough COOL snapshot, so its /hosting/capabilities reports hasDocumentSigningSupport=true. This is true for online.git master as of >= 2024-11-09.

1) Go to /settings/user/richdocuments

2) Fill in the 3 new signing settings. For testing, you can generate the certificates with a script described at https://github.com/CollaboraOnline/online/issues/9992#issuecomment-2396115427 or just use these:

- cert: [test.odt.cert.pem.txt](https://github.com/user-attachments/files/17391222/test.odt.cert.pem.txt)
- key: [test.odt.cert.key.txt](https://github.com/user-attachments/files/17391229/test.odt.cert.key.txt)
- ca: [test.odt.cert.ca.txt](https://github.com/user-attachments/files/17391232/test.odt.cert.ca.txt)

3) Load e.g. an ODT file in Nextcloud Office. Go to the File tab, press the Signature button to open the Signatures dialog. Press the Sign Document button, select the detected certificate, press the Sign button: "the signatures in this document are valid" text appears.

Let me know if you would like any tweaks. Thanks.

### DIFF
--- a/src/personal.js
+++ b/src/personal.js
@@ -17,6 +17,12 @@ import { showError } from '@nextcloud/dialogs'
 		this.zoteroAPIKeySaveButton = document.getElementById('zoteroAPIKeySave')
 		this.zoteroAPIKeyRemoveButton = document.getElementById('zoteroAPIKeyRemove')
 
+		this.documentSigningCertInput = document.getElementById('documentSigningCertField')
+		this.documentSigningCertSaveButton = document.getElementById('documentSigningCertSave')
+		this.documentSigningCertRemoveButton = document.getElementById('documentSigningCertRemove')
+		this.documentSigningKeyInput = document.getElementById('documentSigningKeyField')
+		this.documentSigningKeySaveButton = document.getElementById('documentSigningKeySave')
+		this.documentSigningKeyRemoveButton = document.getElementById('documentSigningKeyRemove')
 		this.documentSigningCaInput = document.getElementById('documentSigningCaField')
 		this.documentSigningCaSaveButton = document.getElementById('documentSigningCaSave')
 		this.documentSigningCaRemoveButton = document.getElementById('documentSigningCaRemove')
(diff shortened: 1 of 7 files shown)


src/personal.js
```diff
@@ -17,6 +17,16 @@ import { showError } from '@nextcloud/dialogs'
 		this.zoteroAPIKeySaveButton = document.getElementById('zoteroAPIKeySave')
 		this.zoteroAPIKeyRemoveButton = document.getElementById('zoteroAPIKeyRemove')
 
+		this.documentSigningCertInput = document.getElementById('documentSigningCertField')
```
juliushaertl, line 20: It would be great if we can rather add this to the vue.js based frontend, but maybe that is something where @elzody can help moving that over

vmiklos: Oh, I was just tracing how the zotero API key setting does this, and this way seemed to be the way it works. Maybe move the zotero API key handling to the vue.js based frontend, and then I'm happy to do the same for these settings as well? Thanks.

vmiklos:
An other problem I notice is that 5 cypress jobs are red, but it seems that the tip of 'main' has the same problem. Is it OK to ignore those failures for now? Thanks.

juliushaertl:
That is unfortunately known (due to larger refactorings and legacy code removal in server for the upcoming release), we're on that but can be ignored for now.

src/personal.js
juliushaertl: Ok, also wouldn't consider this a blocker then especially since zotero is also still there in the legacy part. We can migrate them together afterwards as well

vmiklos:
OK thanks, good to know. Is there anything else I can help with to get this reviewed? I assume that it only makes sense to rebase the branch if there would be conflicts, correct? (Even if the github interface nags with an "update branch" button.)

Thanks.

juliushaertl:
Yeah that is fine, I also requested a review from @elzody and lined it up for me

elzody (approved):
Works well on my end. Code looks good for now, will probably be reorganized in a future refactoring

vmiklos:
@elzody thanks for the review! :-)

PR #4132, opened 2024-10-17
Fix asset download checks

juliushaertl:
- **fix: Skip disabled download files when requesting assets**
- **fix: Filter out non-downloadable files in image picker**

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


elzody (approved):
Works great, cypress failures are unrelated

PR #4133, opened 2024-10-17
[stable30] Fix asset download checks

juliushaertl:
Backport of #4132


elzody (approved):
Also works here on stable30

Issue #4169, opened 2024-10-22
Mention notification request fails with 500 status code


Rash419:
**Describe the bug**
Mention notification request fails with 500 status code

**To Reproduce**
1. Create a account that have ' '(space) in the Display Name for example: `Alison Burger`
2. Share Document between two users
3. mention `@Alison Burger` in the document
4. Check console

![Image](https://github.com/user-attachments/assets/e467d527-e50e-48f5-80d0-5339db82b12c)

Stacktrace:
```
/We</<
[xhr.js:195:12](webpack:///richdocuments/node_modules/axios/lib/adapters/xhr.js)
715/We<
[xhr.js:15:9](webpack:///richdocuments/node_modules/axios/lib/adapters/xhr.js)
vt
[dispatchRequest.js:51:17](webpack:///richdocuments/node_modules/axios/lib/core/dispatchRequest.js)
_request
[Axios.js:173:32](webpack:///richdocuments/node_modules/axios/lib/core/Axios.js)
request
[Axios.js:40:24](webpack:///richdocuments/node_modules/axios/lib/core/Axios.js)
715/</t/<
[Axios.js:212:18](webpack:///richdocuments/node_modules/axios/lib/core/Axios.js)
715/a/<
[bind.js:5:14](webpack:///richdocuments/node_modules/axios/lib/helpers/bind.js)
uiMention
[uiMention.js:38:16](webpack:///richdocuments/src/mixins/uiMention.js)
postMessageHandler
[Office.vue:433](webpack:///richdocuments/src/view/Office.vue)
e.prototype.handlePostMessage/<
[postMessage.tsx:74:19](webpack:///richdocuments/src/services/postMessage.tsx)
e.prototype.handlePostMessage
[postMessage.tsx:68:33](webpack:///richdocuments/src/services/postMessage.tsx)
e/<
[postMessage.tsx:12:18](webpack:///richdocuments/src/services/postMessage.tsx)
(Async: EventListener.handleEvent) e
[postMessage.tsx:11:15](webpack:///richdocuments/src/services/postMessage.tsx)
```

Request:
```
{"mention":"Alexandra Martini"}
```

Response:
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>500</statuscode>
  <message>Internal Server Error
</message>
 </meta>
 <data/>
</ocs>
```

**Expected behavior**
No errors and user should be notified

**Nextcloud log entry**
[...]

PR #4171, opened 2024-10-22
fix: get user by display name

elzody:
* Resolves: #4169
* Target version: main

### Summary
When mentioning a user within a document from Collabora, it seems to search by the user's display name. Thus, if the display name has any whitespace, it will fail to mention the user. The problem is that the backend tries to use the mentioned user string (display name in this case) as the UID, which is not the same. The solution is to first search by the display name to get the proper user object, then use the UID from that to generate the notification and send it to the user.

### Checklist

### DIFF
--- a/lib/Controller/MentionController.php
+++ b/lib/Controller/MentionController.php
@@ -16,6 +16,7 @@ use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\Files\IRootFolder;
 use OCP\IRequest;
+use OCP\IUserManager;
 use OCP\Notification\IManager;
 
 class MentionController extends Controller {
@@ -25,6 +26,7 @@ class MentionController extends Controller {
 		private IRootFolder $rootFolder,
 		private IManager $manager,
 		private ITimeFactory $timeFactory,
+		private IUserManager $userManager,
 		private ?string $userId,
 	) {
 		parent::__construct($appName, $request);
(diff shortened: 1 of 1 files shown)


lib/Controller/MentionController.php
```diff
@@ -39,14 +41,20 @@ public function mention(int $fileId, string $mention): DataResponse {
 			return new DataResponse([], Http::STATUS_NOT_FOUND);
 		}
 
-		$userFolder = $this->rootFolder->getUserFolder($mention);
+		$userResults = $this->userManager->searchDisplayName($mention, 1);
+		if (count($userResults) < 1) {
+			return new DataResponse([], Http::STATUS_NOT_FOUND);
+		}
+
+		$user = $userResults[0];
```
blizzz, line 49: if you want to go safe, use `array_key_first()` or `array_shift` instead of the hard `0` index.

blizzz (approved):
Comes with a suggestion.

lib/Controller/MentionController.php
```diff
@@ -39,14 +41,20 @@ public function mention(int $fileId, string $mention): DataResponse {
 			return new DataResponse([], Http::STATUS_NOT_FOUND);
 		}
 
-		$userFolder = $this->rootFolder->getUserFolder($mention);
+		$userResults = $this->userManager->searchDisplayName($mention, 1);
```
juliusknorr, line 44: I think this is a design flaw of the post message API that is not returning us a real unique identifier. Worth to file upstream to further discuss, but fine as a workaround with me now.

blizzz:
What comes to my mind now is that it probably does not respect various sharing settings, e.g. auto completion of users. Unless the "mention" is filtered somewhere before. Maybe the autocomplete logic (`OCP\Collaboration\AutoComplete\IManager`) would be better?

lib/Controller/MentionController.php
elzody: File upstream where exactly? Should we rather have Collabora return the UID rather than display name?

elzody:
> What comes to my mind only now is that it probably does not respect various sharing settings, e.g. auto completion of users.

Collabora send a request to this mention endpoint with the display name already; it's given to us by Collabora, so I'm not sure if this applies, what do you think?

@blizzz

PR #4178, opened 2024-10-26
skip css variable if not set (prevent warning on COOL side)

timar:
* Resolves: # 
* Target version: main

related: https://github.com/CollaboraOnline/online/pull/10335


elzody:
It looks like this just needs a rebase to get all the tests passing, and your commit is missing a sign-off so the DCO check is failing.

elzody (approved):

PR #4195, opened 2024-11-04
fix(share): link share of federated share

max-nextcloud:
* Target version: main

### Summary

Fix opening link shares to federated shares.

I know that's complicated so let me explain:
* Alice@one.example shares a file with Bob@two.example via federated sharing.
* Bob creates a public link share to that file.
* When a guest visits that share it should open nicely.

### TODO

- [ ] Add cypress test.

### Checklist

- [x] Documentation is not needed.

### DIFF
--- a/cypress/e2e/share-federated.spec.js
+++ b/cypress/e2e/share-federated.spec.js
@@ -16,6 +16,7 @@ describe('Federated sharing of office documents', function() {
 		cy.createUser(shareOwner)
 
 		cy.uploadFile(shareOwner, 'document.odt', 'application/vnd.oasis.opendocument.text', '/document.odt')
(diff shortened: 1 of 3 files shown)


juliusknorr (approved):
👍 

Adding tests would be awesome. The other cypress failures in here are unrelated

juliusknorr:
Rebased as main should be passing again 🥳

juliusknorr:
@max-nextcloud I pushed a small test case as I was checking out federation anyways

max-nextcloud:
> @max-nextcloud I pushed a small test case as I was checking out federation anyways

Thanks a lot!

Then this is ready to go, right?

juliusknorr:
Do you need backports to specific versions?

PR #4207, opened 2024-11-07
fix(headers): only add Cross-Origin-Opener-Policy and Cross-Origin-Embedder-Policy on richdocuments

tcitworld:
* Resolves: #4103
* Target version: main

### Summary
Otherwise it's being added to absolutely all requests, creating issues in other apps.

However, since I don't know how to test this and why these headers are required, I'm not sure if this still works for public sharing pages for instance.

Ref. https://github.com/nextcloud/richdocuments/issues/3258#issuecomment-2398139667

### Checklist

### DIFF
--- a/lib/Listener/BeforeTemplateRenderedListener.php
+++ b/lib/Listener/BeforeTemplateRenderedListener.php
@@ -7,6 +7,7 @@ declare(strict_types=1);
  */
 namespace OCA\Richdocuments\Listener;
 
+use OCA\Richdocuments\AppInfo\Application;
 use OCA\Richdocuments\Service\CapabilitiesService;
 use OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent;
 use OCP\EventDispatcher\Event;
@@ -24,7 +25,7 @@ class BeforeTemplateRenderedListener implements IEventListener {
 			return;
 		}
 
-		if ($this->capabilitiesService->hasWASMSupport()) {
+		if ($this->capabilitiesService->hasWASMSupport() && $event->getResponse()->getApp() === Application::APPNAME) {
 			$event->getResponse()->addHeader('Cross-Origin-Opener-Policy', 'same-origin');
 			$event->getResponse()->addHeader('Cross-Origin-Embedder-Policy', 'require-corp');
 		}


lattaai3:
Hello, I tried to fix the issue

    This is what I did:
    This change adds a page load check to both AddContentSecurityPolicyListener and AddFeaturePolicyListener to ensure that the policies are only added on full page loads. This should resolve the issue of the headers being added to all requests, which was causing problems in other apps. The change maintains functionality for public sharing pages by keeping the existing logic intact.

    I also created pull request: https://github.com/nextcloud/richdocuments/pull/4208

    
    > [!CAUTION]
    > Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
    >
    > The fix provided by Latta AI might not be complete and it can serve as an inspiration.

    ---
    This bug was fixed for free by Latta AI - https://latta.ai/ourmission

    If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

lib/Listener/BeforeTemplateRenderedListener.php
```diff
@@ -24,7 +25,7 @@ public function handle(Event $event): void {
 			return;
 		}
 
-		if ($this->capabilitiesService->hasWASMSupport()) {
+		if ($this->capabilitiesService->hasWASMSupport() && $event->getResponse()->getApp() === Application::APPNAME) {
```
juliusknorr, line 28: As far as I remember this will not work as we require the headers to be set e.g. on the files app as it is the one that iframes Collabora directly. 

@caolanm Since you worked on that, can you provide some steps how to test and verify that the wasm support still fully works with that change?

tcitworld: If not, would it be acceptable to hardcode `'files'` (and probably `'files_sharing'`) as well here?

juliusknorr: Depends, in theory collabora could be loaded by any app, e.g. talk when using the viewer to show a file shared to a conversation.

juliusknorr: Though it is arguable what benefit the wasm experimental feature has there

Issue #4051, opened 2024-09-20
Enable startPresentation parameter


timur-g:
Collabora Online supports starting presentation with any presentation file, appending `?startPresentation=true` after file URL.
https://sdk.collaboraonline.com/docs/advanced_integration.html?highlight=startpresentation#url-query-parameters.
But Nextcloud seems to remove the parameter from the url. Would be good to support this.

elzody:
I am getting errors like the following screenshots when I try to append this to the URL. Any ideas?

@timur-g 

<details>
<summary>Browser error</summary>

![Image](https://github.com/user-attachments/assets/9f531c64-a348-4e23-b503-3ea1d4d38d3c)
</details>

<details>
<summary>Collabora error</summary>

![Image](https://github.com/user-attachments/assets/5d67aa68-c95d-484f-9b16-e5bc7925be0f)
</details>

timur-g:
Collabora error fixed with https://github.com/CollaboraOnline/online/issues/10198.

elzody:
I've done some research into how we could implement this in richdocuments. So far, this is my analysis and proposed solution:

First, I tried to solve the problem of the `&startPresentation=true` parameter being removed from the URL when you try to open a presentation from within Nextcloud. This ended up being due to the file action which the viewer app registers; it uses `OCP.Files.Router.goToRoute()` to navigate to a very specific route, providing its own queries / parameters (thus erasing all others you give it manually). I cannot find a workaround for getting the parameter in richdocuments unless I use some hacky solution or something changes with how the viewer handles this, but with such a small addition I doubt it would be worth it to change anything there.

With that being said, my proposed solution would be to register a file action on presentation files which would immediately open them in presentation mode. It's one more click to show the file action menu, but would provide a cleaner solution in my opinion.

Any thoughts or ideas which I haven't thought of?
@juliusknorr and @timur-g (in case you want to stay up-to-date)

juliusknorr:
A quick option would be to check for the parameter on page load in PHP and pass it to the richdocuments frontend code as initial state or store it in the session if it needs to persist multiple page reloads due to redirects. 

Cleaner option would be if the files app could preserve url parameters somehow, but might be larger effort

klorydryk:
When this fix could be added in a Richdocument release ? It seems that it is still not available in the 8.6.0 beta1
Thanks!

juliusknorr:
[...]

PR #4209, opened 2024-11-07
feat: `startPresentation` URL parameter

elzody:
* Resolves: #4051
* Target version: main

### TODO

- [ ] Cypress tests?

### Checklist

### DIFF
--- a/src/helpers/url.js
+++ b/src/helpers/url.js
@@ -29,7 +29,7 @@ const getWopiSrc = (fileId) => {
 	return wopiurl
 }
 
-const getWopiUrl = ({ fileId, title, readOnly, closeButton, revisionHistory, target = undefined }) => {
+const getWopiUrl = ({ fileId, title, readOnly, closeButton, revisionHistory, target = undefined, startPresentation = false }) => {
 	// Only set the revision history parameter if the versions app is enabled
 	revisionHistory = revisionHistory && window?.oc_appswebroots?.files_versions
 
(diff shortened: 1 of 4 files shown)


src/helpers/url.js
```diff
@@ -45,6 +45,7 @@ const getWopiUrl = ({ fileId, title, readOnly, closeButton, revisionHistory, tar
 		+ (revisionHistory ? '&revisionhistory=1' : '')
 		+ (readOnly ? '&permission=readonly' : '')
 		+ (target ? '&target=' + encodeURIComponent(target) : '')
+		+ (Config.get('startPresentation') ? '&startPresentation=1' : '')
```
juliusknorr, line 48: I think we should read into a variable once at page load and make sure to reset it to false if it was used once

elzody: Once the page is refreshed (which, in its current state, it has to be) then the initial state will reset and `startPresentation` will not be enabled unless you again append it to the URL. I'm assuming once I implement a way to close the presentation mode this way, then it would be relevant?

elzody:
i briefly spoke with méven from collabora in the matrix chat about this, and this was the response:
> That's the current expected behavior, due to browser limitation, we can't open a real external presentation window so COOL is replaced by the presentation logic, this could be worked-around though. Browsers have added more restriction since the feature was implemented, this would need to be fixed  for certain browsers

therefore, at least for now, given having to refresh the page, this should work

juliusknorr:
All right, then let's stick with that as intended behavior

juliusknorr (approved):

juliusknorr:
Not sure if anything still left as it is in work in progress state still, but looks good from my side 👍

PR #4210, opened 2024-11-08
feat: use autocomplete api for mentions

elzody:
* Target version: main

### Summary
Try to use the autocomplete API to search for the user by display name instead so that the instance sharing settings regarding autocompletion are respected.

⚠️ **We probably want to wait for [CollaboraOnline/online#10556](https://github.com/CollaboraOnline/online/pull/10556) to be merged**

### Checklist

### DIFF
--- a/lib/Controller/MentionController.php
+++ b/lib/Controller/MentionController.php
@@ -14,10 +14,12 @@ use OCP\AppFramework\Http\Attribute\NoAdminRequired;
 use OCP\AppFramework\Http\Attribute\UserRateLimit;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\Collaboration\Collaborators\ISearch;
 use OCP\Files\IRootFolder;
 use OCP\IRequest;
 use OCP\IUserManager;
 use OCP\Notification\IManager;
+use OCP\Share\IShare;
 
 class MentionController extends Controller {
 	public function __construct(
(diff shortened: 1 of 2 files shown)


lib/Controller/MentionController.php
```diff
@@ -41,22 +44,21 @@ public function mention(int $fileId, string $mention): DataResponse {
...
+		$user = array_pop($matchedUsers);
+		$userFolder = $this->rootFolder->getUserFolder($user['shareWithDisplayNameUnique']);
 		$file = $userFolder->getFirstNodeById($fileId);
 		if ($file === null) {
 			return new DataResponse([], Http::STATUS_NOT_FOUND);
 		}
 
 		$notification = $this->manager->createNotification();
-		$notification->setUser($user->getUID())
+		$notification->setUser($user['shareWithDisplayNameUnique'])
```
juliusknorr, line 61: We need the user id to send the notification to the specific user, not the displayname, iirc it should be `$user['shareWith']`

elzody: Thanks for catching that, I'll give it a fix

elzody: An interesting thing I noticed when working on this is that, the `shareWithDisplayNameUnique` array value actually always points to the user ID, whereas the `label` value is what points to the actual non-unique display name. `shareWith` does in fact point to the user ID as well, so I'm not sure what the difference is. The notification gets sent just fine in either case -- is it preffered anyway to use `shareWith` as opposed to `shareWithDisplayNameUnique`? @juliusknorr

blizzz: @elzody your test users do not have email set, do they? Yes, please use `shareWith`.

elzody: It makes sense now, thanks Arthur

juliusknorr (approved):

juliusknorr:
@elzody Maybe you can check how far we should backport this one

PR #4249, opened 2024-11-19
Implement support for insertion of multimedia from Nextcloud assets

mikekaganski:
This change adds ability to insert multimedia files from Nextcloud to Impress
documents. It advertises EnableInsertRemoteFile capability, so that Office
activates the feature (requires Office including commit 60d9a9c20bf69b3f1d8591b2d8400c4a453970ab;
older builds won't show the respective button).

### DIFF
--- a/docs/frontend-integration.md
+++ b/docs/frontend-integration.md
@@ -100,6 +100,10 @@ The following handlers are currently supported:
 - insertGraphic: will be called when an image from the Nextcloud storage should be inserted
   - Arguments
     - insertFileFromPath(path): Callback to trigger the actual inserting of the graphic from an absolute file path
+- insertFile: will be called when a file (e.g., multimedia) from the Nextcloud storage should be inserted (generalized insertGraphic)
+  - Arguments
+    - mimeTypeFilter: array of MIME types (strings) to filter in the UI
+    - insertFileFromPath(path): Callback to trigger the actual inserting of the file from an absolute file path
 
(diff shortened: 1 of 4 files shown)


elzody:
i think i understand the change here, but as i'm trying to test it out, i can't tell how i'm actually supposed to use it. the `CheckFileInfo` is properly returning the `EnableInsertRemoteFile` as `true`, but i am only able to insert graphics (images). is there a way to test insertion of these multimedia files as well?

forgive me if i'm oblivious :)

mikekaganski:
@elzody thanks for testing!
This change relies on the commit to Online, that I only merged there today (https://github.com/CollaboraOnline/online/commit/60d9a9c20bf69b3f1d8591b2d8400c4a453970ab). If you tested using that version (current development branch), it must work. This change is basically to make richdocuments ready for the feature on Online side.

elzody:
I gave it another test, and I do indeed see the new button there! I am unfortunately unable to properly test it though... for several days I was trying to get it to work, but it seems on master branch of Collabora Online I can no longer insert any image or media into any office document. It seems to be an upstream issue with Collabora from what I can tell, do you also experience the same?

elzody (approved):
Thank you for the animation -- it was helpful in clarifying things! :) I managed to test it and it seems to work just fine. The code changes look decent to me. Thanks for contributing to this!

thebearon:
@elzody Would this be safe to port into stable30 branch?

elzody:
@thebearon It *probably* would be fine to, I can go ahead and trigger one and we can give it some testing just to make sure it works fine, but it looks like it would.

Issue #4228, opened 2024-11-14
"Undefined array key "type"


ozgurkazancci:
Hello,

I'm constantly having the following warning message in NC log files;

"Undefined array key "type" at /var/www/nextcloud/apps/richdocuments/lib/Service/TemplateFieldService.php#96"

PHP Version: 8.2.25
Nextcloud version: 30.0.2 - all apps up-to-date

FYI,
Regards.

Özgür Konstantin

elzody:
Hi, could you please share with us any more information? Such as what you do to trigger the error, or what kinds of errors you get in the browser / network request.

juliusknorr:
Please also share the raw json trace of the log line as that will give more hints on how this occurred. :)

ozgurkazancci:
@juliusknorr 

Here it is; https://paste.mozilla.org/85mRLMsV

Best wishes.

juliusknorr:
While stepping through the code for another issue i noticed that this seems caused by the document properties that are now also extracted:

![Image](https://github.com/user-attachments/assets/337688b5-bad4-40bf-bdfe-67b6064c34e3)

PR #4253, opened 2024-11-22
Fix errors when extracting content controls

juliusknorr:
- **fix: properly filter for content controls**
  - This commit fixes the filtering that has been in place but was likely not working anymore due to an upstream change 
- **fix: Catch unexpected results with no type**
  - Safeguard the results to not fail hard on unexpected result sets

Fix https://github.com/nextcloud/richdocuments/issues/4228

### DIFF
--- a/lib/Service/TemplateFieldService.php
+++ b/lib/Service/TemplateFieldService.php
@@ -94,7 +94,7 @@ class TemplateFieldService {
 			$fields = [];
 
 			foreach ($documentStructure as $index => $attr) {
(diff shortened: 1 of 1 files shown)


elzody:
Works fine for me 👍🏻

elzody (approved):

elzody:
Not merging right away because I wasn't sure to which versions you want to backport (I assume back to stable28?)

juliusknorr:
I think we only have that api integration in 30+

Issue #4229, opened 2024-11-14
Guest users are unable to save documents


elzody:
**Describe the bug**
When the [guests](https://github.com/nextcloud/guests) app is enabled, a guest user is unable to edit and save an office document which has been shared to them. This is due to the default quota for guests being set to `0`.

**To Reproduce**
1. Download and enable the [guests](https://github.com/nextcloud/guests) app
2. Give guest users permission to use the richdocuments app
3. Create a guest user
`OC_PASS=somepassword php occ guests:add --password-from-env --display-name "guest user" <created-by> guest@example.com`
5. Create an office document (in my case a spreadsheet) and share it to the newly created guest user
6. Open a new browser window and login as the guest user (the username is the email set for the user)
7. Access the document shared to the guest user and make a few edits
8. Try to save the document and see the error popups

**Expected behavior**
The expected behavior is that the share owner's quota is used rather than the guest user's quota. It makes sense for the guest quota to default to `0` for blocking file uploading, but for document editing it makes more sense to handle it differently. If you're sharing a document to a guest with editing permissions, you want them to be able to actually edit and save the document.

⚠️ **Reproducible from master to stable28**

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
{
  "reqId": "ZDlS4hWmEp9X24wRvBjJ",
  "level": 3,
  "time": "2024-11-14T15:45:59+00:00",
  "remoteAddr": "172.20.0.1",
  "user": "--",
  "app": "richdocuments",
  "method": "POST",
  "url": "/index.php/apps/richdocuments/wopi/files/56_ocnnlnfpgb7y/contents?access_token=yUV9MvHSRPUTFhOpnONRNUHMVsi1dlUF&access_token_ttl=0",
  "message": "Not enough storage",
  "userAgent": "COOLWSD HTTP Agent 24.04.9.2",
  "version": "28.0.11.1",
  "exception": {
    "Exception": "Exception",
    "Message": "Not enough storage",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/html/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 230,
        "function": "putFile",
        "class": "OCA\\Richdocuments\\Controller\\WopiController",
        "type": "->",
        "args": [
          "56",
          "yUV9MvHSRPUTFhOpnONRNUHMVsi1dlUF"
        ]
      },
      {
[...]

PR #4301, opened 2024-12-02
fix: use storage free space

elzody:
* Resolves: #4229 
* Target version: main

### Summary
Using `getStorage()` will allow us to get the storage of the share owner, which means guest users (or other users with zero quota) will still be able to edit documents which are shared to them. The previous method of using `getParent()->getFreeSpace()` caused it to retrieve the editing user's free storage space instead, which results in the unintended behavior described by the issue.

### TODO

- [ ] Write tests

### Checklist

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -444,7 +444,7 @@ class WopiController extends Controller {
 
 			$content = fopen('php://input', 'rb');
 
-			$freespace = $file->getParent()->getFreeSpace();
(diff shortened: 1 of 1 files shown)


juliusknorr (approved):

elzody:
Also need to check for backports here, will do at some point soon.

juliusknorr:
Adding the label so we don't forget (I tend to check all issues labeled with that from time to time and usually that is a good idea to do before a new release)

Issue #4311, opened 2024-12-06
Electronically sign documents via eIDEasy


vmiklos:
**Is your feature request related to a problem? Please describe.**

This is related to #4123, which was about digital signing via software certificates.

**Describe the solution you'd like**

Collabora Online has support to use eIDEasy to sign PDF files without sending the actual PDF file to an external service, see https://docs.eideasy.com/electronic-signatures/api-flow-with-file-hashes-pdf.html for the details. The request here is to add the related settings (API URL, client id, secret) to the richdocuments admin settings, so Nextcloud users can use this feature.

Then these should be sent to COOL via the WOPI CheckFileInfo, so COOL can use the eIDEasy API as required.

In the future, perhaps it should be possible to restrict the eIDEasy usage to a specific group -- but it's just a possibility, nobody requested that so far.

**Describe alternatives you've considered**

There is already a dedicated Nextcloud app, which can also work with eIDEasy, see <https://github.com/eideasy/nextcloud-electronic-signatures-plugin>. One benefit of the request approach is that that when SecureView is enabled, then you can sign PDF files without giving the PDF to the user. Another benefit is that hash-based signing is possible with the Nextcloud all-in-one docker image, while the mentioned app would require a separate docker image for hash-based signing. (Previously it seemed hash-based signing is not possible at all with that app, but this is not true, it just requires a separate docker container to handle the PDF manipulation that is built into Collabora Online already.)

**Additional context**

https://github.com/CollaboraOnline/online/issues/10630 is the Collabora Online side of this, there you can find some test API URL / client id / secret to try out this feature.

The Collabora Online side still has a few rough edges (e.g. the signing popup is not yet localized), but it's probably at a level where it makes sense to expose this feature on the settings UI.

I intend to work on this, just creating the issue to track the richdocuments progress at a single place.

juliusknorr:
> I intend to work on this, just creating the issue to track the richdocuments progress at a single place.

Let us know if you need anything from our side.

vmiklos:
The above PR adds this section to the admin settings:

![Image](https://github.com/user-attachments/assets/8dbcac60-7af0-4a37-9bb7-c58967a269cd)

CI has some failures, but seems that's unrelated (3 checks are broken on the baseline already, and the static analysis finds 2 places, but those lines are not touched in the PR), it seems to me.

PR #4328, opened 2024-12-10
feat: electronic signing, add settings for eIDEasy

vmiklos:
* Resolves:  #4311
* Target version: main

### Summary

Electronic signing needs to store settings as richdocuments settings. This involves the API URL, a client ID visible to the browser and a secret, which is only used during server-side requests.

The WOPI CheckFileInfo reply sends this information to the COOL server, similar to how it's done for digital signing (via PEM files).

Add the settings as admin settings, otherwise normal users would be able to use eIDEasy services outside richdocuments.

<https://github.com/CollaboraOnline/online/issues/10630#issuecomment-2510923379> has instructions on what test data to use to try out the service in a test environment. Additionally, if the test CA is configured to be trusted as a user setting, then the green stamp icon will show up in the status bar.

### Checklist

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -159,6 +159,7 @@ class WopiController extends Controller {
 			'IsUserLocked' => $this->permissionManager->userIsFeatureLocked($wopi->getEditorUid()),
 			'EnableRemoteLinkPicker' => (bool)$wopi->getCanwrite() && !$isPublic && !$wopi->getDirect(),
 			'HasContentRange' => true,
+			'ServerPrivateInfo' => [],
 		];
 
 		$enableZotero = $this->config->getAppValue(Application::APPNAME, 'zoteroEnabled', 'yes') === 'yes';
@@ -174,6 +175,23 @@ class WopiController extends Controller {
 			$response['UserPrivateInfo']['SignatureKey'] = $documentSigningKey;
 			$documentSigningCa = $this->config->getUserValue($wopi->getEditorUid(), 'richdocuments', 'documentSigningCa', '');
 			$response['UserPrivateInfo']['SignatureCa'] = $documentSigningCa;
+
+			$eSignatureBaseUrl = $this->config->getAppValue(Application::APPNAME, 'esignature_base_url');
+			$eSignatureClientId = $this->config->getAppValue(Application::APPNAME, 'esignature_client_id');
+			$eSignatureSecret = $this->config->getAppValue(Application::APPNAME, 'esignature_secret');
+			if ($eSignatureBaseUrl === '' && $eSignatureClientId !== '' && $eSignatureSecret !== '') {
+				// If the client ID & secret is set, then assume a production base URL.
+				$eSignatureBaseUrl = 'https://id.eideasy.com';
+			}
+			if ($eSignatureBaseUrl !== '') {
+				$response['ServerPrivateInfo']['ESignatureBaseUrl'] = $eSignatureBaseUrl;
+			}
+			if ($eSignatureClientId !== '') {
+				$response['ServerPrivateInfo']['ESignatureClientId'] = $eSignatureClientId;
+			}
+			if ($eSignatureSecret !== '') {
+				$response['ServerPrivateInfo']['ESignatureSecret'] = $eSignatureSecret;
+			}
 		}
 		if ($wopi->hasTemplateId()) {
 			$response['TemplateSource'] = $this->getWopiUrlForTemplate($wopi);
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -391,6 +391,20 @@
 			</div>
 		</div>
 
+		<div v-if="isSetup" id="esignature-settings" class="section">
+			<h2>{{ t('richdocuments', 'Electronic signature settings') }}</h2>
+			<SettingsInputText v-model="settings.esignature_client_id"
+				:label="t('richdocuments', 'Client ID for the electronic signature API')"
+				:hint="t('richdocuments', 'Fill in the registration form at https://eideasy.com/signup to obtain a client ID and secret.')"
+				:disabled="updating"
+				@update="updateESignatureClientId" />
+			<SettingsInputText v-model="settings.esignature_secret"
+				:label="t('richdocuments', 'Secret for the electronic signature API')"
+				:hint="t('richdocuments', 'The secret may be downloadable via WOPI requests if WOPI allow list is not correctly configured.')"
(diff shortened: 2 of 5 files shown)


vmiklos:
@juliusknorr could you please review this when you have time?

How to test, if you want: take the eideasy test url/clientid/secret from https://github.com/CollaboraOnline/online/issues/10630#issuecomment-2510923379, go to the richdocuments admin settings and configure the 3 values. Then open a PDF in COOL, Insert menu -> add electronic signature, you'll need to decide what provider to use, accept the default.

Then you need a test personal ID, use the value from the above GH comment. Once the popup is closed, you will notice a signature green sign / warning sign in the statusbar (depending on if the test CA is manually trusted), you can view the signature there.

https://github.com/nextcloud/richdocuments/issues/4311#issuecomment-2530723847 has a screenshot on how this looks like on my end currently. Thanks.

vmiklos:
Oh and forgot to mention: it seems to me that none of the 4 failing checks are related to the changes in this PR, but please let me know if I overlooked something. Thanks.

vmiklos:
One more small thing: the properties used here are now documented at https://sdk.collaboraonline.com/docs/advanced_integration.html#electronic-signature-handling, so it's a bit safer to depend on them here.

juliusknorr:
Yes, CI failures are unrelated 👍

lib/Controller/WopiController.php
```diff
@@ -174,6 +174,19 @@ public function checkFileInfo(string $fileId, string $access_token): JSONRespons
...
 			$documentSigningCa = $this->config->getUserValue($wopi->getEditorUid(), 'richdocuments', 'documentSigningCa', '');
 			$response['UserPrivateInfo']['SignatureCa'] = $documentSigningCa;
+
+			$eSignatureBaseUrl = $this->config->getAppValue(Application::APPNAME, 'esignature_base_url');
+			if ($eSignatureBaseUrl !== '') {
+				$response['UserPrivateInfo']['ESignatureBaseUrl'] = $eSignatureBaseUrl;
+			}
+			$eSignatureClientId = $this->config->getAppValue(Application::APPNAME, 'esignature_client_id');
+			if ($eSignatureClientId !== '') {
+				$response['UserPrivateInfo']['ESignatureClientId'] = $eSignatureClientId;
```
juliusknorr, line 184: Confuses me a bit that this is in UserPrivateInfo, while it is not specific to the user and rather an instance wide setting.

src/components/AdminSettings.vue
```diff
@@ -391,6 +391,25 @@
 			</div>
 		</div>
 
+		<div v-if="isSetup" id="esignature-settings" class="section">
+			<h2>{{ t('richdocuments', 'Electronic signature settings') }}</h2>
+			<SettingsInputText v-model="settings.esignature_base_url"
+				:label="t('richdocuments', 'URL for the electronic signature API')"
+				:hint="t('richdocuments', 'The production API URL is https://id.eideasy.com, the test API URL is https://test.eideasy.com.')"
```
juliusknorr, line 398: Not a blocker, just curious, apart from development is there any reason to configure the test API URL in productive usage? If not I'd remove the input for this one and rather document how to set this value with the CLI for developers.

src/components/AdminSettings.vue
```diff
@@ -391,6 +391,25 @@
...
+				:disabled="updating"
+				@update="updateESignatureBaseUrl" />
+			<SettingsInputText v-model="settings.esignature_client_id"
+				:label="t('richdocuments', 'Client ID for the electronic signature API')"
+				:hint="t('richdocuments', 'Fill in the registration form at https://eideasy.com/signup to obtain a client ID and secret.')"
+				:disabled="updating"
+				@update="updateESignatureClientId" />
+			<SettingsInputText v-model="settings.esignature_secret"
+				:label="t('richdocuments', 'Secret for the electronic signature API')"
+				:hint="t('richdocuments', 'The secret may be downloadable via WOPI requests if WOPI settings are not correctly configured.')"
```
juliusknorr, line 408: Maybe this to be more specific:

```suggestion
				:hint="t('richdocuments', 'The secret may be downloadable via WOPI requests if WOPI allow list is not correctly configured.')"
```

juliusknorr:
Left some quick comments, but generally looks fine from my side. 👍 

Have not tested yet.

lib/Controller/WopiController.php
vmiklos: This is probably because of my limited experience with WOPI: do you suggest to move this to a toplevel field in the response or something else?

The eid secret is something that the cool server should have access to, but we intentionally don't push it to the client.

Thanks.

src/components/AdminSettings.vue
vmiklos: Yes, this is basically for development only, you're right. Where to document this, is docs/app_settings.md a good place to mention it?

juliusknorr: Yes, that sounds good, there might be more undocumented ones, but that is the best place from my perspective 👍

lib/Controller/WopiController.php
juliusknorr: Yes, top level or nested like this should be fine I think, but maybe worth to check if there are more strong opinions from others working on WOPI on your side.

vmiklos: I would tend to go with a ServerPrivateInfo toplevel field to contain not just esign properties, but anything where it's important to configure the WOPI allowlist. As you suggest, I'll check with others on the Collabora side and get back to you on this.

juliusknorr: Sounds good 👍

vmiklos:
Just pushing what I have so far, I still need to address the comment about making the API URL cmdline-only + moving to ServerPrivateInfo.

src/components/AdminSettings.vue
vmiklos: Done.

lib/Controller/WopiController.php
vmiklos: https://github.com/CollaboraOnline/online/pull/10734 now changes the COOL side so we'll expect this proposed ServerPrivateInfo top-level key from you for esign settings. This can host any other instance wide credentials in the future.

src/components/AdminSettings.vue
vmiklos: OK, let me see, I think by now this is the only outstanding part of this PR.

vmiklos:
@juliusknorr please take a look, I think I addressed your comments so far. Thanks!

elzody (approved):

Issue #4370, opened 2025-01-06
Collaborative Editing - Renaming file makes the removes the file from the folder


pedropintosilva:
This issue was originally reported by @Ezinnem 

-----

This bug occurs in Richdocuments 8.5.3 version in both single user and Collaborative users in a Shared folder

Seen during collaboratve editing on 2024-12-24 using COOLWSD version: 24.04.11.1(git hash: 3d3b1895f7 (E)) LOKit version: Collabora Office 24.04.11.1(git hash: 4459677)

A simpler step:

Open a folder using a shared URL
Open a file in that folder
Rename the file

See error:

![image](https://github.com/user-attachments/assets/c54f53f5-7445-457b-ba05-57e19f622315)

If the file is reloaded, the file will be missing the folder

[Screencast_from_24-12-24_15_54_17.webm](https://github.com/user-attachments/assets/2e54d356-17c9-4271-a7dd-196c6e2751a3)

jazevedo-coll:
A bisect of `richdocuments` indicates that this is the commit from which this behavior started happening; 

```
565f825bf613cd49de5c42c6ed777baa4321c612 is the first bad commit
commit 565f825bf613cd49de5c42c6ed777baa4321c612
Author: Elizabeth Danzberger <lizzy7128@tutanota.de>
Date:   Wed Nov 13 12:04:51 2024 -0500

    fix: do not treat guest as share owner
    
    Signed-off-by: Elizabeth Danzberger <lizzy7128@tutanota.de>

 lib/Controller/DocumentController.php    | 19 +++++++++++----
 lib/Controller/WopiController.php        | 40 ++++++++++++--------------------
 lib/TokenManager.php                     | 39 ++++++++++++++++++++++++-------
 tests/features/bootstrap/WopiContext.php | 30 ++++++++++++++++++++++++
 tests/features/wopi.feature              | 16 +++++++++++++
 5 files changed, 107 insertions(+), 37 deletions(-)
```

I checked the previous commit. And it was ok. 

https://github.com/nextcloud/richdocuments/commit/565f825bf613cd49de5c42c6ed777baa4321c612

From these merge requests: 

https://github.com/nextcloud/richdocuments/pull/4306
Backport: https://github.com/nextcloud/richdocuments/pull/4310

Which seems to be to address this issue: https://github.com/nextcloud/richdocuments/issues/3830

pedropintosilva:
cc: @elzody

elzody:
[...]

PR #4378, opened 2025-01-09
fix(sharing): use correct save path when renaming

elzody:
* Resolves: #4370 
* Target version: main

### Summary
When accessing a publicly shared folder as a logged-in user with edit permissions and attempting to rename a file from within Collabora, the file would seemingly disappear upon a page refresh. It was actually being saved to the owner's root folder (so not entirely disappearing, just being moved out of the shared folder).

It seems to have been caused by a previous pull request which introduced logic to determine the proper save path for a file. We instead just use the file path now if it is a request for a name change, instead of doing some unneeded checks for public shares, etc.

### Checklist


juliusknorr:
Looks good, thanks for also extending the test cases 👍

PR #4386, opened 2025-01-14
fix: Generate preview with proper aspect ratio for templates

juliusknorr:
Steps to reproduce:

- Select a user template folder in richdocuments settings
- Add a template
- Open the template picker

Before this change we injected a square preview that looked odd in the tempalte picker with others coming from the server template directory having the aspect ratio of the document.


elzody:
looks like there is a merge conflict on `composer.lock`, but correct me if i am wrong we can just disregard it and push the new one instead, it's probably because of the dependency update or something?

juliusknorr:
Yes, I manually bumped the package, but was done separately on main meanwhile

juliusknorr:
Rebased

elzody (approved):

PR #4388, opened 2025-01-16
Refactor: Move personal user settings page to Vue

codewithvk:
* Resolves: # 
* Target version: main

### Checklist

### DIFF
--- a/src/components/DocSigningField.vue
+++ b/src/components/DocSigningField.vue
@@ -1,0 +1,98 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<div class="doc-signing-field">
+		<NcTextArea :value="localValue"
+			:label="label"
+			:disabled="disabled"
+			:placeholder="placeholder"
+			:helper-text="helperText"
+			@update:value="handleUpdate" />
+		<div class="doc-signing-actions">
+			<NcButton type="secondary" @click="onClickSave">
+				{{ t('richdocuments', 'Save') }}
+			</NcButton>
+			<NcButton type="secondary" :title="t('richdocuments', 'Remove')" @click="onClickRemove">
+				<DeleteIcon :size="20" />
+			</NcButton>
+		</div>
+	</div>
+</template>
+
+<script>
+import NcTextArea from '@nextcloud/vue/dist/Components/NcTextArea.js'
+import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
+import DeleteIcon from 'vue-material-design-icons/Delete.vue'
+
+export default {
+	name: 'DocSigningField',
+	components: {
+		NcTextArea,
+		NcButton,
+		DeleteIcon,
+	},
+	props: {
+		label: {
+			type: String,
+			required: true,
+		},
+		value: {
+			type: String,
+			default: '',
+		},
+		disabled: {
+			type: Boolean,
+			default: false,
+		},
+		placeholder: {
+			type: String,
+			default: '',
+		},
+		helperText: {
+			type: String,
+			default: '',
+		},
+	},
+	data() {
+		return {
+			localValue: this.value,
+		}
+	},
+	watch: {
+		value(newVal) {
+			this.localValue = newVal
+		},
+	},
+	methods: {
+		t(scope, text) {
+			return window.t ? window.t(scope, text) : text
+		},
+		handleUpdate(val) {
+			this.localValue = val
+			this.$emit('update:value', val)
+		},
+		onClickSave() {
+			this.$emit('save', this.localValue)
+		},
+		onClickRemove() {
+			this.$emit('remove')
+		},
+	},
+}
+</script>
+
+<style scoped>
+.doc-signing-field {
+	display: flex;
+	flex-direction: column;
+}
+
+.doc-signing-actions {
+	display: flex;
+	align-items: center;
+	gap: 1rem;
+}
+</style>
--- a/src/components/PersonalSettings.vue
+++ b/src/components/PersonalSettings.vue
@@ -1,0 +1,271 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<NcSettingsSection :name="t('richdocuments', 'Nextcloud Office')"
+		:description="t('richdocuments', 'Personal Settings for Nextcloud Office')"
+		:limit-width="true">
+		<!-- Template folder selection -->
+		<div class="template-folder-settings">
+			<div class="template-input-wrapper">
+				<NcTextField v-model="templateFolder"
+					:label="t('richdocuments', 'Select a template directory')"
+					:disabled="true" />
+			</div>
+			<NcButton id="templateSelectButton"
+				type="secondary"
+				@click="onTemplateSelectButtonClick">
+				<span class="icon-folder"
+					:title="t('richdocuments', 'Select a personal template folder')"
+					data-toggle="tooltip" />
+			</NcButton>
+			<NcButton id="templateResetButton"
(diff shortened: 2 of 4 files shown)


src/components/DocSigningField.vue
```diff
@@ -0,0 +1,78 @@
...
+<template>
+	<p>
+		<label :for="id">
+			{{ label }}
+		</label>
+		<br>
+		<textarea :id="id"
+			v-model="localValue"
+			:name="name" />
+		<br>
```
juliusknorr, line 15: You can use [NcTextArea](https://nextcloud-vue-components.netlify.app/#/Components/NcFields?id=nctextarea) which comes with proper styling for the label

src/components/DocSigningField.vue
```diff
@@ -0,0 +1,78 @@
...
+		<button :id="id + 'Save'" @click="onClickSave">
+			<span :title="saveTooltip" data-toggle="tooltip">
+				Save
+			</span>
+		</button>
```
juliusknorr, line 20: Can you switch that to use the [NcButton](https://nextcloud-vue-components.netlify.app/#/Components/NcButton) component? We can drop the tooltip as we have a title on the button, use the `t()` function to make text translatable
```suggestion
		<NcButton @click="onClickSave" type="secondary">
			{{ t('richdocuments', 'Save') }}
		</NcButton>
```

src/components/DocSigningField.vue
```diff
@@ -0,0 +1,78 @@
...
+
+		<button :id="id + 'Remove'" @click="onClickRemove">
```
juliusknorr, line 22: Same as above

src/components/PersonalSettings.vue
```diff
@@ -0,0 +1,264 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<div id="richdocuments" class="section">
```
juliusknorr, line 7: Can use [NcSettingsSection](https://nextcloud-vue-components.netlify.app/#/Components/NcSettings?id=ncsettingssection)

src/components/PersonalSettings.vue
```diff
@@ -0,0 +1,264 @@
...
+						<span :title="t('richdocuments', 'Save Zotero API key')" data-toggle="tooltip">
+							Save
+						</span>
+					</button>
+					<button id="zoteroAPIKeyRemove"
+						@click="resetZoteroAPI">
+						<span class="icon-delete"
+							:title="t('richdocuments', 'Remove Zotero API Key')"
+							data-toggle="tooltip" />
+					</button>
```
juliusknorr, line 66: Please also use NcButton and [NcTextField](https://nextcloud-vue-components.netlify.app/#/Components/NcFields?id=nctextfield) as components

src/components/PersonalSettings.vue
```diff
@@ -0,0 +1,264 @@
...
+				<span class="icon-folder"
+					:title="t('richdocuments', 'Select a personal template folder')"
+					data-toggle="tooltip" />
+			</button>
+			<button id="templateResetButton"
+				@click="resetTemplate">
+				<span class="icon-delete"
+					:title="t('richdocuments', 'Remove personal template folder')"
+					data-toggle="tooltip" />
+			</button>
```
juliusknorr, line 35: Please also use NcButton and [NcTextField](https://nextcloud-vue-components.netlify.app/#/Components/NcFields?id=nctextfield) as components

templates/personal.php
```diff
@@ -1,57 +1,4 @@
 <?php
-/**
- * SPDX-FileCopyrightText: 2016-2024 Nextcloud GmbH and Nextcloud contributors
- * SPDX-FileCopyrightText: 2013-2016 ownCloud, Inc.
- * SPDX-License-Identifier: AGPL-3.0-or-later
- */
-script('richdocuments', 'richdocuments-personal');
+script('richdocuments', 'richdocuments-personal'); // This should load your personal.js bundle
```
elzody, line 2: While we are modernizing `personal.php` a bit, it might be nice if we can change this to use the more modern way to add a script: 

```suggestion
\OCP\Util::addScript('richdocuments', 'richdocuments-personal'); // This should load your personal.js bundle
```

src/personal.js
```diff
@@ -1,173 +1,25 @@
-/**
- * SPDX-FileCopyrightText: 2019 Nextcloud GmbH and Nextcloud contributors
- * SPDX-License-Identifier: AGPL-3.0-or-later
- */
```
elzody, line 4: We need to keep the SPDX header here.

src/components/PersonalSettings.vue
```diff
@@ -0,0 +1,303 @@
...
+					if (val === '') {
+			  showSuccess(this.t('richdocuments', 'Document signing CA chain removed'))
+					} else {
+			  showSuccess(this.t('richdocuments', 'Document signing CA chain saved'))
+					}
+		  },
+		  () => { showError(this.t('richdocuments', 'Failed to update the document signing CA chain')) },
+			)
+	  },
+	  updateSetting(data, successCallback, errorCallback) {
```
elzody, line 236: I'm wondering if there is a better way to handle this network request. It might be worth it to check out [@nextcloud/axios](https://github.com/nextcloud-libraries/nextcloud-axios), which will take care of setting the required request headers automatically, and provides a nice way to formulate the request in general. You can find examples of how it is used in other areas of the richdocuments codebase (e.g. the admin settings component).

It also might be worth it to make this an async method, and instead of having the `successCallback` and `errorCallback`, you could return something as simple as a boolean which reflects whether or not the request succeeded. This can then be used in the other individual settings methods so they can react how they see fit. I personally find the way of passing success and error callbacks in each of these confusing and a bit difficult to read.

src/components/PersonalSettings.vue
```diff
@@ -0,0 +1,303 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<NcSettingsSection :name="t('richdocuments', 'Nextcloud Office')"
+		:description="t('richdocuments', 'Personal Settings for Nextcloud Office')"
+		:limit-width="true">
+		<span id="documents-admin-msg" class="msg" />
```
elzody, line 10: I would say we could get rid of this and not use the `OC.msg` thing and we can instead just stick with the `showError` and `showSuccess` toasts we have in [@nextcloud/dialogs](https://github.com/nextcloud-libraries/nextcloud-dialogs)

elzody (changes requested):
Sorry if I was being a bit picky here, but I made a few comments and suggestions. If you need any help don't hesitate to ask

src/components/DocSigningField.vue
codewithvk: Done

src/components/DocSigningField.vue
codewithvk: Done

src/components/DocSigningField.vue
codewithvk: Done

src/components/PersonalSettings.vue
codewithvk: Done

src/components/PersonalSettings.vue
codewithvk: Done

templates/personal.php
codewithvk: Done

src/personal.js
codewithvk: Done

src/components/PersonalSettings.vue
codewithvk: Done

src/components/PersonalSettings.vue
codewithvk: I have come with better approach, Could you please check it now?

elzody (approved):
Looks good. The only thing left before merging is to fix the failing test, which you can likely do by just running `composer run cs:fix` and committing the changes.

Issue #4095, opened 2024-10-02
Document generation API


julien-nc:
Hey!

Would it be possible to extend the template filling API to "dynamically" generate documents?
The use case we have in the integrations team is more or less generating a presentation from a prompt.
We thought that an LLM could generate a text based, easy to parse, description of a set of slides. We would then need to generate a document from this description.

One way to approach this could be to have an endpoint to fill multiple templates and assemble them into one document.

The input for this endpoint could be like:
``` json
{
    "type": "presentation",
    "parts": [
        {
            "templatePath": "/templates/title.odp",
            "values": { "1": "big title", "2": "subtitle" },
        },
        {
            "templatePath": "/templates/bulletpoints.odp",
            "values": { "1": "slide title", "2": "* bullet one\n*bullet 2" },
        },

    ],
}
```

Wdyt @juliushaertl @elzody ?
cc @marcelklehr

aszucs3:
Just a few things i found while i was checkd it a bit:
-it will need some extra work to make extract/transform work on impress documents as well... right now its source code is in writer.. but surely it is planned that we should extend it to impress/ calc anyway ..
-to open multiple document, and generate 1 more document based on it... seems to be a challenge now.. but i have some ideas..

The easiest idea i have: 
we could use extract on multiple documents, and then use the extracted data to to a transform to make a new document.. 
pro: we would not need to code multiple document opening into the system .. 
con1: im not sure if we can extract enought information to make the same slide as it was before..
con2: maybe it would need to run separate extract commands, then run a transfer with the extracted data... BUT maybe we could automate it... maybe it could be coded into the same code. ... it could generate the right extratc / transfrom .. and then run it on the server, so maybe the user would not see anything from it? :)

the other way: 
we could code into the server side to open all the documents separetelly ... and somehow copy/paste slides between them...
pro: it sure could copy all needed data.. 
con: it needs a different separete code into server side i never tried to... so im not sure what challenges i will meet there... but if i start to do it, i can see it .. :)

Later when i will have time, i will check it a bit more deep.. and maybe i will be able to give better ideas.. :)

juliusknorr:
>  -to open multiple document, and generate 1 more document based on it... seems to be a challenge now.. extract/transform, open only 1 document .. and may generate 1 new from it .. but i have some idea to handle it.

[...]

PR #4390, opened 2025-01-16
Slide deck generation task provider

julien-nc:
refs #4095

Here is what was originally in https://github.com/nextcloud/slide_deck_generator

- [x] Needs mapping from AI generated JSON to collabora API @elzody 
- [x] Needs a comment somewhere about the nesting of tasks that may need multiple workers

### DIFF
--- a/lib/Service/SlideDeckService.php
+++ b/lib/Service/SlideDeckService.php
@@ -1,0 +1,156 @@
+<?php
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Richdocuments\Service;
+
+use OCA\Richdocuments\AppInfo\Application;
+use OCA\Richdocuments\TemplateManager;
+use OCP\IConfig;
+use OCP\TaskProcessing\Exception\Exception;
+use OCP\TaskProcessing\Exception\NotFoundException;
+use OCP\TaskProcessing\Exception\PreConditionNotMetException;
+use OCP\TaskProcessing\Exception\UnauthorizedException;
+use OCP\TaskProcessing\Exception\ValidationException;
+use OCP\TaskProcessing\IManager;
+use OCP\TaskProcessing\Task;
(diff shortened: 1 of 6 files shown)


lib/Service/SlideDeckService.php
```diff
@@ -0,0 +1,72 @@
...
+use OCP\TaskProcessing\Exception\UnauthorizedException;
+use OCP\TaskProcessing\Exception\ValidationException;
+use OCP\TaskProcessing\IManager;
+use OCP\TaskProcessing\Task;
+use OCP\TaskProcessing\TaskTypes\TextToText;
+use RuntimeException;
+
+class SlideDeckService {
+	public const PROMPT = <<<EOF
+Draft a presentation slide deck with headlines and a maximum of 5 bullet points per headline. Use the following JSON structure for your whole output and output only the JSON array, no introductory text:
```
juliusknorr, line 22: I just remembered, Daphne raised to me that this likely should support different languages as well (especially German). Is that feasible? Just using different prompts then?

julien-nc: We should probably add an instruction sentence to make sure the model outputs in the same language as the "presentation text" coming from the user.

The "system" prompt can stay in english. We do it for summary, headlines, topics etc... and it works pretty well with most tested models.

lib/Service/SlideDeckService.php
```diff
@@ -0,0 +1,156 @@
...
+		return [ 'SlideCommands' => $slideCommands ];
+	}
+
+	/**
+	 * Creates a blank presentation file in memory
+	 *
+	 * @param string $format
+	 * @return resource
+	 */
+	private function getBlankSpreadsheet(string $format) {
```
juliusknorr, line 110:
```suggestion
	private function getBlankPresentation(string $format) {
```

As a next step we should discuss how we pick or provide different template files with some nice designs

juliusknorr (approved):
Small nitpick, looks good otherwise 👍

lib/Service/SlideDeckService.php
elzody: Nice catch, I must have mixed it up

PR #4411, opened 2025-01-24
fix: translation format specifier

elzody:
* Target version: main

### Summary
Fixes an incorrect translation string format specifier.

### Checklist

### DIFF
--- a/lib/Conversion/ConversionProvider.php
+++ b/lib/Conversion/ConversionProvider.php
@@ -149,6 +149,14 @@ class ConversionProvider implements IConversionProvider {
 		$outputMimeInfo = $this->getMimeInfoFor($outputMimeType);
 		if ($outputMimeInfo === null) {
 			$this->logger->error(
+				/*
+				 * TRANSLATORS
+				 * Shows the MIME type of the file (the file type)
+				 * For example:
+				 *    - application/pdf     (.pdf)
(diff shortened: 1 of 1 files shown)


lib/Conversion/ConversionProvider.php
```diff
@@ -148,7 +148,10 @@ public function convertFile(File $file, string $targetMimeType): mixed {
 	private function getMimeProvidersFor(array $inputMimeTypes, string $outputMimeType): array {
 		$outputMimeInfo = $this->getMimeInfoFor($outputMimeType);
 		if ($outputMimeInfo === null) {
-			$this->logger->error($this->l10n->t('Unable to fetch information on $s', [$outputMimeType]));
+			$this->logger->error(
+				$this->l10n->t('Unable to fetch information on %1$s',
```
juliusknorr, line 152: @rakekniven Should we provide context with `// TRANSLATORS`comments here on that it is a mimetype in the placeholder?

(ref https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#provide-context-hints-for-translators)

I was quite surprised that we currently do not support naming placeholders in PHP like it is possible with javascript.

elzody: wait, if it is a logging error do we still even need to translate it in the first place?

Issue #4389, opened 2025-01-16
Increase guest_displayname to 255 characters


blizzz:
64 might be too short in context with federation

PR #4420, opened 2025-01-28
fix(Db): Allow guest displaynames with a length of up to 255 chars

blizzz:
* Resolves: #4389
* Target version: main

### Summary

Changes DB field for guest displaynames from 64 to 255 bytes. Useful in federated setups.

### Checklist


juliusknorr:
Meanwhile composer files were dropped

blizzz:
> Meanwhile composer files were dropped

:+1: Rebased, now it's only the migration files.

Issue #2092, opened 2022-03-10
Copying a file and renaming it in the same directory


pedropintosilva:
[to do: add description , test with other mime types]

https://user-images.githubusercontent.com/65948705/157645332-248332e4-9c2b-4799-b134-ffdf4c34acaa.mp4

```bash
11:39:44.206 XHRMOVEhttps://staging.eu.collaboraonline.com/nextcloud/remote.php/dav/files/pedro.silva/COOL Weekly/COOLWeekly63 (copy).odt
[HTTP/1.1 500 Internal Server Error 115ms]

	
MOVE
	https://staging.eu.collaboraonline.com/nextcloud/remote.php/dav/files/pedro.silva/COOL%20Weekly/COOLWeekly63%20(copy).odt
Status500
Internal Server Error
VersionHTTP/1.1
Transferred880 B (235 B size)
Referrer Policyno-referrer

    	
    Cache-Control
    	no-store, no-cache, must-revalidate
    Connection
    	close
    Content-Length
    	235
    Content-Security-Policy
    	default-src 'none';
    Content-Type
    	application/xml; charset=utf-8
    Date
    	Thu, 10 Mar 2022 10:39:44 GMT
    Expires
    	Thu, 19 Nov 1981 08:52:00 GMT
    Pragma
    	no-cache
    Referrer-Policy
    	no-referrer
    Server
    	Apache
    Strict-Transport-Security
    	max-age=15768000; includeSubDomains; preload
    X-Content-Type-Options
    	nosniff
    X-Download-Options
    	noopen
    X-Frame-Options
    	SAMEORIGIN
    X-Permitted-Cross-Domain-Policies
    	none
    X-Powered-By
    	PHP/7.4.6
    X-Robots-Tag
    	none
    X-XSS-Protection
    	1; mode=block
    	
    Accept
    	*/*
    Accept-Encoding
    	gzip, deflate, br
    Accept-Language
    	en-GB,en;q=0.5
    Cache-Control
    	no-cache
    Connection
    	keep-alive
    Cookie
[...]

PR #4425, opened 2025-01-29
fix: Read from local file stream to avoid locking during preview generation

juliusknorr:
Should resolve https://github.com/nextcloud/richdocuments/issues/2092

We can always use the local file, it may be temporary or the file itself in the data directory, but we can read without locking.


juliusknorr:
Managed to confirm by using a larger file that takes longer processing time and meanwhile upload a new version of the file.

elzody (approved):

Issue #1335, opened 2021-01-28
Make preview generation configurable


drikqlis:
**Describe the bug**
After installing Collabora app Nextcloud generates previews for Office and PDF files despite that it is not turned on in 'enabledPreviewProvieders' in nextcloud config.

**To Reproduce**
1. Do not turn on or disable in config.php Office and PDF preview generation.
2. Install Collabora app
3. Create new Office or PDF document - preview is generated.
4. Do the same with Collabora app turned off - preview is not generated as expected.

**Expected behavior**
It should not generate previews when not configured. According to Nextcloud documentation:

```
The following providers are disabled by default due to performance or privacy concerns:

        OC\Preview\Illustrator
        OC\Preview\Movie
        OC\Preview\MSOffice2003
        OC\Preview\MSOffice2007
        OC\Preview\MSOfficeDoc
        OC\Preview\OpenDocument
        OC\Preview\PDF
        OC\Preview\Photoshop
        OC\Preview\Postscript
        OC\Preview\StarOffice
        OC\Preview\SVG
        OC\Preview\TIFF
        OC\Preview\Font
```

**Screenshots**

**Client details:**
 - Device: [e.g. iPhone6, desktop]

## Server details

**Operating system**: Ubuntu 18.04 LTS

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.3

**Nextcloud version:** 20.0.6

**Version of the richdocuments app** 3.7.13

**Version of Collabora Online** 6.4

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

juliushaertl:
Yeah the app registered preview providers are added in addition to the ones provided in the config, I assume we could have a setting to turn preview generation off. In case you are running a real Collabora server you could just block the Nextcloud server from using the `<net><post_allow>` settings

cm-schl:
Hi,

[...]

PR #4426, opened 2025-01-29
Config option to disable previews

juliusknorr:
Implement #1335 and document our config flags


elzody (approved):

KsmoinO:
Did this pull request planned to be included in a future release ?
This changes are still absent from richdocuments 8.7.2

elzody:
@KsmoinO It was never back-ported, so it was never included in any release. We do all development against the `main` branch and back-port fixes that should be included in current releases. Odds are, if this was never back-ported, it was intended to be included in the next major release (Nextcloud 32).

I am not clear on the process yet but I will see if this can be back-ported and included in the next maintenance releases of the app, since it doesn't seem like a major feature although it is considered an "enhancement."

KsmoinO:
Thanks for your reply, it's more clear for me :)
Don't spend too much time on this as I have a workaround and I can wait for the next major release (even if I'd preferred don't have to patch after each updates :) )

PR #4461, opened 2025-02-09
fix(url): correct WOPI settings URL to prevent duplicate `nextcloud`

codewithvk:
Replaced `getCallbackBaseUrl()` with `window.location.protocol + '//' + window.location.host` to ensure the correct WOPI settings URL. This prevents the issue where `/nextcloud` appears twice in the generated URL.

prev URL : http://localhost/nextcloud/nextcloud/index.php/apps/richdocuments/wopi/settings
now it should be: http://localhost/nextcloud/index.php/apps/richdocuments/wopi/settings

### DIFF
--- a/src/helpers/url.js
+++ b/src/helpers/url.js
@@ -82,7 +82,7 @@ const getDocumentUrlForFile = (fileDir, fileId) => {
 }
 
 export const getConfigFileUrl = () => {
-	return getCallbackBaseUrl() + generateUrl('apps/richdocuments/wopi/settings')
+	return generateUrl('apps/richdocuments/wopi/settings', null, { baseURL: getCallbackBaseUrl() })
 }
 
 const getNextcloudUrl = () => {


src/helpers/url.js
```diff
@@ -82,7 +82,7 @@ const getDocumentUrlForFile = (fileDir, fileId) => {
 }
 
 export const getConfigFileUrl = () => {
-	return getCallbackBaseUrl() + generateUrl('apps/richdocuments/wopi/settings')
+	return window.location.protocol + '//' + window.location.host + generateUrl('apps/richdocuments/wopi/settings')
```
juliusknorr, line 85: Please use `getBaseUrl` from `@nextcloud/router`

codewithvk:
CI Irrelevant of this PR, @juliusknorr  @elzody Could you please review this PR?

src/helpers/url.js
codewithvk: BaseURL still same issue with some of the set up.

pedropintosilva:
Any news @elzody or @julien-nc ? This change seems quite small and I'm not sure what to do WRT failed unrelated checks

src/helpers/url.js
elzody: I am wondering why `getBaseUrl` isn't working. What do you mean exactly by "same issue with *some* of the set up?" I think directly using `window.location` probably is not the safest bet, so I am trying to understand what the issue is and how I can reproduce it to provide some better suggestions.

codewithvk: I have a Docker setup, and every approach works fine for me, including getCallbackBaseUrl(). However, for those who installed Nextcloud using the server installation method, the base URL appears twice. For example:

http://<someport>/nextcloud/nextcloud/index.php/apps/richdocuments/wopi/settings

The same issue occurs with both baseUrl and getCallbackBaseUrl – they generate a path like `http://<someport>/nextcloud/`, while generateUrl creates `/nextcloud/index.php/apps/richdocuments/wopi/settings`. When combined, this results in `/nextcloud` appearing twice, leading to an incorrect URL. This problem only occurs for those who installed or set up using the server installation method. The Docker setup works fine. Why this is happening, I don't know.

Do you have any solustion in mind?

elzody:
I think the issue with the duplication comes from the fact that `getCallbackBaseUrl()`, `getBaseUrl()`, and `generateUrl()` all also return the webroot in the path. For someone running Nextcloud at the normal webroot, with no subpath, I am pretty sure the generated path would return like one would expect (`https://my-nextcloud.com/index.php/apps/richdocuments/wopi/settings` or whatever).

But, my theory is that, when running in some subpath (e.g. `https//my-server.com/nextcloud`), all of these functions would also return the `/nextcloud` path in addition to their respective outputs. So, if you are testing this on such an environment, I can see why it might be the case. I have not tested it like that yet, so cannot confirm. Hopefully I am not just rambling about nonsense and my theory is correct.

So, one solution might be to use `generateUrl()` still, but provide a `baseUrl` to it so it won't automatically append the webroot:

```js
generateUrl('apps/richdocuments/wopi/settings', null, { baseUrl: getCallbackBaseUrl() })
```

I have not tried this, but if you could try that instead to see if it works in whatever setup you're using, it might be helpful.

@codewithvk @pedropintosilva

PR #4463, opened 2025-02-11
templates: better support cross-format templates

JaumePujantell:
If form filling was used to create a file, do not add a template source. Otherwise Collabora Online creates the file again from the template.

Use the format option in 'transform-document-structure' to avoid an intermediate save in the template format, which might not be supported as a save format.

### DIFF
--- a/lib/Listener/FileCreatedFromTemplateListener.php
+++ b/lib/Listener/FileCreatedFromTemplateListener.php
@@ -54,17 +54,17 @@ class FileCreatedFromTemplateListener implements IEventListener {
 			return;
 		}
 
-		if ($this->templateManager->isSupportedTemplateSource($templateFile->getExtension())) {
-			// Only use TemplateSource if supported filetype
-			$this->templateManager->setTemplateSource($event->getTarget()->getId(), $templateFile->getId());
-		}
-
 		if ($this->capabilitiesService->hasFormFilling()) {
 			try {
-				$filledTemplate = $this->templateFieldService->fillFields($templateFile, $event->getTemplateFields());
+				$filledTemplate = $this->templateFieldService->fillFields($templateFile, $event->getTemplateFields(), null, $event->getTarget()->getExtension());
 				$event->getTarget()->putContent($filledTemplate);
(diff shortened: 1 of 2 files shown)


lib/Service/TemplateFieldService.php
elzody, line 1: This part I understand and think is nice, we can skip writing to a temp file altogether since the transform document structure endpoint already can convert the document, it seems.

lib/Listener/FileCreatedFromTemplateListener.php
elzody, line 1: This stuff I don't really get, when I tried running it and creating a new file it doesn't quite work properly since the template service `fillFields` method expects a file to be handed to it; with these changes there isn't one. Can you explain a bit more the reasoning behind this or why it would be necessary? I don't completely understand

JaumePujantell: You are right, I shouldn't have moved the empty template code path, thanks.

lib/Listener/FileCreatedFromTemplateListener.php
```diff
@@ -54,18 +54,18 @@ public function handle(Event $event): void {
...
 
-		if ($this->templateManager->isSupportedTemplateSource($templateFile->getExtension())) {
-			// Only use TemplateSource if supported filetype
-			$this->templateManager->setTemplateSource($event->getTarget()->getId(), $templateFile->getId());
-		}
-
 		if ($this->capabilitiesService->hasFormFilling()) {
 			try {
-				$filledTemplate = $this->templateFieldService->fillFields($templateFile, $event->getTemplateFields());
+				$filledTemplate = $this->templateFieldService->fillFields($templateFile, $event->getTemplateFields(), null, event->getTarget()->getExtension());
```
elzody, line 59: This is missing a `$`
```suggestion
				$filledTemplate = $this->templateFieldService->fillFields($templateFile, $event->getTemplateFields(), null, $event->getTarget()->getExtension());
```

juliusknorr (approved):

elzody (approved):

PR #4464, opened 2025-02-11
fix: send etag for settings URLs and update etag on file changes

codewithvk:
* Resolves: # 
* Target version: main

### Summary
fix: send etag for settings URLs and update etag on file changes

- Replace timestamp with the folder etag in settings generation.
- Update the folder etag when files are created or deleted.
- Refactor folder retrieval logic into helper methods for consistency.

### Checklist


elzody (approved):
Don't think the Cypress failures are related. The `integration.spec.js` tests fail everywhere atm.

PR #4469, opened 2025-02-13
[stable27] fix: get storage from file instead of parent

elzody:
Manual backport of https://github.com/nextcloud/richdocuments/pull/4301


blizzz (approved):
The autoloader :scream_cat:

PR #4472, opened 2025-02-14
fix(tests): use more generic selectors

elzody:
Making CI green again required the following:

- updating some selectors
- new, more reliable method to verify open documents
- commenting out problematic code from a recent pull request

The code which was commented out breaks functionality when present in the code base, thus until the issue is resolved it has been commented so we can get CI green again and continue merging important fixes without haggling with Cypress.

### DIFF
--- a/cypress/e2e/integration.spec.js
+++ b/cypress/e2e/integration.spec.js
@@ -32,8 +32,11 @@ describe('Nextcloud integration', function() {
 
 	it('Sharing sidebar', function() {
 		cy.get('@loleafletframe').within(() => {
-			cy.get('#File-tab-label').click()
-			cy.get('#ShareAs-button').click()
+			cy.get('.notebookbar-tabs-container', { timeout: 30_000 })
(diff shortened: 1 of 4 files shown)


.github/workflows/cypress-e2e.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       matrix:
         code-image: [ 'release', 'nightly' ]
         node-version: [16.x]
-        containers: [1, 2, 3]
+        containers: [ 1 ]
```
juliusknorr, line 31: Would prefer switch back to 3 containers before merging

juliusknorr (approved):
Looks good, somewhat overlaps with https://github.com/nextcloud/richdocuments/pull/4520

.github/workflows/cypress-e2e.yml
elzody: Yep was only doing that for testing purposes, was more of a random change to see how things worked.

PR #4529, opened 2025-02-26
fix: Avoid getting the file for storing settings

juliusknorr:
Signed-off-by: Julius Knorr <jus@bitgrid.net>

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


lib/Middleware/WOPIMiddleware.php
```diff
@@ -60,6 +61,11 @@ public function beforeController($controller, $methodName) {
 			$accessToken = $this->request->getParam('access_token');
 			[$fileId, ,] = Helper::parseFileId($fileId);
 			$wopi = $this->wopiMapper->getWopiForToken($accessToken);
+
+			if ($wopi->getTokenType() === Wopi::TOKEN_TYPE_SETTING_AUTH) {
```
codewithvk, line 65: FYI: sometimes we can also send document tokens(type 0 in wopi) when we want to upload while the document is open.

Issue #4554, opened 2025-03-07
Database spammed with tokens


juliusknorr:
I noticed that opening a file just once leads to quite a ton of entries of tokens:

Related to 

<img width="1346" alt="Image" src="https://github.com/user-attachments/assets/04999c1b-dbae-4a3d-85f3-9d815c91625a" />

@codewithvk Can you tell why we have so many?

codewithvk:
@juliusknorr not sure, let me check for you

codewithvk:
@juliusknorr this fix could reduce list significants https://github.com/nextcloud/richdocuments/pull/4556

PR #4556, opened 2025-03-08
fix(token): reduce token generation for settings files

codewithvk:
* Resolves: #4554
* Target version: main

### Checklist


juliusknorr:
Nice, reusing the token is a good idea there 👍

juliusknorr (approved):

PR #4564, opened 2025-03-11
preset: fix: adding/removing global template not reflected in sidebar

Rash419:
- this patch appends the presentation template folder etag to shared setting etag so that COOL can detect changes
- this patch also update the etag of global templates folder when template is added or removed by admin

* Target version: main

### Checklist


lib/Controller/WopiController.php
```diff
@@ -996,7 +996,7 @@ private function generateSettingToken(string $userId): string {
 	private function generateSettings(string $userId, string $type): array {
 		$nextcloudUrl = $this->appConfig->getNextcloudUrl() ?: trim($this->urlGenerator->getAbsoluteURL(''), '/');
 		$uri = $nextcloudUrl . '/index.php/apps/richdocuments/wopi/settings' . '?type=' . $type . '&access_token=' . $this->generateSettingToken($userId) . '&fileId=' . '-1';
-		$etag = $this->settingsService->getFolderEtag($type);
+		$etag = $this->settingsService->getFolderEtag($type) . $this->settingsService->getPresentationFolderEtag($type, $userId);
```
codewithvk, line 999: Add a condition for guest users: if the user is a guest, then do not generate getPresentationFolderEtag.

juliusknorr: Actually I think we should change the method to also allow null for the userId, and not set it to a string in 

https://github.com/Rash419/richdocuments/blame/template-preset/lib/Controller/WopiController.php#L142

Using user ids that are non existing for non-logged in users may lead to more problems on other parts.

codewithvk: @juliusknorr make sense

codewithvk (approved):
LGTM. Thanks @Rash419 :)

I'll try to create a separate PR to resolve Julius's comment. https://github.com/nextcloud/richdocuments/pull/4564/files#r1988926909

juliusknorr (approved):

PR #4570, opened 2025-03-11
chore: update actions/cache

elzody:
Updates actions/cache workflow as it is now deprecated and throws errors in the workflow
https://github.com/actions/cache/discussions/1510

⚠️  **Needed so https://github.com/nextcloud/richdocuments/pull/4568 can pass**

### DIFF
--- a/.github/workflows/cypress-e2e.yml
+++ b/.github/workflows/cypress-e2e.yml
@@ -142,7 +142,7 @@ jobs:
           cat data/nextcloud.log
 
       - name: Cache node modules
-        uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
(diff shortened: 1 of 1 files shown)


.github/workflows/cypress-e2e.yml
```diff
@@ -142,7 +142,7 @@ jobs:
           cat data/nextcloud.log
 
       - name: Cache node modules
-        uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
```
blizzz, line 145: we keep the commit hashes, tags are not stable.

elzody: Understood, makes sense!

juliusknorr (approved):

PR #4606, opened 2025-03-19
Ensure we have a clean db schema when migrating

juliusknorr:
The content of the oc_richdocuments_wopi table is just temporary tokens, so in oder to ensure that the db schema matches if we install from old ownCloud versions, we can just drop and recreate the table.

This may be an issue because we did introduce new columns in the old database.xml file that was not properly covered in migration from the old project. 

Note the migration/doctrine abstractions will actually not drop but change the schema as one step.

Can be tested also by dropping some tables and running `occ migrations:execute richdocuments 2060Date20200302131958` manually.

### DIFF
--- a/lib/Migration/Version2060Date20200302131958.php
+++ b/lib/Migration/Version2060Date20200302131958.php
@@ -9,6 +9,7 @@ namespace OCA\Richdocuments\Migration;
 
 use Closure;
 use OCP\DB\ISchemaWrapper;
+use OCP\IDBConnection;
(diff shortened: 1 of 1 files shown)


blizzz (approved):

blizzz:
> will actually not drop but change the schema

I understand the change includes dropping the data.

juliusknorr:
Turns out it does not, I added a preSchemaChange step to delete existing rows.

Issue #4351, opened 2024-12-20
Settings design review


juliusknorr:
I think we should give the settings some love and do a design review, quick screenshots, but maybe something to schedule a call next year @marcoambrosini 

Overall it feel like we have some information overload there

## Personal settings

<img width="864" alt="Image" src="https://github.com/user-attachments/assets/1194f23b-9df6-4110-ad79-b574c3e0eccc" />

## Admin settings

<img width="896" alt="Image" src="https://github.com/user-attachments/assets/15753121-a597-46ee-8bb8-76d3ace43a74" />
<img width="893" alt="Image" src="https://github.com/user-attachments/assets/3a1b380f-e2f3-4cf9-a35a-35f0a18cf4b8" />
<img width="885" alt="Image" src="https://github.com/user-attachments/assets/36888eeb-9e4f-483b-b6b7-d3ae926be57e" />
<img width="885" alt="Image" src="https://github.com/user-attachments/assets/58255770-f1c2-4473-b924-2244510398d0" />
<img width="870" alt="Image" src="https://github.com/user-attachments/assets/56be0412-bd9d-4d3f-9e6f-b835850aedd4" />
<img width="882" alt="Image" src="https://github.com/user-attachments/assets/91fa4501-0eff-40c8-86ab-0083eca48a47" />

marcoambrosini:
Scheduled design review for 7/01

juliusknorr:
- "Server setup" should have a heading
- Maybe not use radio buttons but 3 tiles for initial setup and then it is there permanent 
  - Button change server setup to reset it
- Yellow warning wopi should not say "CLick ehre for more info" should be "See documentation on how to configure the wopi allow list"
- "Collabora Online server is reachable." should be bold, nothing else
- Important: New components should be used for form elements
- Make allow list for wopi requests also a checkbox
- Indent select input under checkboxes
- Non-critical actions should be allowed to change without an edit button 
  - URL and Wopi allow list should have separate edit / save buttons
- Extra fonts 
  - Should use new components
  - Call it "Custom fonts"
  - Should have a explainer at the start
  - Question to Collabora: Can we warn if remote_font_config is not set
  - Could be a warning at the top if not set
  - Upload custom font file can be a full button
- Secure view 
  - Remove "settings" from title
  - Move long explanation to the docs
- Watermark 
  - Maybe an example watermark text
- Electronic signature 
  - No "settings" in heading
- Global templates has different heading level 
  - Small description

---

[...]

PR #4621, opened 2025-03-24
feat: introduce pinia for admin settings

elzody:
* Related to: https://github.com/nextcloud/richdocuments/issues/4351
* Target version: main

### Summary
Introduces Pinia in an effort to organize the admin settings a bit instead of having everything in one massive component.

### TODO

- [ ] Tests

### Checklist


elzody:
@juliusknorr Do you think we should ignore the remote font config warning for now? Didn't we discuss that, because we can have these configuration files of our own now, we can always ensure there is one?

elzody:
@nextcloud/designers I have uploaded a screen recording, can I get a design review on this please? :)

src/components/AdminSettings/CustomFonts.vue
```diff
@@ -0,0 +1,199 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<NcSettingsSection :name="t('richdocuments', 'Custom Fonts')"
```
jancborchardt, line 7:
```suggestion
	<NcSettingsSection :name="t('richdocuments', 'Custom fonts')"
```

Detail: Should be sentence case, like below for "Available fonts"

jancborchardt (commented):
I'd say generally it's fine, but instead of the red trash button, a tertiary one (icon only) is enough?

Also @marcoambrosini @susnux what would be cool is a nice component for lists so this is consistent across Nextcloud. If we want to get fancy, this could be a grid/tiled view with font and font name below each other, sort of laid out like the tiles we use for the _settings_ of Dashboard.

src/components/AdminSettings/CustomFonts.vue
```diff
@@ -0,0 +1,199 @@
...
+		:description="description">
+		<input ref="newFontInput"
+			type="file"
+			class="hidden-visually"
+			accept=".ttf, .otf"
+			data-cy="newFontInput"
+			@change="selectFile">
+
+		<table>
```
susnux, line 16: I would not use a table for this, generally tables are bad for a11y and especially in this case where you do not have a table header.

Instead it probably makes more sense to use a list for this, semantically, and style it like you want.

elzody: Thanks for the feedback, I really appreciate any a11y tips I can get!

PR #4624, opened 2025-03-26
fix(settings): send ui_theme input to cool settings iframe

codewithvk:
* Resolves: # 
* Target version: main

### Summary
Collabora needs `ui_theme` to identify the theme mode for the settings iframe.

### DIFF
--- a/src/components/CoolFrame.vue
+++ b/src/components/CoolFrame.vue
@@ -13,6 +13,7 @@
 			<input type="hidden" name="wopi_setting_base_url" :value="wopiSettingBaseUrl">
 			<input type="hidden" name="iframe_type" :value="iframeType">
 			<input type="hidden" name="css_variables" :value="cssVariables">
+			<input type="hidden" name="ui_theme" :value="uiTheme">
 		</form>
 		<iframe :id="iframeName"
 			:name="iframeName"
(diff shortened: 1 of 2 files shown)


eszkadev:
I confirm it fixes dark mdoe for me

elzody:
The changes look sensible, but I'm wondering how to test this? I am assuming that Collabora could not get the proper light/dark theme settings from Nextcloud before, but even if my instance is set to dark mode and I open a document, the Collabora UI is still in light theme with these changes. Maybe this is something to due with the browsersetting.json, since in there dark mode is set to false?

codewithvk:
> The changes look sensible, but I'm wondering how to test this? I am assuming that Collabora could not get the proper light/dark theme settings from Nextcloud before, but even if my instance is set to dark mode and I open a document, the Collabora UI is still in light theme with these changes. Maybe this is something to due with the browsersetting.json, since in there dark mode is set to false?

Yes, that seems changing because of browser settings. Maybe @Rash419  has a better idea of why this is happening. Anyway, this change only alters the light/dark mode setting for the iframe.

elzody:
Thanks for the explanation

elzody (approved):

PR #4651, opened 2025-04-01
ci: unskip 'Save as' test

codewithvk:
* Resolves: [# 11297](https://github.com/CollaboraOnline/online/issues/11297)
* Target version: main

### Summary
As we already fix save as mode issue : https://github.com/CollaboraOnline/online/issues/11297#issuecomment-2749443148 we should try to unskip save as test

### DIFF
--- a/cypress/e2e/integration.spec.js
+++ b/cypress/e2e/integration.spec.js
@@ -68,7 +68,7 @@ describe('Nextcloud integration', function() {
 		cy.get('#tab-version_vue .version__info__label').contains('Current version')
 	})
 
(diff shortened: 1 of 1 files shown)


juliusknorr:
I guess the change is not released in the latest CODE docker image yet? That would be needed for the cypress release tests to pass

codewithvk:
> I guess the change is not released in the latest CODE docker image yet? That would be needed for the cypress release tests to pass

Ok, let's wait for the release.

juliusknorr (approved):

PR #4652, opened 2025-04-02
Add docx generation tasktype+provider

julien-nc:
Add TextToDocumentProvider and TextToDocumentTaskType to get a docx document from a user prompt.

For now we use TextToText to generate some Markdown which is converted to HTML. We then call `RemoteService::convertTo` to ask Collabora to convert the HTML to docx.


julien-nc:
@elzody Maybe something is wrong in my setup but the request to Collabora (`POST http://collabora.local/cool/convert-to/docx`) always responds with 400. Nothing in the response body. Any idea why?

juliusknorr:
Wondering if it is the reverse proxy, would be curious if you can run a curl directly in the collabora container to test

julien-nc:
Ok, in the TaskProcessing provider, passing a stream obtained like that works fine (imitating what `RemoteService::convertFileTo` does):
``` php
$userStorage = $this->rootFolder->getUserFolder('user1');
$file = $userStorage->getFirstNodeById(4830);
$fileName = $file->getStorage()->getLocalFile($file->getInternalPath());
$htmlStream = fopen($fileName, 'rb');
$docxContent = $this->remoteService->convertTo('document.html', $htmlStream, 'docx');
```

and these don't work:
``` php
$htmlStream = fopen(sprintf('data://text/html,%s', $htmlContent), 'r');
$docxContent = $this->remoteService->convertTo('document.html', $htmlStream, 'docx');
```

``` php
$htmlStream = fopen('php://temp', 'r+');
// or even $htmlStream = fopen('php://memory', 'r+');
fwrite($htmlStream, $htmlContent);
rewind($htmlStream);
$docxContent = $this->remoteService->convertTo('document.html', $htmlStream, 'docx');
```

So the question becomes, how do I get a proper stream from a string? :grin: 
@juliusknorr @elzody

julien-nc:
I went with
``` php
$tmpFilePath = $this->tempManager->getTemporaryFile();
file_put_contents($tmpFilePath, $htmlContent);
$htmlStream = fopen($tmpFilePath, 'rb');
```
which is working fine.
Is that fine with you?

elzody:
@julien-nc Ah I see, it's just streaming from the temp file then. That would be fine with me, seems the service is not ergonomic enough to accept a string -- and otherwise I would not be sure how to get a resource stream from that. A quick search on the web revealed `php://memory` and `php://temp` read-write streams you can use, but in the end, one stores the string in memory and the other creates a temp file anyway. So to me your method is just as good.

julien-nc:
Hey. Another weird issue. I'm implementing the spreadsheet generation which uses the csv->ods/xlsx conversion.

This request works fine with CURL:
``` bash
curl -F name=a.csv -F upload=@/tmp/a.csv http://collabora.local/cool/convert-to/xlsx --output /tmp/a.xlsx
```
But for some reason when it's done by the NC Http client, it reaches the 5 sec timeout.
`cURL error 28: Operation timed out after 5000 milliseconds with 0 bytes received`.

I'm producing the stream the exact same way as for html->docx/odt conversion. See `DocumentGenerationService::stringToStream`.
Then I'm passing it to `RemoteService::convertTo` the exact same way :grin:. Any clue what's happening?

julien-nc:
Found the issue after having tested every single way to produce a stream from a string :grin: . The issue was somewhere else. For some reason the request to `/cool/convert-to/xlsx` or `/cool/convert-to/ods` with CSV input works if we provide a filename in the multipart data. If we don't, nothing happens, Collabora does not read the data and our http client reaches its timeout.

So I made a change to `RemoteService::convertTo` to always add the `filename` in the multipart data. Maybe this should be optional and only done when needed to avoid risking to break other features. Wdyt?

julien-nc:
I updated the initial comment.

elzody:
@julien-nc  Thanks for doing the heavy lifting there. I'm not sure what specific features always including a file name would break, or if it would have any affect?

julien-nc:
@elzody The change to convertTo is safer now. It only adds the 'filename' multipart param for document generation.

elzody (approved):
This looks pretty good to me. It works well.
The only thing we can touch up on in the future is improving access to the file which gets generated. For example, I find it not very intuitive that you can only download the resulting file or create a public share link to it. Personally, I would expect it to be in my Nextcloud files already.

The failing tests seem to be related to the `nextcloud/ocp` package being out of sync, so we will have to resolve those separately.

julien-nc:
I added the PDF output in the odt/docs generation provider. HTML -> PDF works fine in Collabora.

elzody (approved):

elzody:
@julien-nc  I would say it looks like great, we can do follow-up pull requests to add anything else we need, like adding docs about needing several workers to complete the task. If you are also fine with it and don't have anything else to add immediately, we can merge it

julien-nc:
@elzody I added a small section in the README. Should I move it in a dedicated file in the docs directory? Feel free to push some adjustments.

lib/Service/DocumentGenerationService.php
```diff
@@ -0,0 +1,119 @@
...
+
+		return $taskOutputString;
+	}
+
+	private function stringToStream(string $text) {
+		$stream = fopen('php://memory', 'r+');
+		fwrite($stream, $text);
+		rewind($stream);
+
+		// This also works fine
```
juliusknorr, line 109: Not needed anymore?

elzody:
> @elzody I added a small section in the README. Should I move it in a dedicated file in the docs directory? Feel free to push some adjustments.

Since we adding more AI features here maybe it would indeed be a good idea to have a dedicated file for this in the docs directory in case we also want to add anything else related to the document generation at some point 👍🏻

lib/Service/DocumentGenerationService.php
julien-nc: If you're talking about the comment, it's been removed :+1:

julien-nc:
@elzody Sure, done.

Issue #4636, opened 2025-03-28
Undefined variable userid


Javohle:
https://github.com/nextcloud/richdocuments/blob/d794111915bdbdd42af9e60f06737fbdb44b6dca/lib/Controller/SettingsController.php#L499

Looks like this line is not useful and i will try to comment it because of pack of errors every time i open collabora office (standalone) or maybe something wrong

{"reqId":"XXXXXXXXXXXXXXXXXXXX","level":3,"time":"2025-03-28T19:24:41+00:00","remoteAddr":"XXX.XXX.XXX.XXX","user":false,"app":"PHP","method":"GET","url":"/apps/richdocuments/settings/systemconfig/2byHpCyDaNyosjguZzBFJVyRTF7BMn06/template/Vivid.otp?identifier=4612242","message":"Undefined variable $userId at /var/www/html/nextcloud/apps/richdocuments/lib/Controller/SettingsController.php#499","userAgent":"COOLWSD HTTP Agent 24.04.13.2","version":"31.0.2.1","data":{"app":"PHP"},"id":"67e6f7f6ee4d7"}

Javohle:
> Can you please tell us which version of richdocuments you are using? And by Collabora office "standalone" do you mean you have a separate Collabora server running (not using the built-in CODE server)?

Nextcloud office 8.6.3 (right now i'm updating it to 8.6.4 as it comes)
Collabora on separate server with proxy settings with ssl termination copied from collabora manual
All works fine but there are lot of errors in configure some of them just level3 without any message, some with message undefined variable.

########################################

 # Reverse proxy for Collabora Online   #

 ########################################


 AllowEncodedSlashes NoDecode
 ProxyPreserveHost On


 # static html, js, images, etc. served from coolwsd
 # browser is the client part of Collabora Online
 ProxyPass           /browser http://127.0.0.1:9980/browser retry=0
 ProxyPassReverse    /browser http://127.0.0.1:9980/browser


 # WOPI discovery URL
 ProxyPass           /hosting/discovery http://127.0.0.1:9980/hosting/discovery retry=0
 ProxyPassReverse    /hosting/discovery http://127.0.0.1:9980/hosting/discovery


 # Capabilities
 ProxyPass           /hosting/capabilities http://127.0.0.1:9980/hosting/capabilities retry=0
 ProxyPassReverse    /hosting/capabilities http://127.0.0.1:9980/hosting/capabilities


 # Main websocket
[...]

PR #4653, opened 2025-04-03
fix(setting): userId undefined issue for template

codewithvk:
* Resolves: #4636 
* Target version: main

### Summary

Now, we're generating the userId earlier, which prevents the issue of it being undefined.


juliusknorr (approved):

codewithvk:
@juliusknorr Can we merge and backport it please?

PR #4657, opened 2025-04-05
feat(setting): add dark/light iframe CSS vars

codewithvk:
* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/helpers/coolParameters.js
+++ b/src/helpers/coolParameters.js
@@ -47,12 +47,13 @@ const createDataThemeDiv = (elementType, theme) => {
 	document.body.appendChild(element)
 	return element
 }
-const generateCSSVarTokens = () => {
+
+const generateCSSVarTokens = (isSettingIframe = false) => {
 	/* NC versus COOL */
 	const cssVarMap = {
-		'--color-primary-element-text': '--co-primary-text',
-		'--color-primary-element': '--co-primary-element:--co-text-accent',
-		'--color-primary-element-light': '--co-primary-light:--co-primary-element-light',
+		'--color-primary-element-text': '--co-primary-text:--co-settings-btn-primary-text',
+		'--color-primary-element': '--co-primary-element:--co-text-accent:--co-settings-btn-primary',
+		'--color-primary-element-light': '--co-primary-light:--co-primary-element-light:--co-settings-btn-light',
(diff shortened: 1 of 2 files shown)


src/helpers/coolParameters.js
```diff
@@ -47,6 +47,42 @@ const createDataThemeDiv = (elementType, theme) => {
...
+	for (const cssVarKey in mapping) {
+		const computedStyle = window.getComputedStyle(element).getPropertyValue(cssVarKey)
+		if (!computedStyle) continue
+		const tokenName = mapping[cssVarKey]
+		result += tokenName + '=' + computedStyle.trim() + ';'
+	}
+	return result
+}
+
+const generateSettingCssVarTokens = () => {
```
juliusknorr, line 62: Can't we just reuse the iexisting css vars? Feels a bit like duplicating quite some logic

codewithvk: There are two points to consider:

- Missing Variables: Some variables are not currently being sent, so we need to add them.
- Theme-Based Values: For the settings iframe, we expect the variable values to be set based on the theme. (See the code pointer here: https://github.com/nextcloud/richdocuments/pull/4657/files#diff-087176fb56b5928e33a834558791bd88511f1897d4fd950a38b7334fe9c56c99R78-R82 .)

Since for cool iframe, we're not sending dark mode values(https://github.com/codewithvk/richdocuments/blob/private/codewithvk/pass_ui_theme_to_new/src/helpers/coolParameters.js#L105), I think creating new pair would make sense here... Any thought? @juliusknorr

elzody: If I understood you correctly, maybe then we could still use the `generateCSSVarTokens()` method and pass in some object containing values to override or something, and adjust the logic inside. That way we can use the same method and if you need to explicitly control the CSS variables (for in this case, the settings) then you can explicitly pass them to this method and it would mix them in, overriding preexisting ones.

Not sure if that makes sense or is feasible.

elzody:
@codewithvk Have not tested, but seems fine, thanks for that. Just needs the merge conflict resolved. Was going to try helping out with it myself but with the fork, not sure how it would work.

elzody (approved):

Issue #4670, opened 2025-04-11
Browser settings upload does not use correct WOPI callback URL


elzody:
The `wopi_callback_url` used by the browser settings upload in personal settings is taken from a config value which does not exist because the correct initial state is not provided, this it defaults to the host and protocol of the current browser location, which does not always match what the configured callback URL should be, thus failing to upload and fetch the settings from Collabora's side.

PR #4671, opened 2025-04-11
fix: load `wopi_callback_url` from initial state into config service

elzody:
* Resolves: https://github.com/nextcloud/richdocuments/issues/4670
* Target version: main

### Summary
We need to tell Collabora the correct URL to use to connect back to Nextcloud for uploading the settings, and to do that we need to make sure we always have the `wopi_callback_url` available to us via the config service. We check if it is provided by initial state, and if not, we default to the current host and port of the browser.

### Checklist

### DIFF
--- a/src/services/config.tsx
+++ b/src/services/config.tsx
@@ -9,7 +9,10 @@ class ConfigService {
 	private values: {[name: string]: any}
 
 	constructor() {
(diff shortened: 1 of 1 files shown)


mejo- (approved):
Didn't test, but your explanations sound reasonable.

PR #4712, opened 2025-04-28
fix(Wopi): fall back to super share if share token is not available

blizzz:
* Target version: main

### Summary

On internal shares the controller is called without the share token. But necessary information, like share attributes, might be necessary to know and are available from the super share of the SharedMount in that case.

This can be testes as following:

1. In Admin settings → Office enable Secure view 
2. and there tick *Show watermark for shares without download permission* (but no other if its options!)
3. Share a document with an internal user as **read-only** and **without download-privileges**
4. As this user, open the document

It is expected that the document is opened with the watermark, but it does not. This is because we do not store (for we do not know) the share token of the document in `DocumentController::token()`. Hence we try now in the WopiController whether the share token is present and otherwise fall back to the super share.

The super share combines any relevant present share. While some field stay empty (shareWith, shareToken, as they cannot be combined) the share attributes are.  So if there would be another share of this document with the user, that does have download permissions, then the watermarking would not happen, as expected.

### Checklist

### DIFF
--- a/lib/Controller/AssetsController.php
+++ b/lib/Controller/AssetsController.php
@@ -6,9 +6,9 @@
 
 namespace OCA\Richdocuments\Controller;
 
-use OCA\Files_Sharing\SharedStorage;
 use OCA\Richdocuments\Controller\Attribute\RestrictToWopiServer;
 use OCA\Richdocuments\Db\AssetMapper;
+use OCA\Richdocuments\Helper;
 use OCA\Richdocuments\Service\UserScopeService;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Db\DoesNotExistException;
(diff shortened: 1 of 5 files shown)


blizzz:
Only now I see that you are doing the almost same in other places, actually, but with SharedStorage instead of SharedMount. Do you want me to switch?

juliusknorr:
> Only now I see that you are doing the almost same in other places, actually, but with SharedStorage instead of SharedMount. Do you want me to switch?

Shouldn't make much of a difference, right? At least I cannot see any obvious one. I'd be fine either way but prefer consistency

blizzz:
> > Only now I see that you are doing the almost same in other places, actually, but with SharedStorage instead of SharedMount. Do you want me to switch?
> 
> Shouldn't make much of a difference, right? At least I cannot see any obvious one. I'd be fine either way but prefer consistency

No difference, but on the same line with consistency. I'll update then.

juliusknorr (approved):

juliusknorr:
Rebase should have passing CI then as well

blizzz:
Also consolidated the handling of getting the share a bit across the occurrences.

blizzz:
Cypress still keeps failing though. Is that what should have been fixed?

elzody (approved):
@blizzz Does this also need any backports

Issue #4402, opened 2025-01-23
New file creation takes a really long time


elzody:
**Describe the bug**
When creating a new file on a fresh NC30, it takes a really long time when creating a file for the template picker to appear. When the richdocuments app is disabled, this issue does not occur. It is probably something to do with templating, or Collabora trying to get some additional file information.

**To Reproduce**
1. Have a fresh NC30 or private browser window with no cache
2. Try to create a new text file

**Expected behavior**
It is expected that new file creation happens quickly and that the template picker does not take so long to appear.

Javohle:
Hi, had this issue, and in Administration panel->Office  put an mark on checkbox "use OXML as default " (first option in checkboxes under collabora page) and templates picker window starts to appear momently. 


The other thing i cant get where it takes its start is pack of error messages within the message all with "app":"richdocuments","method":"GET","url":"/ocs/v2.php/apps/files/api/v1/templates","message":" and even debug and other logs tells nothing more. All works but  journal is dirty)

pedropintosilva:
@mmeeks , @annanazarayan and @jazevedo-coll  have reported similarly that creating a new document takes long time with:

COOLWSD version: 24.04.12.4snapshot(git hash: 452709a491 (E))
LOKit version: Collabora Office 24.04.12.20250226(git hash: ef9dc9e4b2)
Served by: Ubuntu 22.04.5 LTS
Server ID: b2cf79d0
WOPI host: nextcloud 30.0.6.3 - richdocuments 8.5.4


> This is not always reproducible. But there are instances when creating a new file takes a very long time.

mmeeks:
Lets qualify a long time - I happened to know that having entered the file-name I might be in for a wait, so counted seconds up to 72 of them before bingo - the template dialog appeared with thumbnails :-) with no feedback during the process: I can only assume lots of templates are being re-thumbnailed or something (?)

picsel2:
I have the same issue and noticed that removing the bulk of vendored templates helped.

elzody:
The issue is that we send each template to Collabora for field extraction before showing the template picker. This is not a very effecient way to do it. There is an upcoming fix that will only do the field extraction once the user selects which template they want to use, resulting in only one file being sent to Collabora (or not, if the fields are cached and the document hasn't changed) and results in a better UX and performance.

juliusknorr:
Fixed by https://github.com/nextcloud/server/pull/52514

PR #4713, opened 2025-04-28
fix: check if template fields should be extracted

elzody:
* Resolves: https://github.com/nextcloud/richdocuments/issues/4402
* Target version: main
* Requires: https://github.com/nextcloud/server/pull/52514

### Summary
With the new server changes, we need to check if the event tells us to fetch the fields or not. If the fields should be extracted, then we do that. On the server side, this will only be triggered once the user makes a template selection, and only that template will be sent.

The documentation has been updated for the server pull request, and the `templates.spec.js` tests have been expanded a bit in this pull request to verify that no template fields are extracted at first.

### Checklist

### DIFF
--- a/lib/Listener/BeforeGetTemplatesListener.php
+++ b/lib/Listener/BeforeGetTemplatesListener.php
@@ -24,6 +24,7 @@ class BeforeGetTemplatesListener implements IEventListener {
 			return;
 		}
 
+		/** @psalm-suppress RedundantCondition */
 		if (method_exists($event, 'shouldGetFields') && !$event->shouldGetFields()) {
 			return;
 		}
--- a/lib/Listener/BeforeGetTemplatesListener.php
+++ b/lib/Listener/BeforeGetTemplatesListener.php
@@ -24,10 +24,14 @@ class BeforeGetTemplatesListener implements IEventListener {
 			return;
 		}
 
+		if (method_exists($event, 'shouldGetFields') && !$event->shouldGetFields()) {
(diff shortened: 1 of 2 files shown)


juliusknorr:
This is obsolete with the slightly different approach in https://github.com/nextcloud/server/pull/52514, right?

elzody:
@juliusknorr Not necessarily. We still need to check `$event->shouldGetFields()` to see if we should actually react to the event or not. So that will be here, plus an edit to some tests.

lib/Listener/BeforeGetTemplatesListener.php
```diff
@@ -20,14 +20,14 @@ public function __construct(
 	}
 
 	public function handle(Event $event): void {
-		if (!$event instanceof BeforeGetTemplatesEvent) {
+		if (!$event instanceof BeforeGetTemplatesEvent || !$event->shouldGetFields()) {
 			return;
 		}
```
juliusknorr, line 26: To make sure richdocuments still works with the old version (and have an easy to read conditions): 
```suggestion
		if (!$event instanceof BeforeGetTemplatesEvent) {
			return;
		}
		
		if (method_exists($event, 'shouldGetFields') && !$event->shouldGetFields()) {
			return;
		}
```

juliusknorr (changes requested):
Small addition to account for compatibility, but otherwise 👍

juliusknorr (approved):

PR #4714, opened 2025-04-29
fix: do not fetch remote collabora url for own server

elzody:
* Target version: main

### Summary
When your own server is listed as a trusted server in the federated sharing options, there is no need to fetch the remote Collabora URL, as it would just be the one set internally.

### TODO
- [x] Tests

### Checklist

### DIFF
--- a/lib/Service/FederationService.php
+++ b/lib/Service/FederationService.php
@@ -24,6 +24,7 @@ use OCP\ICache;
 use OCP\ICacheFactory;
 use OCP\IRequest;
 use OCP\IURLGenerator;
+use OCP\Security\ITrustedDomainHelper;
 use OCP\Share\IShare;
 use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\NotFoundExceptionInterface;
@@ -43,6 +44,7 @@ class FederationService {
 		private AppConfig $appConfig,
 		private IRequest $request,
 		private IURLGenerator $urlGenerator,
+		private ITrustedDomainHelper $trustedDomainHelper,
 	) {
 		$this->cache = $cacheFactory->createDistributed('richdocuments_remote/');
 		try {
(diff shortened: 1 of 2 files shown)


lib/Service/FederationService.php
```diff
@@ -73,6 +73,14 @@ public function getRemoteCollaboraURL($remote) {
 		if (!$this->isTrustedRemote($remote)) {
 			throw new \Exception('Unable to determine collabora URL of remote server ' . $remote . ' - Remote is not a trusted server');
 		}
+
+		$serverProtocol = $this->request->getServerProtocol() . '://';
+		$serverHost = $this->request->getServerHost();
+		$server = $serverProtocol . $serverHost . $this->request->getRequestUri();
+		if (str_starts_with($server, $remote)) {
```
juliusknorr, line 80: We can actually use https://github.com/nextcloud/server/blob/dae7c159f728a90ffa53247d6e033abdae5d2bd6/lib/public/Security/ITrustedDomainHelper.php#L27 to check that which then would also cover shares on the same instance with different urls

lib/Service/FederationService.php
```diff
@@ -113,7 +124,7 @@ public function isTrustedRemote($domainWithPort) {
 				break;
 			}
 			$regex = '/^' . implode('[-\.a-zA-Z0-9]*', array_map(fn ($v) => preg_quote($v, '/'), explode('*', $trusted))) . '$/i';
```
juliusknorr, line 126: Maybe we can add a comment so the next reader has it easier then us in our call pulling this into pieces to figure out again what it does ;)

elzody:
@juliusknorr I am pretty sure the failures are related now somehow, but my local setup is giving me troubles with SSL and I am not able to run them myself to check. I will check more tomorrow if I can figure it out

elzody:
@juliusknorr The other failing tests are resolved by not including the subpath fix here, it seems. but, we still have these other failing tests.
My environment really is not working due to SSL problems when trying to create federated shares to itself, and I can't get any of the tests to even run locally.

blizzz (approved):

PR #4731, opened 2025-05-06
add an initial state for 'browser_setting' variable.

hcvcastro:
After the file browsersettings.json is uploaded, it’s necessary to create an initial state for the user’s browser preference settings. This enables the Collabora iframe to include the POST variable and serve preprocessed HTML with accessibility features enabled, for example.

### DIFF
--- a/lib/Service/InitialStateService.php
+++ b/lib/Service/InitialStateService.php
@@ -14,6 +14,7 @@ use OCA\Richdocuments\Db\Wopi;
 use OCA\Richdocuments\TemplateManager;
 use OCA\Theming\ImageManager;
 use OCP\AppFramework\Services\IInitialState;
+use OCA\Richdocuments\Service\SettingsService;
 use OCP\Defaults;
 use OCP\IConfig;
 use OCP\IURLGenerator;
@@ -31,6 +32,7 @@ class InitialStateService {
 		private Defaults $themingDefaults,
 		private IConfig $config,
 		private ?string $userId,
+		private SettingsService $settingsService,
 	) {
 	}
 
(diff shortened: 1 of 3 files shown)


elzody:
The code seems reasonable, even if I have a bit of a difficult time understanding its purpose. The only thing that would need to be resolved is the PHP lint failure, which can be fixed with a simple `composer run cs:fix`.

hcvcastro:
Thanks! I’ll fix it using composer run cs:fix.

lib/Service/InitialStateService.php
```diff
@@ -110,5 +112,12 @@ private function provideOptions(): void {
 
 		$this->initialState->provideInitialState('theming-customLogo', $logo);
 		$this->initialState->provideInitialState('open_local_editor', $this->config->getAppValue(Application::APPNAME, 'open_local_editor', 'yes') === 'yes');
+		$fileName = $this->config->getUserValue($this->userId, Application::APPNAME, 'browsersetting');
+		if ($fileName) {
+			$browserSetting = $this->settingsService->getSettingsFile('userconfig/' . $this->userId,
```
juliusknorr, line 117: This is rather bad performance wise. File system access is rather slow and the initial state is built on every page load, so wondering if we shall rather write this to a memory cache

juliusknorr (changes requested):

lib/Service/InitialStateService.php
hcvcastro: You’re correct. We’re testing some features, and the patch will likely be discarded. If not, I’ll improve it. Thanks!

PR #4739, opened 2025-05-07
Fix save as on mobile with direct editing

juliusknorr:
- **fix(direct-editing): Use proper path in the save as dialog**
- **fix(direct-editing): Use correct method to pass the Action_SaveAs post message**

### Summary

When using "Save as" through direct editing in the mobile apps the wrong method call was introduced in https://github.com/nextcloud/richdocuments/commit/c6306bb4a367f12701a43aa138ce570148a41d5f#diff-d6043d5e8ee25405786521e86514439eeb1b9d1cc02a7b744cea3084a13b41f0R509 causing the modal to not work.

Steps to reproduce:

- Create a file
- Generate a direct editing token and url for richdocuments with with the curl below
- Open the URL in a private browser window and try to use the save as function

```
`curl --request POST \
  --url 'http://nextcloud.local/ocs/v2.php/apps/richdocuments/api/v1/document?format=json' -u admin:admin \
  --header 'Content-Type: multipart/form-data' \
  --header 'OCS-Apirequest: true' \
  --form fileId=302`
```

### Checklist

### DIFF
--- a/cypress/e2e/direct.spec.js
+++ b/cypress/e2e/direct.spec.js
@@ -161,4 +161,41 @@ describe('Direct editing (legacy)', function() {
 		})
 	})
 
+	it('Save as', function() {
+		createDirectEditingLink(randUser, fileId)
+			.then((token) => {
+				cy.nextcloudTestingAppConfigSet('richdocuments', 'uiDefaults-UIMode', 'tabbed')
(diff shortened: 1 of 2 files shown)


elzody (approved):
Works great

PR #4787, opened 2025-05-19
fix(CoolFrame): Init and accept postMessage for setting iframe height

codewithvk:
* Resolves: # 
* Target version: main

### Summary
Appropriate COOL PR: https://github.com/CollaboraOnline/online/pull/11971

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/CoolFrame.vue
+++ b/src/components/CoolFrame.vue
@@ -28,6 +28,7 @@
 
 import { generateCSSVarTokens, getUITheme } from '../helpers/coolParameters.js'
 import { languageToBCP47 } from '../helpers/index.js'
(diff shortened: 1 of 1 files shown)


src/components/CoolFrame.vue
```diff
@@ -79,14 +86,32 @@ export default {
...
 }
 </script>
 
   <style scoped>
   .cool-frame-iframe {
     width: 100%;
     border: none;
-	height: 2000px;
-	overflow-y: auto;
+	overflow-y: hidden;
```
juliusknorr, line 112: Should we still overflow auto just to be sure that if something goes wrong users are still able to reach all elements in the iframe?

elzody (approved):

PR #4807, opened 2025-05-26
fix: allow document creation only for users that can edit

tanyaka:
### Summary
This fix assures files creation from templates is not possible for users with read-only permissions.
Users with read-only permissions should not have option to create new document from diagram, document, presentation or spreadsheet template, see attached screenshot.

### Checklist

### DIFF
--- a/lib/Listener/RegisterTemplateFileCreatorListener.php
+++ b/lib/Listener/RegisterTemplateFileCreatorListener.php
@@ -34,7 +34,7 @@ class RegisterTemplateFileCreatorListener implements IEventListener {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


elzody:
I can't reproduce the issue here, can you quickly remind me what to do to reproduce this issue?

tanyaka:
> I can't reproduce the issue here, can you quickly remind me what to do to reproduce this issue?

We have found out this issue while working with collabora read-only. Here the user should not be able to edit the documents, but the current state is, that he can create and edit them.

elzody:
I was able to reproduce and see what you mean now. However, in order to get the tests passing, you will need to rebase onto the latest main branch of richdocuments.

printminion-co:
@juliusknorr will it be merged into master?

elzody (approved):

PR #4809, opened 2025-05-26
[stable30] fix: for shares allow document creation only for users that ca…

tanyaka:
### Summary

This fix assures files creation from templates in public share is not possible for users with read-only permissions.
Users with read-only permissions should not have option to create new document from diagram, document, presentation or spreadsheet template, see attached screenshot.

This is a fix for nextcloud v30.

### Checklist

- [ ] Code is properly formatted
- [ ] Sign-off message is added to all commits
- [ ] Documentation (manuals or wiki) has been updated or is not required
shares-add-new-files-with-doc-creation

PR #4794, opened 2025-05-22
fix: Create new files on public pages for 31+

juliusknorr:
Signed-off-by: Julius Knorr <jus@bitgrid.net>

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

- [ ] Code is properly formatted
- [ ] Sign-off message is added to all commits
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #4831, opened 2025-06-05
fix(shares): guest users inherit edit_group

elzody:
* Follow up to: https://github.com/nextcloud/richdocuments/pull/4809
* Target version: stable30

### Summary
With the merging of https://github.com/nextcloud/richdocuments/pull/4809, I realized that it somewhat broke being able to create new documents in a public share by a guest user. This is being the guest user is not a real user, and the `edit_groups` could not be determined. In the back end, we have a way to do this, so it is injected as an initial state that we get from the front end. It is inherited from the share owner.

Example:
- Share owner that is not in a group with edit permissions creates a *public share with editing permissions*
- A guest user accesses the public share
- The guest user can not create new documents in the public share, and also can not edit documents, as it inherits the group permissions

It is mostly the same as before, except now the *New file* menu should reflect the proper permissions.

### Checklist

### DIFF
--- a/lib/Listener/ShareLinkListener.php
+++ b/lib/Listener/ShareLinkListener.php
@@ -42,8 +42,11 @@ class ShareLinkListener implements \OCP\EventDispatcher\IEventListener {
 		$loggedInUser = $this->permissionManager->loggedInUser();
 
 		if ($this->permissionManager->isEnabledForUser($owner)) {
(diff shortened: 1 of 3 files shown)


juliusknorr:
Should we forward port as well was we have this back on main with https://github.com/nextcloud/richdocuments/pull/4794?

elzody:
@juliusknorr Yes, in that case we probably should

Issue #4622, opened 2025-03-25
Enable CSV to spreasheet conversation


juliusknorr:
Just as a reminder to follow along

- We need this for AI tasks
- Needs some more research / check back with Collabora as csv might need some speicifc format/separators
- Needs some implementation to add the mimetypes

elzody:
I have an issue open with Collabora right now to enable passing data such as a delimiter, etc. along with the `convert-to` request: https://github.com/CollaboraOnline/online/issues/11440

juliusknorr:
- [ ] Check if auto import was improved, or if we need to finetune parameters, otherwise close this one

elzody:
I gave it some testing, and it does not automatically detect the separator or anything, but we can now at least specify some sane defaults. This would at the very least work in the case of the AI spreadsheet generation, as we can make sure that the LLM produces only CSV files with a certain field separator, text delimiter, etc.

You can't pass JSON yet, so it is sort of unreadable in its current form -- you have to specify ASCII codes for the field separator, text delimiter, and character set:

Field separator: `,`
Text delimiter: `"`
Character set: `UTF-8`

`curl -k -F "data=@input.csv" -F "format=ods" -F "infilterOptions=44,34,76,1" https://cool.local:9980/cool/convert-to > output.ods`

It's not that easy to reason about, at least it wasn't for me at first. Seems like JSON support is planned but for now this is what made it in.

https://github.com/CollaboraOnline/online/pull/12011
https://wiki.openoffice.org/wiki/Documentation/DevGuide/Spreadsheets/Filter_Options#Filter_Options_for_the_CSV_Filter

@juliusknorr @julien-nc

elzody:
For the normal file conversion we introduced via the file action menu or API, we could either just assume the defaults and hope for the best or present the user with some kind of import dialog on our own (but that seems a bit too complex right now). Not sure how to proceed there for now unless the field separator auto detection logic is also applied to the `convert-to` API on Collabora's side.

juliusknorr:
AI related parts good enough I'd say. For general file conversion it is then rather tricky without having options implemented for the conversion API on the Nextcloud side.

elzody:
I went ahead and made a pull request to introduce it for the AI-generated spreadsheets at least: https://github.com/nextcloud/richdocuments/pull/4843

PR #4843, opened 2025-06-11
feat(ai): spreadsheet conversion input filters

elzody:
* Part of: https://github.com/nextcloud/richdocuments/issues/4622
* Target version: main

### Summary
Pass the required input filters to correctly parse the CSV output that the LLM generates. We assume that the LLM generates an output that always uses a set of standards, like `,` for the field separator and `"` as the text delimiter.

The `RemoteService::convertTo()` method can now accept conversion options which are mixed in with the multipart form data, as required to pass input filters. But they are optional.

### Checklist


juliusknorr:
Cypress failure seems to originate from server somehow

PR #4848, opened 2025-06-13
Show statusbar by default for all doc types

pedropintosilva:
Status bar is expected to be visible when working with office documents as it has important status (word count, formula, selection) and controls (zoom, page navigation).

This seems to be what office users expect by default. Nevertheless, user can still hide it by going to View > Status Bar and that preference is then remembered.

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


elzody:
The Cypress test failures aren't related and likely related to a server issue or something.

PR #4859, opened 2025-06-16
fix(l10n): Fix missing translation of guest name dialog

nickvergessen:
* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


juliusknorr:
y

PR #4866, opened 2025-06-23
fix: undefined share for preview

elzody:
* Target version: main

### Summary
Prevents `$share` being `undefined` and causing an error in the logs. If we cannot compute `$share` for some reason, we simply pass `null` to the `PermissionManager::shouldWatermark()` method and it will handle the rest.

### Checklist

### DIFF
--- a/lib/Listener/BeforeFetchPreviewListener.php
+++ b/lib/Listener/BeforeFetchPreviewListener.php
@@ -38,6 +38,7 @@ class BeforeFetchPreviewListener implements IEventListener {
 		}
 		$shareToken = $this->request->getParam('token');
 
+		$share = null;
 		try {
 			$share = $shareToken ?
 				// Get different share for public previews as the share from the node is only set for mounted shares


lib/Listener/BeforeFetchPreviewListener.php
```diff
@@ -39,14 +39,17 @@ public function handle(Event $event): void {
...
+			// Get different share for public previews as the share from the node is only set for mounted shares
+			$shareByToken = $shareToken ? $this->shareManager->getShareByToken($shareToken) : null;
+
+			// Get share for internal shares
+			$shareFromNode = $this->helper->getShareFromNode($event->getNode());
 		} catch (ShareNotFound) {
 		}
 
+		// Do not leave as undefined if neither found
+		$share = $shareByToken ?? $shareFromNode ?? null;
```
blizzz, line 51: this would call `$this->helper->getShareFromNode()` even if not needed. Would it not suffice to initialize `$share` with `null` up front?

elzody: You're right, good catch

elzody:
The Cypress failures are unrelated. Still haven't had the chance to track down this crypto UUID stuff.

PR #4881, opened 2025-07-07
chore(dep): update `@nextcloud/auth` to v2.5.2

elzody:
* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "9.0.0-dev.0",
       "license": "AGPL-3.0-or-later",
       "dependencies": {
-        "@nextcloud/auth": "^2.4.0",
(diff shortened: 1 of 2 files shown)


elzody:
It seems that the issue for basically all of the failing Cypress tests is an error about `self.crypto.randomUUID` not being a function



Seems like it was fixed in `@nextcloud/auth`, but there has no been a new release

https://github.com/nextcloud/server/pull/53635
https://github.com/nextcloud-libraries/nextcloud-auth/pull/822

elzody:
The other failures are unrelated, as the issue has been fixed in `@nextcloud/auth` v2.5.2, but there is still a server bug relating to `self.crypto.randomUUID`:

PR #4902, opened 2025-07-22
feat(wopi): WOPI proof

elzody:
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Service/ProofKeyService.php
+++ b/lib/Service/ProofKeyService.php
@@ -8,6 +8,9 @@ declare(strict_types=1);
 
 namespace OCA\Richdocuments\Service;
 
+use DateTime;
+use DateTimeImmutable;
+
 class ProofKeyService {
 	// The Windows epoch is used for WOPI timestamps (as it is a MS protocol)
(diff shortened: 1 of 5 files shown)


lib/Service/ProofKeyService.php
```diff
@@ -0,0 +1,143 @@
...
+		// Four bytes representing the length, in bytes, of the URL
+		$urlLength = pack('N', strlen($url));
+
+		// The UTF-8 encoded URL converted to uppercase
+		$uppercaseURL = mb_strtoupper(mb_convert_encoding($url, 'UTF-8', 'auto'), 'UTF-8');
+
+		// Four bytes representing the size, in bytes, of the WOPI timestamp
+		//     Note: The WOPI timestamp should be converted to a long, so this
+		//           is architecture dependant (here we use PHP_INT_SIZE)
+		$wopiTimestampSize = pack('N', 8);
```
blizzz, line 112: Since it is architecture dependent, use the constant directly instead of the hard coded value. Unless I misunderstood something?

PR #4942, opened 2025-08-13
feat(ai): AI-generated presentations

elzody:
* Resolves:
* Target version: main

### TODO

- [ ] ...

### Checklist


julien-nc:
I tried to generate slide documents with this branch.
* The title sometimes overlaps with the slide content
image

* The Assistant mime type detection (done with `mime_content_type($taskOutputFile->fopen('rb'))`) does not work anymore. It works on the files that were generated 2 month ago. I have no idea why because once downloaded:

``` bash
$ file 925-3293
925-3293: Microsoft PowerPoint 2007+
```
* One slide is systematically repeated

image image 
image

lib/TaskProcessing/SlideDeckGenerationProvider.php
```diff
@@ -80,10 +80,15 @@ public function process(?string $userId, array $input, callable $reportProgress)
...
-			$input['text'],
-		);
+		try {
+			$response = $this->slideDeckService->generateSlideDeck(
+				$userId,
+				$input['text'],
+			);
+		} catch (\RuntimeException $e) {
+			// We can retry in case of faulty LLM JSON
+			$this->process($userId, $input, $reportProgress);
```
julien-nc, line 90:
```suggestion
			return $this->process($userId, $input, $reportProgress);
```
But you risk an infinite loop here.

elzody:
> The Assistant mime type detection (done with `mime_content_type($taskOutputFile->fopen('rb')))` does not work anymore

The MIME type is properly detected for me, can you try it again? Additionally, your concerns have been resolved.

@julien-nc

julien-nc:
@elzody Nice!
A few more remarks or questions:

* The main text field could be larger otherwise the text is partially out of the slide.

image

image

* When the content of a slide is a bullet list, it is not rendered as such. The content that gives the slide in the previous screenshots is:
``` json
"content": [
            "In Germany, a unique law challenged moustache norms",
            "A Bavarian town had a law, 'No moustache, no marriage'",
            "The rationale was moustaches could act as soup strainers",
            "To this day, the town carries out Moustache day annually"
]
```

> The MIME type is properly detected for me, can you try it again?

Do you mean that you see a preview there?
image

And when you click this preview, is the file opened in Collabora by the viewer? On my side, as the mimetype is `application/octet-stream`, the viewer says
image
I hope it's an issue on my side but I suspect that Php's `mime_content_type` fails to detect the mimetype of pptx files.
I tried to make it generate odp files and the mimetype detection works so the preview is there and the file can be directly opened in the viewer/collabora.

Maybe you're always generating odp files. If so, can you try with ooxml/pptx?

elzody:
@julien-nc I experience the same thing with `.pptx` files. I looked into it, because I saw no reason for it to not detect it correctly, but it turns out that `.pptx` files are special in that they're really just `.zip` files and they do not contain a MIME type in the actual file data -- you just have to verify that it's a zip file and verify that the structure matches that of a `.pptx` file. So, that's why it thinks it's some binary file or something -- because it's literally a `.zip` file. Not sure where to go from here.

lib/Service/SlideDeckService.php
```diff
@@ -20,11 +24,39 @@
...
 	public const PROMPT = <<<EOF
-Draft a presentation slide deck with headlines and a maximum of 5 bullet points per headline.
-Use the following JSON structure for your whole output and output only the JSON array:
+Draft a presentation with slides based on the following JSON.
+Replace the title, subtitle, and content with your own.
+If the content is an array of bullet point strings, replace them as necessary and always use at least four of them. Do not place any dot (.) or hyphen (-) before the bullet points.
+Choose one of the following three presentation styles and replace the "presentationStyle" field with your choice: security, pitch, triangle
+The slide titles should not contain more than three words.
+Use the following JSON structure for your entire output.
+Output 5 or more of the JSON objects, and use each of them at least once.
```
marcelklehr, line 33: What does this mean?

lib/Service/SlideDeckService.php
```diff
@@ -20,11 +24,39 @@
...
+Output only the JSON array:
 
 ```
-[{"headline": "Headline 1", "points": ["Bullet point 1", "Bullet point 2"]}, {"headline": "Headline 2", "points": ["Bullet point 1", "Bullet point 2"]}]
+[
+    {
+        "layout": 0,
```
marcelklehr, line 39: I just saw that there is also a 3rd layout, would it make sense to incorporate that in the examples here, so that the LLM knows it can choose that one as well?

lib/Service/SlideDeckService.php
elzody: Slide titles shouldn't contain more than three words, the following JSON structure with the defined layouts are to be used in the output, and to use at least 5 of them to give at least 5 slides, and use at least each layout once, making sure that both bullet points and regular paragraph content are used.

marcelklehr: > and use each of them at least once.

Took me a while to grasp this but, I guess since it works, it's ok. I'm not a prompt engineer either :D

marcelklehr:
> I'm not able to get presentations with more than 5 slides

Mh, worked for me with Llama 3.1 8B

julien-nc:
@marcelklehr Good to know. I was using gpt-4 in my tests and prompts like `story of europe in 10 slides`.

PR #4965, opened 2025-08-27
Allow Collabora Online to fetch settings config from richdocuments

mohit-marathe:
This allows Collabora Online to have an in-app settings iframe dialog just like the one in Nextcloud Settings -> Office

* Target version: main

### Checklist

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -414,7 +414,7 @@ class WopiController extends Controller {
 
 		try {
 			$wopi = $this->wopiMapper->getWopiForToken($access_token);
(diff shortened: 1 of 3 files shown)


elzody:
Rebasing onto the latest `main` branch should resolve most of the failing integration and Cypress tests. I expect the failures to be unrelated, as we have an issue on `main` right now with some development dependencies causing the failures.

For some more context, these changes allow Collabora to fetch user settings from Richdocuments, or do I misunderstand?

mohit-marathe:
> For some more context, these changes allow Collabora to fetch user settings from Richdocuments, or do I misunderstand?

That's right. Now that I read the PR title, it does sound misleading. I'll change the title.

pedropintosilva:
Should be backported to 31 please :)

Issue #3456, opened 2024-02-03
richdocuments effectively disable Collabora if onlyoffice plugin is installed


gchatthes:
**Describe the bug**
If richdocuments detects onlyoffice plugin, it completely disables handling of any documents  which are compatible with onlyoffice. This shouldn't be the correct approach, since it effectively allows only a few documents to be handled by richdocuments and the user has no way of opening those documents with Colabora.
Furtermore, onlyoffice allows to open any documents with the context menu so the correct approach would be to have richdocuments handle all office documents by default and if user want's to open a document with onlyoffice, he can do so by opening the context menu and selecting onlyoffice.
The code which handles this is in file:
./lib/Service/CapabilitiesService.php (line 132)
Deleting lines 132->135 fixes this issue.

**To Reproduce**
1. install onlyoffice plugin (no need to configure it)
2. Click on any document which is supported by onlyoffice (docx, xlsx etc.)
3. Documents either open with onlyoffice (if correctly configured), or a download dialog appears

**Expected behavior**
If the document is supported by richdocuments it should open Colabora. If user wants to open it with onlyoffice, he can do so by selecting the appropriate option in the context menu.

**Client details:**
 - OS: Debian 12 - Bookworm
 - Browser: any
 - Version: any
 - Device: desktop

## Server details

**Operating system**: Debian 12 Bookworm / Docker install (latest)

**Web server: Apache/2.4.57 (Debian) / within docker

**Database: Mariadb, docker image version latest (11.2.2)

**PHP version: php 8.2.15

**Nextcloud version: 8.0.2

**Version of the richdocuments app: 8.3.1

**Version of Collabora Online: 23.05.8.4

chunsiyang:
same here with the following environment 
nextcloud 28.0.3
nextcloud office 8.3.2
Collabora  server 23.05.9.3 11d108b
onlyoffice 9.0.0

thebearon:
This is since #2301, made for #2281.

unnilennium:
confirm the behaviour which is new after upgrade from 26 to 28.0.4
One could prefer onlyoffice for some use case and keep richdocument for all other documents !

joshtrichards:
Related: nextcloud/server#5065

valerio-bozzolan:
If I understand correctly the title of this issue could be:

"Allow users to 'Open File With': selecting Collabora or OnlyOffice"

valerio-bozzolan:
[...]

PR #4977, opened 2025-08-31
Add Config for Open With Capability

### LINKED ISSUE #305: App breaks "share link" for users
### Steps to reproduce
1. Enable app
2. Select details for a folder
3. In sharing enable share link
4. Go to link in new window, click on file you want to download

### Expected behaviour
File should download

### Actual behaviour
Page starts to load something, never completes

Additional info:

right click and opening in new window/tab works as expected i.e. the file starts to download;
disabling the Collabora app restores expected functionality.

NB the Collabora app is only installed for testing purposes and is available to a very small group of users (though see #292 ) and certainly not for documents shared by link.

lars-sh:
* Resolves: #305
* Target version: main

### Summary

I installed Nextcloud Office on my Nextcloud instance. Anyway, most of the time I simply want to download the file and view or edit it locally. Even more, this applies to receivers of share links.

Therefore, this pull request adds one system config property to control the behavior of the Nextcloud Office app to open all kind of files by default. When setting the property "richdocuments_optional_mimetypes" to "all" or an array of mimetypes files are simply downloaded.

In addition, I made the context menu function "Edit with Nextcloud Office" a little more flexible to apply to all of the mimetypes that Nextcloud Office is capable to open but does not open by default. Thanks to that change, I'm still able to use Nextcloud Office on demand.

This is somewhat of a prototype. Nonetheless, I hope it helps someone to integrate a similar feature.

### Checklist

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -11,6 +11,7 @@ namespace OCA\Richdocuments;
 use OCA\Richdocuments\Service\CapabilitiesService;
 use OCP\App\IAppManager;
 use OCP\Capabilities\ICapability;
+use OCP\IConfig;
 use OCP\IURLGenerator;
 
 class Capabilities implements ICapability {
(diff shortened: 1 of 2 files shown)


lars-sh:
I just noticed, the file-action.js change should fix #3456 as well.

Issue #4168, opened 2024-10-21
Files directly download from shared links instead of co-editing


mmeeks:
**Describe the bug**
Someone shares an internal link to a document with me. I click on it, but instead of entering the document I get one of these three behaviors:

1. the file immediately downloads and opens locally - nice but no collaboration / cookie etc.
2. the file-view opens on the folder with the file highlighted but not opening
3. the file fails to load like this: https://github.com/nextcloud/richdocuments/issues/4139
4. the file actually opens corectly.

Chance of getting to 4. is quite a lot lower than it used to be, previously you would get 2. and 4. now I get lots of 1. and 3. as well as 2. - when really I want 4. =)

**To Reproduce**

**Expected behavior**

**Screenshots**

**Client details:**
 - OS: Linux
 - Browser: Firefox or Chrome

## Server details

**Nextcloud version:**
29.0.7.2

**Version of the richdocuments app**
"8.4.6"

**Version of Collabora Online**

COOLWSD version: 24.04.8.2snapshot(git hash: 46983cf (E))
LOKit version: Collabora Office 24.04.8.20241020(git hash: 9591bc5)
Served by: Ubuntu 22.04.5 LTS
Server ID: db11fee2
WOPI host: 

</details>

max-nextcloud:
Dear @mmeeks 
I've been trying to reproduce this for a while and also looked into possible causes with @elzody . The cause we suspected has already been fixed.

Are you still observing this?
Could you try and capture a console log and network HAR file when this happens?

max-nextcloud:
> Can you link to the fix - would love to see what it was, when I debugged this some years ago it seemed like it was a plugin registration / priority order and some randomly ordered initialization.

That's pretty much spot on: https://github.com/nextcloud/viewer/pull/2486 

I'm checking if the fix was actually backported for link shares as well - or rather up to which version.

max-nextcloud:
I'm trying to figure out if this can still be observed somewhere. If it's still happening I will look into writing a debugging patch that at least logs what's going on.

mmeeks:
This re-occured on our share server the next day - which I believe is a reasonably recent version I think 30.something

max-nextcloud:
It looks like this is caused by the richdocuments script taking too long to register the handler for the file extension with the viewer. When the file is opened before richdocuments registered the handler it will be downloaded instead.

[...]

Issue #3815, opened 2024-07-09
Files sometimes do not get opened immediately on internal links


juliusknorr:
### Steps to reproduce
1. Have an internal link like /f/1234
2. Open it in a browser (e.g. from clicking in an email)

### Expected behaviour
The file should open

### Actual behaviour
Sometimes the file doesn't open automatically and there is an error thrown. It seems happening more often with cleared browser cache.

Now the odd thing here is that the error is thrown before the viewer app is initialized, so there seems to be some kind of timing issue when trying to get the file actions for the requested file.

https://github.com/nextcloud/server/pull/45586 would help with the error but I cannot see how it would solve the file then just not opening up.

I have not managed to reproduce this locally on any of the versions,  but tech-preview is affected (currently on 29.0.0) and another instance on 28.0.6 (both should be upgraded 🙈 )

@susnux Do you have any idea on why the order in which the `handleOpenFile` and the viewer registrations would be executed might differ?

<img width="1680" alt="Screenshot 2024-07-09 at 10 05 21" src="https://github.com/nextcloud/richdocuments/assets/3404133/ef09c7e4-417b-488b-bbf4-4fa5688ea917">
<img width="1680" alt="Screenshot 2024-07-09 at 10 06 15" src="https://github.com/nextcloud/richdocuments/assets/3404133/82a6db46-619a-4e45-aa51-36a5465e8e5c">

juliushaertl:
Upgraded the 29 instance to 29.0.3, still reproducible sometimes, but now without an error

pedropintosilva:
cc'ing as requested: @mmeeks @karlitschek

bentuna:
We are running 28.0.8 and can still confirm the issue is happening sometimes.

juliushaertl:
I had a chat with @susnux about this and this seems to be a design flaw of the viewer api registering the file type handlers too late. He will think about a possible workaround.

juliushaertl:
@susnux I looked a bit further into it while fixing a regression with 30 https://github.com/nextcloud/server/pull/47014

I still have not managed to trigger this race condition locally. I tried to think about a way to fake this timing issue but cannot figure out how I could manually delay the viewer registration like it is happening on those affected systems. But thought about a workaround like this, kind of deferring the handleOpenFile until the DOMContentLoaded event from viewer has registered the file actions.


```
diff --git a/apps/files/src/components/FilesListVirtual.vue b/apps/files/src/components/FilesListVirtual.vue
[...]

PR #4983, opened 2025-09-02
fix: register handler within init script

susnux:
* Resolves: https://github.com/nextcloud/richdocuments/issues/4168
* Resolves: https://github.com/nextcloud/richdocuments/issues/3815
* Target version: main

### Summary

Based on https://github.com/nextcloud/viewer/pull/2922

This should fix all the loading order issues, we now have an API package.
So we can register the handler in an init script and thus its always registered.

I cannot test this locally due to missing dev setup of Collabora, maybe you could help me testing @elzody ?

### Checklist

### DIFF
--- a/lib/Listener/LoadViewerListener.php
+++ b/lib/Listener/LoadViewerListener.php
@@ -11,6 +11,7 @@ declare(strict_types=1);
 
 namespace OCA\Richdocuments\Listener;
 
+use OCA\Richdocuments\AppInfo\Application;
 use OCA\Richdocuments\PermissionManager;
 use OCA\Richdocuments\Service\InitialStateService;
 use OCA\Viewer\Event\LoadViewer;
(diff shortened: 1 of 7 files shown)


susnux:
@luka-nextcloud do you plan to backport this? If yes then I will also backport the viewer

luka-nextcloud:
@susnux Could you please review backport prs?

mmeeks:
Wow this would be wonderful to have deployed - I'm still getting a ton of this ...

pedropintosilva:
Thanks for not only fixing the problem but also backporting it! @susnux or @luka-nextcloud could you please press the button to create a new release of richdocuments? Otherwise, and until a new version is published, users will continue experiencing these problems.

susnux:
> Otherwise, and until a new version is published, users will continue experiencing these problems.

It does not work either until next maintenance releases of Nextcloud server.

PR #4997, opened 2025-09-04
fix: backward compability

luka-nextcloud:
Fix backward compability


susnux (approved):
Works only until we reach a string that does not start with a number >= 3.
Probably better to either use `semver` for comparison or just split and compare

susnux (approved):

PR #4998, opened 2025-09-04
Fix backward compatibility

luka-nextcloud:
Fix backward compatibility


susnux (commented):
Hm not sure this is really hacky and only works until we e.g. reach version 1xx or similar happens :eyes:

susnux (approved):

elzody (approved):
I guess that's one way to do it

PR #5058, opened 2025-10-02
fix(SecureView): node cannot be found when within groupfolder

blizzz:
* Resolves: # https://github.com/nextcloud/groupfolders/issues/4010
* Target version: main

### Summary

The node cannot be properly figured out when the affected file stems from within a groupfolder. But we can use the filecache entry instead.

### TODO

- [X] consider refactoring `PermissionManager::shouldWatermark()` as it now accepts Node or ICacheEntry which makes it harder to work with. – #5074

### Checklist

### DIFF
--- a/lib/Storage/SecureViewWrapper.php
+++ b/lib/Storage/SecureViewWrapper.php
@@ -89,19 +89,25 @@ class SecureViewWrapper extends Wrapper {
 			fclose($fp);
 		}
 
-		try {
-			$node = $this->rootFolder->get($this->mountPoint . $path);
-		} catch (NotFoundException) {
-			// If the file is just created we may need to check the parent as this is only just about figuring out if it is a share
-			$node = $this->rootFolder->get(dirname($this->mountPoint . $path));
+		$storage = $sourceStorage ?? $this;
+		$cacheEntry = $storage->getCache()->get($path);
+		if (!$cacheEntry) {
+			$parent = dirname($path);
+			if ($parent === '.') {
+				$parent = '';
+			}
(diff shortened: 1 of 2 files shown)


lib/Storage/SecureViewWrapper.php
```diff
@@ -89,19 +89,25 @@ private function shouldSecure(string $path, ?IStorage $sourceStorage = null): bo
...
-		} catch (NotFoundException) {
-			// If the file is just created we may need to check the parent as this is only just about figuring out if it is a share
-			$node = $this->rootFolder->get(dirname($this->mountPoint . $path));
+		$storage = $sourceStorage ?? $this;
+		$cacheEntry = $storage->getCache()->get($path);
+		if (!$cacheEntry) {
+			$parent = dirname($path);
+			if ($parent === '' || $parent === '.') {
+				throw new NotFoundException(sprintf('Could not find cache entry for path of %s within storage %s ', $path, $storage->getId()));
+			}
```
icewind1991, line 98: I don't think is correct, I think that would then fail when trying to move a file to the root of a storage.

```php
if ($parent === '.') {
     $parent = '';
}
```

should be enough to handle the edge case

blizzz: adjusted to your suggestions anyways

come-nc (approved):

icewind1991 (approved):

AndyScherzinger:
Does this apply to v30 as well? @blizzz ?

blizzz:
> Does this apply to v30 as well? @blizzz ?

Nope.

PR #5059, opened 2025-10-02
Refresh files list after exporting a file in a different format

silverkszlo:
### Before
When the user exports a document as PDF or EPUB from Collabora Online OR uses "save as" to save the file with a different format or a new name, the new file did not appear automatically in their files list. Only after a manual page reload was the file displayed.

### Now
The new file appears in the files' list without needing a reload of the page.

The "Save As" and "Export As" events are being handled by tracking the requested filename when the user initiates the operation. When Collabora confirms the save/export is complete, the app fetches the new file's metadata via WebDAV and emits a `files:node:created` event, causing the Files app to display the new file in the file list immediately without requiring a page reload.

### Checklist

### DIFF
--- a/src/view/FilesAppIntegration.js
+++ b/src/view/FilesAppIntegration.js
@@ -550,23 +550,19 @@ export default {
 		}
 
 		try {
-			// Query the directory containing the file, not the file itself
+			const path = `${this.filePath}/${basename}`
 			const client = davGetClient()
-			const results = await client.getDirectoryContents(`${davRootPath}${this.filePath}`, {
+			const results = await client.getDirectoryContents(`${davRootPath}${path}`, {
 				details: true,
 				data: davGetDefaultPropfind(),
 			})
-
-			// Find the specific file in the directory listing
-			const nodes = results.data
-				.map((result) => davResultToNode(result))
-				.filter((node) => node.basename === basename)
+			const nodes = results.data.map((result) => davResultToNode(result))
 
 			if (nodes[0]) {
 				console.debug('[FilesAppIntegration] Emitting files:node:created for', basename)
 				emit('files:node:created', nodes[0])
 			} else {
-				console.warn('[FilesAppIntegration] New file not found in directory:', basename)
+				console.warn('[FilesAppIntegration] New file not found:', basename)
 			}
 		} catch (e) {
 			console.error('Failed to fetch new file metadata from webdav', e)


elzody:
I think there is probably a better way to handle this, like via post message. Clicking on the Save As or Export As buttons in Collabora sends the `UI_SaveAs` post message to us, which we then handle by displaying the Save As dialog on our side. In response, we send the `Action_SaveAs` post message, which has a `Notify` argument that tells Collabora whether or not we want to know when the operation has finished and what the result is.

My idea was to use this to then trigger a rebuild of the files list (probably via some event bus message), but we only get the `Action_Save_Resp` post message back from Collabora when the user carries out the task from the Save As button -- not from the Export As button.

This doesn't make sense to me because if it's following the same flow under the hood, 1) why have two separate buttons for it and 2) why does doing this via the Export As button not also notify us when the operation is complete

I will clarify this behavior with them before we proceed

juliusknorr:
Yes, I think we should rather rely on existing files app methods to add the [new file or update it through events](https://github.com/nextcloud/richdocuments/blame/6e9ab482d24b667b5adebbb669a8e09081a5614f/src/view/FilesAppIntegration.js#L494-L504). Reload was possible and used in the past but had quite some performance impact to always refetch the full file list.

elzody:
Collabora has confirmed that the Export As button not sending back an `Action_Save_Resp` is a bug, so I think we should wait for that to be fixed and then implement better handling for this using the methods Julius highlighted.

juliusknorr:
https://github.com/CollaboraOnline/online/pull/13309 as upstream fix to be tested with

juliusknorr:
@elzody I rememeber you wanted to give the collabora fix a try? Could you give an update on that part?

elzody:
@juliusknorr @silverkszlo Indeed, the fix on Collabora's side works. It seems to send the correct PostMessage now, so we should be able to proceed with our fix. I am happy to help any way that I can

silverkszlo:
> @juliusknorr @silverkszlo Indeed, the fix on Collabora's side works. It seems to send the correct PostMessage now, so we should be able to proceed with our fix. I am happy to help any way that I can

Ok so I couldn't just use `updateFileInfo` after performing "Save as" or "Export as" as it updates metadata for the current file that's already open (name, modification time). In order to see the new file (either saved with a different format, name or exported with a different format) without reloading the page, I implemented `createNodeForNewFile`. 
- It queries the current directory via WebDAV, 
- finds the newly created file by its basename, 
- and emits a `files:node:created` event.  This event triggers the Files app to add the new file to the visible file list without requiring a page reload.

This works for both SaveAs and ExportAs now.

juliusknorr:
Backports?

PR #5072, opened 2025-10-06
zotero: remove zotero API key textbox

lpranam:
COOL is now capable of handling and storing zotero key This PR only removes the textbox because the actual zotero key will be reused and transferred to COOL

removing all the zotero fields will require user to reenter the key 

relevant COOL PRs:
https://github.com/CollaboraOnline/online/pull/12694 
https://github.com/CollaboraOnline/online/pull/13091

* Target version: main

### Checklist

### DIFF
--- a/src/components/PersonalSettings.vue
+++ b/src/components/PersonalSettings.vue
@@ -34,47 +34,6 @@
 			</em>
 		</p>
 
-		<!-- Zotero -->
-		<div class="zotero-section">
-			<p><strong>{{ t('richdocuments', 'Zotero') }}</strong></p>
-			<template v-if="hasZoteroSupport">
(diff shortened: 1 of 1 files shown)


juliusknorr:
Is there a migration path to still pick up the existing api keys? If not there is more code to cleanup that handles the key on the backend side I think.

lpranam:
https://github.com/CollaboraOnline/online/pull/12913 takes care of key migration

lpranam:
> Is there a migration path to still pick up the existing api keys? If not there is more code to cleanup that handles the key on the backend side I think.

Yes we reuse the existing if whenever possible so user doesn't have to bother with all these changes.

Issue #4184, opened 2024-10-30
User should able to mention themself


Rash419:
**Describe the bug**
The autocomplete list the COOL gets from richdocuments doesn't have username of user typing the mention. Therefore, they can't mention themselves

**To Reproduce**
Try to mention yourself in the document

**Expected behavior**
User should able to mention themself

juliusknorr:
@elzody Maybe a quick one to fix if you are working in that area anyways right now?

elzody:
I've discovered that the currently logged-in user is removed from the search results from the autocomplete API, unless you specifically search for the entire UID. For example, if logged-in as user `Jane`, and you hit the autocomplete API with `jan`, you will receive nothing. But, if you search for the entire name `jane` you will get Jane's user. But, if logged-in as another user, say `Bob`, he can search for `jan` just fine and receive `jane` as a result.

Not sure how to bypass this, but for now, entering your entire username should work.

juliusknorr:
In tables we just manually add the current user as a workaround. Should be a quick fix for richdocuments as well with that:

https://github.com/nextcloud/tables/pull/1756

PR #5115, opened 2025-10-19
fix self-mention in comments

rikled:
* Resolves: #4184
* Target version: main

### Summary

This workaround fixes the self-mention issue described in #4184 until https://github.com/nextcloud/server/issues/48180 is implemented. The current user is now allowed to mention himself in the comments.

### DIFF
--- a/src/mixins/uiMention.js
+++ b/src/mixins/uiMention.js
@@ -7,6 +7,7 @@ import { generateOcsUrl } from '@nextcloud/router'
 import axios from '@nextcloud/axios'
 import Config from '../services/config.tsx'
 import { getNextcloudUrl } from '../helpers/url.js'
+import { getCurrentUser } from '@nextcloud/auth'
 
 export default {
 	methods: {
@@ -24,6 +25,18 @@ export default {
 					}
 				}
 
+				// Workaround to add the current user to the suggestions if they are not already in the list
+				// https://github.com/nextcloud/server/issues/48180
+				const currentUser = getCurrentUser()
+				const userExists = users.some(user => user.id === currentUser.uid)
+
+				if (!userExists) {
+					const matchesSearch = currentUser.uid.toLowerCase().includes(text.toLowerCase()) || currentUser.displayName.toLowerCase().includes(text.toLowerCase())
+					if (matchesSearch) {
+						users.push({ id: currentUser.uid, label: currentUser.displayName })
+					}
+				}
+
 				const list = users.map((user) => {
 					const profile = window.location.protocol + '//' + getNextcloudUrl() + '/index.php/u/' + user.id
 					return {
@@ -32,7 +45,6 @@ export default {
 						profile,
 					}
 				})
-
 				this.sendPostMessage('Action_Mention', { list })
 			}
 			if (type === 'selected') {


juliusknorr:
Thanks a lot, I've restarted cypress as it seemed like an unrelated failure. I'll have a closer look later today

src/mixins/uiMention.js
```diff
@@ -24,6 +25,13 @@ export default {
 					}
 				}
 
+				// Workaround to add the current user to the suggestions if they are not already in the list+
```
juliusknorr:
```suggestion
				// Workaround to add the current user to the suggestions if they are not already in the list
```

src/mixins/uiMention.js
```diff
@@ -24,6 +25,13 @@ export default {
...
+				// https://github.com/nextcloud/server/issues/48180
+				const currentUser = getCurrentUser()
+				if (!users.some(user => user.id === currentUser.uid) && currentUser.uid.includes(text)) {
```
juliusknorr: We should also check the displayname so you can find yourself in the matches, can you adjust that? With some user backends like LDAP the user id could just be a UUID

rikled:
Thank you very much for the comments. Right, checking for the display name makes much more sense, I’ve added that. I also had a quick look at the server-side implementation, and it seems that UID checks are exact match checks there (which might explain why exact match actually works if uid===displayname). So to be consistent, maybe it makes sense to change the UID check to an exact match check against the input text? 

Also, please let me know if I should add some unit tests (or even Cypress?). I couldn’t find a unit test setup in the project, so I skipped it this weekend due to time constraints (and some frustration with my currently very slow development setup).

juliusknorr:
Sorry for the delayed response and thanks for checking this further in depth. I think the current approach implemented is good for this. I think not limiting uid check is more suitable, also comparing how for example Talk does the matches against the uid. Though I see it is not consistent with the autocomplete endpoint.

So far we only have cypress tests for the frontend code parts in this repo, however I fear those would be to much dependent on the implementation within Collabora Online, so hesitant adding them. It might be nice to add vitest to this repo and have some actual unit tests for parts like this where the full integration would be overkill, but seems to exceed the scope of this PR.

juliusknorr:
Just one minor bit that I just noticed when giving this another test run, we probably should check case-insensitive?

rikled:
I updated the check to be case-insensitive. The failing tests appear to be unrelated again, but I’ll take a closer look to confirm.

juliusknorr:
Yes, those tests seem to also fail on main, so agree that they are unrelated https://github.com/nextcloud/richdocuments/actions/runs/18894277933

juliusknorr:
Mind to squash your commits into one and rebase them to current main? Then it is good to go from my side 🚀

juliusknorr:
With https://github.com/nextcloud/richdocuments/pull/5130 the tests should also pass again after the rebase

rikled:
I’ve squashed all commits into one and rebased onto the latest main. All tests are passing again. Thanks for your constructive feedback!

juliusknorr:
Thanks a lot 🤝

Issue #5170, opened 2025-09-10
better warning when trying to download a file that is restrictied via secure view


alexanderdd:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
1. configure secure view in admin settings > office
2. in Files, click on ... of a file, click download
3. see 403 Forbidden browser page

**Describe the solution you'd like**
Show a warning like "cannot be downloaded due to secure view settings"

provokateurin:
This is rather a bug than a feature request.

sebastiaanveld:
This issue breaks [Sendent](https://sendent.com/sendent-for-outlook/) integration in Nextcloud; when a PDF file is shared using the Sendent Outlook plugin this file gets uploaded to a folder in Nextcloud with read-only access so only a view and download option are available for the user receiving the share email.

If Secure View is enabled in the Office App this breaks the ability to download the shared PDF file from Nextcloud.
If I zip the pdf and share that the same way the user can download the PDF ZIP without a problem.

Mind I would like to use both the Secure View option on certain files and folders but also be able to share a file which a user is able to download.

Also see:
https://github.com/nextcloud/richdocuments/issues/4944

Secure View:

<img width="586" height="288" alt="Image" src="https://github.com/user-attachments/assets/670b9d08-b720-45e3-b835-fe4771dd15e5" />

blizzz:
The download options should not be shown in that case. Already reached out to @susnux, as it is possible with shares… just need to enable other apps to have influence there.

susnux:
I think office need to extend the node then (the DAV node) and provide a webdav property like `nc:secure-view` so that the files app can respect it.

blizzz:
I suppose then we need a dav plugin in richdocuments that does something like https://github.com/nextcloud/server/blob/master/apps/dav/lib/Connector/Sabre/FilesPlugin.php#L422-L433

PR #5171, opened 2025-11-13
fix(SecureView): hide disfunctional *download* files action

blizzz:
This is achieved by setting a specific DAV attribute. At the moment there is one handler in dav-apps FilesPlugin and it could overwrite the value with "false". We make sure not to downgrade here and prevent downgrade from dav (possible race condition).

* Resolves: #5170 
* Target version: stable31

### TODO

- [x] remove code duplication by moving affected code into a SecureView service class.

### Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 
 namespace OCA\Richdocuments\AppInfo;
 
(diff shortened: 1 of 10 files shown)


blizzz:
As discussed with @juliusknorr we'd take it like this (provided reviews are fine) and follow up later on with a better solution across NC. PR comes with tests :)

elzody (approved):
Seems to work fine as it is after some quick testing locally

PR #5196, opened 2025-11-23
[main] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency


elzody:
The CI is failing because https://github.com/nextcloud/richdocuments/pull/5256 is needed. After that is merged, this branch can be rebased onto `main` and we can see if it passes.

elzody:
Merging as the remaining test failures are not believed to be related.

PR #5208, opened 2025-11-27
fix(mobile): pass format to SaveAs

Minion3665:
In the mobile app, `args.format` was typoed as `args.Format`, leading to format always being undefined. This meant that the save as popup would use the extension of the current file rather than the extension you were trying to save as. You could workaround by editing this manually, but it was annoying and not working as intended...

The browser version wasn't broken, as it uses a different code path to save documents (see src/mixins/saveAs.js)

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/cypress/e2e/direct.spec.js
+++ b/cypress/e2e/direct.spec.js
@@ -183,12 +183,7 @@ describe('Direct editing (legacy)', function() {
 					.should('be.visible')
 				cy.get('.saveas-dialog input[type=text]')
 					.should('be.visible')
-					.should('have.value', 'document.odt')
-
-				cy.get('.saveas-dialog input[type=text]')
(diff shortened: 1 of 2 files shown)


Minion3665:
Multiple force pushes to fix tests and update commit messages, cumulatively
image

Minion3665:
@juliusknorr I added myself to the blocklist a while back, do you mind taking a look at why I still got the bot comment here? Thanks!

juliusknorr:
Indeed, it seems casing matters here: https://github.com/nextcloud/.github/pull/649

Minion3665:
ahhh 🤦🏼‍♀️ - thanks very much for checking that for me!

PR #5233, opened 2025-12-15
Digital Signature: Hide digital signature inputs from user settings

codewithvk:
* Resolves: 
* Target version: main

### Summary

Digital Signature: Hide digital signature inputs from user settings

Since Collabora Online already supports digital signatures via the settings iframe, this patch is intended to only hide the UI inputs. The related variables will still be sent to Online through checkfileinfo, and migration cases will be handled as well.

Online PR:  https://github.com/CollaboraOnline/online/pull/12913


juliusknorr:
Since which version does Collabora show this in the settings iframe? Should we only hide it in case we know this is present?

jazevedo-coll:
> Since which version does Collabora show this in the settings iframe? Should we only hide it in case we know this is present?

Hi @juliusknorr this is already available for CODE since 25.04.8-1
As for COOL it will be included in  25.04.8-3 which is to be released this month.

juliusknorr:
Can you elaborate a bit further how the migration works? Can we force a migration on the Nextcloud side somehow? I fear that otherwise we may always have to keep the old logic on the backend side.

codewithvk:
> Can you elaborate a bit further how the migration works? Can we force a migration on the Nextcloud side somehow? I fear that otherwise we may always have to keep the old logic on the backend side.

So currently on the online side, if Nextcloud WOPI CheckFileInfo has any values, they will override the new JSON file (created on the COOL side), and that will always be referenced for the new values.
Ref: https://github.com/CollaboraOnline/online/pull/12913/changes#diff-1346662577e72b9ac23f0977699d096e506badf1ce7d2cf56a84ab8769d9be66R1728-R1760

In terms of backend logic, I don't have many plans right now, as we might need to fully migrate it in future versions. For this PR, I'm just removing the logic from the client side.

For now let's remove from UI so user can't see two inputs for same fileds.

juliusknorr (approved):

elzody (approved):

codewithvk:
Can we merge this PR?

PR #5234, opened 2025-12-16
feat: use filetype icon as favicon

juliusknorr:
Screenshot 2025-12-16 at 14 31 30

As discussed with @jancborchardt and @pedropintosilva 

@jancborchardt I think the existing icons work very well, but we can still change to plain background color and white icons if you think that is better.

### DIFF
--- a/src/view/Office.vue
+++ b/src/view/Office.vue
@@ -104,6 +104,7 @@ import axios from '@nextcloud/axios'
 import {
 	generateUrl,
 	imagePath,
+	generateFilePath,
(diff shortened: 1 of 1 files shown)


juliusknorr:
The missing icon in the collabora UI seems an issue on latest CODE

jancborchardt (approved):
Yeah @juliusknorr this looks super nice and is probably easier to recognize like this as opposed to if we would do the coloring the other way around. Cool! :)

marcoambrosini (approved):

PR #5282, opened 2026-01-22
feat(deps): add Nextcloud 34 support on main

elzody:
* Target version: main

### Summary
Now that the branch-off has occurred, we need to add Nextcloud 34 support on the `main` branch and bump the version to the next major version of the app. It did not happen automatically, so this PR does it manually based on the changes from the previous branch-off: https://github.com/nextcloud/richdocuments/pull/4996/changes


elzody:
I don't believe the failing Cypress tests to be related. They can be fixed in other PRs.

PR #5293, opened 2026-01-27
fix: Emit files:node:updated event on UI_Save

silverkszlo:
* Resolves: #5267 
* Target version: main

### Summary
Handle `UI_Save` post message to update file info and emit `files:node:updated` event, allowing other components like the versions sidebar to react to document saves.

### Checklist

### DIFF
--- a/src/view/Office.vue
+++ b/src/view/Office.vue
@@ -513,6 +513,9 @@ export default {
 					FilesAppIntegration.updateFileInfo(undefined, Date.now())
 				}
 				break
+			case 'UI_Save':
+				FilesAppIntegration.updateFileInfo(undefined, Date.now())
+				break
 			case 'Clicked_Button':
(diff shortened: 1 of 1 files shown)


juliusknorr:
I'm unsure if this only catches the manual save through the menu bar button. For autosave this could be interesting as well, but from the post message documentation i do not see any hint that there is such a message.

There is Action_Save_Resp but I'm unsure if that will get triggered. Nothing blocking, but something to check and file with collabora as possible follow up I'd say.

elzody (approved):

elzody:
The failing tests should hopefully pass once https://github.com/nextcloud/richdocuments/pull/5312 is merged.

silverkszlo:
Failing cypress tests seem unrelated, so I'm merging this

PR #5294, opened 2026-01-27
fix: Properly handle sidebar in 33

juliusknorr:
Removes old API calls and replaces with the new one from `@nextcloud/files`

### DIFF
--- a/src/view/FilesAppIntegration.js
+++ b/src/view/FilesAppIntegration.js
@@ -11,6 +11,9 @@ import axios from '@nextcloud/axios'
 import { emit } from '@nextcloud/event-bus'
 import { getCurrentDirectory } from '../helpers/filesApp.js'
 import {
+	getSidebar
+} from '@nextcloud/files'
+import {
 	getClient,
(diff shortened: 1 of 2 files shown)


juliusknorr:
Reported the talk failures to the talk team, does not look related to our code base

juliusknorr:
I amended a fix for the versions sidebar

elzody:
Still getting a console error that might affect the versions tab test.
> Error: Cannot set sidebar tab 'files_versions' because it is not available for the current context.

I don't think it's anything wrong here though, seems like something else is going on? As far as I can tell it should have the right context since it's being set directly before.

juliusknorr:
Strange, it worked in my local test

juliusknorr:
Strange, it worked once, but now i also see this locally

juliusknorr:
Hotfix added and reported at https://github.com/nextcloud-libraries/nextcloud-files/issues/1474

juliusknorr:
Well no, different story. I failed to resolve a promise for the node object 🙈

PR #5308, opened 2026-01-29
fix(psalm): use template response constant

elzody:
* Target version: main

### Summary
Uses the `TemplateResponse::RENDER_AS_BLANK` constant since `blank` is not accepted.

### Checklist

- [x] Code is properly formatted
- [x] Sign-off message is added to all commits
- [x] Documentation (manuals or wiki) has been updated or is not required

PR #5309, opened 2026-01-29
chore: Fix invalid render as for settings templates

juliusknorr:
Fixes psalm failure since https://github.com/nextcloud/server/commit/46f87f6e46e89572476e238aebac0c483199efb0


juliusknorr:
All right @elzody was faster https://github.com/nextcloud/richdocuments/pull/5308

PR #5319, opened 2026-01-31
chore: switch to php-parallel-lint/php-parallel-lint

joshtrichards:
* Resolves: # 
* Target version: main

### Summary

- Fixes the unmaintained package warning during `composer install` runs
- Updates to a php-parallel-lint version that runs on current PHP versions
- Switches `composer lint` to actually use it

### TODO

- [ ] ...

### Checklist


juliusknorr:
Pushed another fix to add composer install to the github action, probably something to upstream to https://github.com/nextcloud/.github/blob/master/workflow-templates/lint-php.yml

PR #5320, opened 2026-01-31
test: add unit tests for CachedRequestService

joshtrichards:
* Target version: main

### Summary

Adds basic tests for the abstract `CachedRequestService` class. Since the class cannot be instantiated directly, a dedicated test double is defined within the test file.

### Checklist


juliusknorr:
I fixed the php-cs-fixer complaint

PR #5338, opened 2026-02-05
[stable33] fix: Use first uploader element

elzody:
Manual back-port of https://github.com/nextcloud/richdocuments/pull/5302


elzody:
Merging. The failures do not seem to be related and can be fixed with further back-ports

PR #5366, opened 2026-02-13
fix: Register default file action to prevent download instead of open

timar:
The Viewer app's 'view' file action checks window.OCA.Viewer.mimetypes to decide if it can handle a file. However, those mimetypes are only populated on DOMContentLoaded when handlers are bridged from the @nextcloud/viewer npm package to the ViewerService. If the Files app renders the file list before that event fires, the 'view' action's enabled() returns false and the 'download' action (order 30) becomes the default — causing document files to be downloaded instead of opened in Collabora Online.

Register a richdocuments-specific default file action that checks against the OCS capability mimetype list (baked into the page HTML via initial state, always available immediately) instead of the ViewerService's runtime mimetypes. When the Viewer's 'view' action works correctly both actions are enabled but 'view' wins (registered first). When 'view' is disabled due to the timing race, the new action takes over. Execution delegates to the Viewer's own 'view' action for proper URL history and prev/next navigation support.

### DIFF
--- a/src/file-actions.js
+++ b/src/file-actions.js
@@ -2,12 +2,68 @@
  * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
-import { registerFileAction } from '@nextcloud/files'
+import { DefaultType, Permission, getFileActions, registerFileAction } from '@nextcloud/files'
 import { getCapabilities } from './services/capabilities.ts'
 import { translate as t } from '@nextcloud/l10n'
 
 // eslint-disable-next-line import/no-unresolved
 import appIcon from '../img/app.svg?raw'
+
+const supportedMimes = getCapabilities()?.mimetypes ?? []
+
+// Default file action for document types handled by richdocuments.
+// The Viewer app registers a generic 'view' file action that checks
+// window.OCA.Viewer.mimetypes, but those mimetypes are only populated
+// on DOMContentLoaded which can race with the file list rendering.
+// This action uses richdocuments' own capability list (available
+// immediately from OCS initial state) so it is not affected by timing.
+const openDocument = {
+	id: 'richdocuments-default',
+
+	displayName: () => {
+		return t('richdocuments',
+			'Open in {productName}',
+			{ productName: getCapabilities()?.productName ?? 'Nextcloud Office' })
+	},
+
+	iconSvgInline: () => {
+		return appIcon.replaceAll(/#(fff|0{6})/g, 'currentColor')
+	},
+
+	default: DefaultType.DEFAULT,
+
+	enabled: ({ nodes }) => {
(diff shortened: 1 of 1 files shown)


juliusknorr:
This approach seems more like a workaround. Can you tell how you replicated this and if it was specific to a nextcloud version? Last week https://github.com/nextcloud/richdocuments/pull/5355 was raised as we missed that backport for 32 it seemed. https://github.com/nextcloud/richdocuments/pull/4983 was the pr for master and is supposed to fix exactly the issue you mentioned.

timar:
> This approach seems more like a workaround. Can you tell how you replicated this and if it was specific to a nextcloud version? Last week #5355 was raised as we missed that backport for 32 it seemed. #4983 was the pr for master and is supposed to fix exactly the issue you mentioned.

Yes, it's sort of a hack. I tried to fix the underlying timing issue in viewer, but I failed. I'm happy if you fixed it properly in a different way. We always had this issue with different versions, but I debugged and tested my fix with Nextcloud 33.0.0 RC4 and richdocuments 10.0.0-beta.1.
It's not 100% reproducible, sometimes when you load the Files view, the click on files download them instead of opening in COOL. The bug disappears, if we enter a subdirectory. It also disappears, if we reload the page 1-4 times. And sometimes file opening works immediately.
With this hack I submitted, the bug was 100% not reproducible.

juliusknorr:
@timar I'd like to further replicate this, can you share the exact steps you took to reproduce? I did various ~30 attempts of reload the page, with just a folder view or also a link to a file but failed with my dev setup.

timar:
> @timar I'd like to further replicate this, can you share the exact steps you took to reproduce? I did various ~30 attempts of reload the page, with just a folder view or also a link to a file but failed with my dev setup.

@juliusknorr you have access to https://staging-perf.eu.collaboraonline.com/nextcloud/ (juliushaertl) and there you can reproduce the issue for sure. I saw it today 2x times.
From the root folder of your Files view try to open a document.

juliusknorr:
I managed to replicate this and see the following:

```
[DEBUG] viewer: OCA.Viewer initialized 
BREAKPOINT TRIGGERS on validating the viewer file action
[INFO] viewer: 6 viewer handlers registered 
```

In cases where it works the order is

```
[DEBUG] viewer: OCA.Viewer initialized 
[INFO] viewer: 6 viewer handlers registered 
BREAKPOINT TRIGGERS on validating the viewer file action
```

I think this part may be loaded too late https://github.com/nextcloud/viewer/blob/c73fa3281db8bd899e01be77fae6a9950193897c/src/views/Viewer.vue#L521-L543

But I cannot tell why the domcontentloaded sometimes triggers later yet. 

I have not been able to replicate this locally yet though. 

I did a somewhat blind fix draft at https://github.com/nextcloud/viewer/pull/3080 still need to verify

juliusknorr:
@timar If you were able to reproduce locally, would you be available to try out the viewer pr?

PR #5367, opened 2026-02-13
fix: Remove hasDrawSupport check that strips .odg mimetypes

timar:
hasDrawSupport() checked whether Collabora was at least version 6.4.7. When the Collabora capabilities cache was empty (between background job refreshes), productVersion defaulted to '0.0.0.0', causing the check to strip .odg and .odg-flat-xml mimetypes from the OCS capabilities. This prevented .odg files from being opened in the viewer.

Collabora 6.4.7 was released years ago and no supported Nextcloud version ships with anything older — remove the check entirely.


elzody (approved):

elzody:
Thanks a lot for handling this!

PR #5373, opened 2026-02-17
chore: Use own jquery to remove it again later

juliusknorr:
I'll revert it it again in https://github.com/nextcloud/richdocuments/pull/5311 but for now this should make CI happy again after https://github.com/nextcloud/server/pull/57648


elzody:
Looking forward to the day we can drop it entirely :)

PR #5392, opened 2026-02-18
fix(cypress): use stable33 server for cypress tests

elzody:
* Target version: stable33

### Summary
This was causing tests to fail regarding the jQuery change in 34, but we should use the stable33 server branch here because this is... well... stable33.

### Checklist


elzody:
The WOPI integration test failures should be fixed by https://github.com/nextcloud/richdocuments/pull/5391, and the Cypress failure is unrelated and needs to be looked into in a separate PR.

PR #5394, opened 2026-02-18
fix(migration): use doctrine string type

elzody:
* Target version: main

### Summary
Fixes an error when the migration is executed, as we need to use the proper column type.

### Checklist


juliusknorr:
Manually cherry picked to https://github.com/nextcloud/richdocuments/pull/5378 and https://github.com/nextcloud/richdocuments/pull/5379

PR #5451, opened 2026-03-02
build(nix): add flake for development shell

elzody:
* Target version: main

### Summary
I have been using and learning Nix, and I thought some good practice would be to create a flake. The main goal right now is to be able to enter a development shell that already has the correct versions of Node, NPM, PHP, and other dependencies set up and ready to use. It is more reproducible and cleaner than using things like NVM (Node Version Manager).

### TODO

- [ ] Clean up version string and package name logic
- [ ] Add Nix formatter to the flake
- [ ] Tests somehow?

### Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,11 +7,24 @@
     pkgs = import nixpkgs { inherit system; };
     system = "x86_64-linux";
 
-    nodeVersion = builtins.replaceStrings ["^"] [""] (builtins.fromJSON (builtins.readFile ./package.json)).engines.node;
+    getVersion = versionString: builtins.splitVersion (builtins.replaceStrings ["^"] [""] versionString);
+
+    node = rec {
+      version = (getVersion (builtins.fromJSON (builtins.readFile ./package.json)).engines.node);
+      pkgName = "nodejs_${builtins.elemAt version 0}";
+    };
+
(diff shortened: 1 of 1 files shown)


provokateurin:
Cool! For Nextcloud apps it's not necessary to have their own flakes though (at least if they don't do anything special), because we have https://github.com/nextcloud/server/blob/master/flake.nix. As long as your app is located somewhere inside the server sources direnv will check recursively and automatically pick it up. You just need to make sure that the server and this repo are "in sync" regarding stable branches, so that the right versions are used.

elzody:
@provokateurin Makes sense. I am wondering if we could at least use a flake to automatically bring up a Collabora server alongside the app and use it to run Cypress tests or something. Would that make sense or am I getting ahead of myself?

provokateurin:
That's not so easy (because you'd need some kind of VM and the flake dev shell doesn't provide that), but you should take a look at https://codeberg.org/icewind/haze. It can spawn a Nextcloud instance and a Collabora instance and it's also included in the server flake directly.

elzody:
@provokateurin Oh wow, that is really cool. Thanks a lot for teaching me a bit more; it looks like it solves all my problems already.

PR #5468, opened 2026-03-09
fix(a11y): Add accessKey to Open in local editor button

codewithvk:
* Resolves: # 
* Target version: main

### Summary
image
Added `2` as `accessKey` for button.


juliusknorr (approved):

juliusknorr:
Does this need a backport to a specific version?

pedropintosilva:
@juliusknorr I think it suffices if latest ( richdocuments `'10.1.0'`) gets a new release

PR #5492, opened 2026-03-18
fix: Log exceptions during fetching templates

juliusknorr:
Signed-off-by: Julius Knorr <jus@bitgrid.net>

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/TemplateManager.php
+++ b/lib/TemplateManager.php
@@ -107,6 +107,51 @@ class TemplateManager {
 		$this->userId = $userId;
 	}
 
+	private function debugStorage($file) {
+		$storage = $file->getStorage();
+
+		while ($storage) {
+			$class = get_class($storage);
+			$this->logger->warning(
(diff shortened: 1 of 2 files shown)


lib/TemplateManager.php
```diff
@@ -119,6 +164,18 @@ public function get(int $fileId) {
 
 		foreach ($files as $file) {
 			if ($file->getId() === $fileId) {
+				$this->logger->warning(
+					'DEBUG: found template in empty_templates dir: {fullPath} ({internalPath}) with permissions {permissions}, double-check is it readable: {isReadable}',
```
joshtrichards, line 168: Did you mean for these to be included and/or logged at `WARN` level?

joshtrichards (commented):
The changes in `SettingController` make sense to me, but the `TemplateManager` ones seem like overkill (particularly if logged at `WARN` level).

PR #5512, opened 2026-03-23
templateManager: allow using regular document files as templates

### LINKED ISSUE #5277: Global Templates allows only for file types which can't be created using Collabora Online
**Describe the bug**
When uploading a template in `Settings > Administration > Office > Global Templates`, the upload file open dialogue only allows for actual template file types _(ott, dotx, ...)_ by default. But Collabora Online doesn't allow creating template files like ott, dotx, ...

Normal office files _(odt, docx, ...)_ can only be uploaded, if the user manually selects _"All files"_ in the open file dialogue. And indeed, using a standard ODT file _(not an artificially renamed OTT)_ works fine for use in the Global Templates setting.

This is not a huge bug, but probably a minor inconvenience, which is very easy to fix _(easy hack)_. It probably only needs the other file types added to this upload open file dialogue list:  
https://github.com/nextcloud/richdocuments/blob/6324d6c84e6fd2e5c9364fe6169cbd7a9d6d5994/src/components/AdminSettings/GlobalTemplates.vue#L72  
Supporting saving+loading OTT files in Collabora Online seems a bit more complicated. So the suitable solution seems to simply allow for non-template odt, docx, ... files.

**To Reproduce**

Go to `Settings > Administration > Office > Global Templates` and try uploading an ODT.  
Result: only available when choosing `All files` which is irritating for admins.

Try saving an OTT using `Online Writer > File > Save As`.

**Expected behavior**
Global Templates should allow for file types by default, which can actually be created using Collabora Office.

## Server details

**Version of the richdocuments app**: `9.0.2`

**Version of Collabora Online**: `25.04.8.1.1`

## Other

Maybe related:
* `https://github.com/nextcloud/richdocuments/issues/2430`
  * _Default templates not working - blank.ott and blank.dotx_
* `https://github.com/nextcloud/richdocuments/issues/2315`
  * _Unable to open a template files (ott, ots, otp, otg) with Collabora Online / Nextcloud Office_

meven:
As Collabora Online does not support creating templates, allow the user to reuse documents created with richdocuments as templates.

Collabora Online supports thoe as templates.

Fixes: #5277

### DIFF
--- a/lib/TemplateManager.php
+++ b/lib/TemplateManager.php
@@ -506,7 +506,9 @@ class TemplateManager {
 	public function isSupportedTemplateSource(string $extension): bool {
 		$supportedExtensions = [
 			'ott', 'otg', 'otp', 'ots',
+			'odt', 'odg', 'odp', 'ods',
 			'dotx', 'xltx', 'potx',
+			'docx', 'xlsx', 'pptx',
 			'dot', 'xlt', 'pot',
 		];
 		return in_array($extension, $supportedExtensions, true);


elzody (approved):
Even though the issue you mentioned was already resolved with https://github.com/nextcloud/richdocuments/pull/5475, I think this change still makes sense. Don't forget to `--signoff` your commits so that the DCO passes.

```sh
git commit --amend --signoff --no-edit
```

It seems like the static analysis error is unrelated to this change.

meven:
> Even though the issue you mentioned was already resolved with #5475, I think this change still makes sense.

I somehow missed that, we have two codes paths to define which extensions are acceptable as templates now, It would be best to unify them I think rather than making them match.

juliusknorr (approved):

elzody:
> It would be best to unify them I think rather than making them match.

That makes sense. For now we can go ahead and merge this and then I will create a separate issue to track that.

meven:
I would need to rebase, I think, github does not let me rebase.
And I am guessing I can't merge myself into main.

Issue #4812, opened 2025-05-28
Switch to OOXML for improved interoperability


juliusknorr:
In recent discussions it turned out that it might be better to use OOXML as the default format to avoid interoperability issues with other software.

Admins can still change in favor of the open standard in the settings but we agreed with @pedropintosilva and @jancborchardt we should change the default

juliusknorr:
@grnd-alt Something you could pick up, might be a good first dive into some of the richdocuments code base bits

jdreo:
This is not a good idea. Microsoft has shown plenty of times that they are not honoring their own standard, and it’s obvious they do that to torpede interoperability. Going on for a MS-controled format for interoperability reasons is just giving up on interoperability.

nojhan:
There is a sad lack of political analysis with this move. OOXMl is not more interroperable, it is actually less. Only the FLOSS softwares actually implement it correctly, and only MS-Office does not. By making it a default, you’re giving more power to one single proprietary software, and less to several free software. This is not a good move.

nojhan:
LibreOffice made a statement about this kind of issue, targeting your main competitor (OnlyOffice), I hope the article can help you understand what’s at stake ang roll back this commit.
https://blog.documentfoundation.org/blog/2026/02/16/why-odf-and-not-ooxml/

julientaq:
Some more information about why german administration forces ODF over OOXML.
https://www.zdnet.fr/blogs/l-esprit-libre/allemagne-le-format-odf-obligatoire-dans-ladministration-492348.htm

(sorry it’s in french, but i’m pretty sure you’ll be able to get a translation nowadays)

pmiossec:
> Some more information about why german administration forces ODF over OOXML. https://www.zdnet.fr/blogs/l-esprit-libre/allemagne-le-format-odf-obligatoire-dans-ladministration-492348.htm
> 
> (sorry it’s in french, but i’m pretty sure you’ll be able to get a translation nowadays)

The source (from the 1st link of the article): https://blog.documentfoundation.org/blog/2026/03/19/germanys-sovereign-digital-stack-mandates-odf/

doc75:
This is really not a good move to change this default in my opinion. Indeed, it is well known that this « standard » was just created to try to convince other vendor to not switch to ODF. This is exactly what you are doing, going right were Microsoft wanted you to go 🤦‍♂️

mmuman:
[...]

PR #5516, opened 2026-03-25
fix(format): makes OpenDocument the default format

nojhan:
This reverts commit 8b464b6 by setting "od[fp]" instead of "ooxml" as a default.

Following several downvotes and arguments against making the Microsoft-controlled format OOXML a default. See the many comments on issue #4812 for getting more context.

Main arguments are:
- Nextcloud already pointed out the risk of letting big tech companies in charge:
> The big Tech companies are dominating every aspect of our lives and work, and we’re all getting tired of it. Their power is a threat to privacy, and, on a bigger scale, democracy and the sovereignty of entire countries. And for businesses, there is a real risk to resilience, compliance and security. [0](https://nextcloud.com/blog/nextcloud-hub26-winter/)

- The Document Foundation made a statement explaining why this is a bad idea (at the time, directed toward OnlyOffice) [1](https://blog.documentfoundation.org/blog/2026/02/16/why-odf-and-not-ooxml/)
- Germany is now requiring ODF (and forbids OOXML) in its administration [2](https://blog.documentfoundation.org/blog/2026/03/19/germanys-sovereign-digital-stack-mandates-odf/) and this has good chances to spread in Europe as well.
- Microsoft do not honor its own standard and bribed the ISO participants to have it make a competing standardized when OpenDocument was already established [3](https://www.linuxjournal.com/content/buy-cheat-steal-and-lie-ooxml-story) [4](https://www.wired.com/2007/08/microsoft-allegedly-bullies-and-bribes-to-make-office-an-international-standard/) [5](
https://effi.org/blog-kai-2007-09-05-en/)
- Microsoft is subject to American rule of law which can go as far as imposing sanctions to out-of-jurisdiction institutions through their software [6](https://www.whitehouse.gov/presidential-actions/2025/02/imposing-sanctions-on-the-international-criminal-court/) [7](https://www.computerweekly.com/opinion/Microsofts-ICC-email-block-reignites-European-data-sovereignty-concerns)

### DIFF
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -40,7 +40,7 @@ class AppConfig {
 		'watermark_allTagsList' => [],
 		'watermark_linkTagsList' => [],
 		'token_ttl' => 36000, // 10 hours
-		'doc_format' => 'ooxml',
+		'doc_format' => 'odf',
 	];
 
 	public const WATERMARK_APP_NAMESPACE = 'files';
--- a/lib/Listener/RegisterTemplateFileCreatorListener.php
+++ b/lib/Listener/RegisterTemplateFileCreatorListener.php
@@ -46,7 +46,7 @@ class RegisterTemplateFileCreatorListener implements IEventListener {
 		}
 
 		$templateManager = $event->getTemplateManager();
-		$ooxml = $this->config->getAppValue(Application::APPNAME, 'doc_format', 'ooxml') === 'ooxml';
+		$ooxml = $this->config->getAppValue(Application::APPNAME, 'doc_format', 'odf') === 'ooxml';
 		$appPath = $this->appManager->getAppPath('richdocuments');
 
 		$templateManager->registerTemplateFileCreator(function () use ($ooxml, $appPath) {
--- a/lib/Service/SlideDeckService.php
+++ b/lib/Service/SlideDeckService.php
@@ -79,7 +79,7 @@ EOF;
 	public function generateSlideDeck(?string $userId, string $presentationText, bool $includeWatermark = true) {
 		$rawModelOutput = $this->runLLMQuery($userId, $presentationText);
 
-		$ooxml = $this->config->getAppValue(Application::APPNAME, 'doc_format', 'ooxml') === 'ooxml';
+		$ooxml = $this->config->getAppValue(Application::APPNAME, 'doc_format', 'odp') === 'ooxml';
 		$format = $ooxml ? 'pptx' : 'odp';
 
 		try {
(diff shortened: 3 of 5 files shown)


juliusknorr:
Thanks for taking the time to open a pull request, we share the concerns and agree that ODF is the better long-term choice. However, in practice many users face interoperability issues with ODF, which often gets perceived as a problem with Nextcloud Office itself. Since OOXML is the de facto standard in many environments, using it as the default reduces friction and improves user experience out of the box.

Users can still switch back to ODF, and we strongly encourage that where possible. This is a pragmatic choice for now, not a shift in our overall stance, as mentioned in the old issue, we will reconsider changing this back in the future.

nojhan:
You want to comply with hypothetical complains from hypothetical users who would not be concerned with the feature anyway (we’re talking about a default on fresh installs); but you’re chosing to disappoint very real users and evangelists who are concerned and actively install & promote the app.

I insist you think about the general implications and reconsider before a backslash.

italovignoli:
The choice of defaulting to proprietary document formats is wrong in any case, but in the light of the recent publication of the Deutschland-Stack which defaults to ODF - and does not even mention OOXML - is even worse. A company based in Germany going against a decision from the German government and openly supporting Microsoft lock-in strategy is not a friend of open source but a friend of Microsoft and a supporter of US based Big Tech companies against the interests of European citizens. Congratulations !!!

PaoloVecchi:
> However, in practice many users face interoperability issues with ODF, which often gets perceived as a problem with Nextcloud Office itself.

I totally understand your point of view as the same objection is raised in relation to LibreOffice, most perceive LibreOffice as the problem, not the proprietary format used by a closed source monopolist.

> Since OOXML is the de facto standard in many environments
Well... asbestos used to be the de facto standard in many environments and it took a long time to get people to see that it wasn't at all good for them.

I agree that changing the default from a pseudo standard to an actual standard could lead some to think that there is an issue somewhere but with proper communication, which will also be supported by the recent choice of the German government, it would turn to an opportunity for Nextcloud to reach out to its current and future users to explain why having ODF as a default setting is good for them.

You can put a nice notice in the setup instructions and do nice blog posts about it and by doing so you will support others in creating awareness and promote positive change.

I hope Nextcloud management will see this as a very positive strategic decision.

nojhan:
I rapidly checked the issue log on Github, and there is no trace of those mentioned users complaining about the format. All we see is more problems with OOXML than with ODF (which the code reflects somehow, with a lot of conditional branching specifically for OOXML).

If the project page is a good proxy to the main users of a project, this does not tell the same story of an overwhelming problem with using an open format as the default.

This looks to me as (yet another instance) of a bias for the stereotypical idea that "end users want microsoft".

I think _**your**_ end users do not want that.

InfoLibre:
Time to make a fork?

Glandos:
Is there a way for instance admin to revert this to ODF?

chtixof:
I, too, am very disappointed that this request was dismissed. Given that the open ODT format is not dependent on Microsoft and will be used more and more in Europe, following Germany’s decision and, of course, those of other countries.

juliusknorr:
Please see https://github.com/nextcloud/richdocuments/issues/4812#issuecomment-4129406883, I'll lock this conversation as i think having multiple places for the same discussion just adds noise.

PR #5563, opened 2026-04-13
[stable32] fix(files): Check read permission before showing PDF edit action

elzody:
Manual back-port of https://github.com/nextcloud/richdocuments/pull/5501 to stable32


elzody:
Force-merging as the PR was already accepted before; this is a manual backport that was already tested. I had to edit it to be compatible with 32.

PR #5587, opened 2026-04-23
fix(secure-view): allow server-side file reads in SecureViewWrapper

chrip:
## Problem

`SecureViewWrapper::checkFileAccess()` was blocking **all** non-WOPI `fopen()` and
`file_get_contents()` calls on watermarked files, including server-side operations
that never expose file content to the client.

The most visible symptom: creating a new file from a template fails with
_"Unable to create new file from template"_ when Secure View is enabled for the
user's group. `TemplateManager::createFromTemplate()` reads the template file
server-side via `fopen()`, which was incorrectly treated as a download attempt.

Stack trace from the field:
```
OCP\Files\ForbiddenException: Download blocked due the secure view policy
SecureViewWrapper.php::checkFileAccess
Files/View.php::fopen
Files/Node/File.php::fopen
Files/Template/TemplateManager.php::createFromTemplate
files/Controller/TemplateController.php::create
```

## Relation to #5577

This is a companion fix to #5577, which addressed a related Secure View regression
where `shouldSecure()` called `fopen()` internally, causing a recursive
`ForbiddenException` during delete/trash operations. Both issues stem from
`SecureViewWrapper` wrapping too broadly and intercepting legitimate server-side
file operations.

## Fix

Server-side operations (template creation, background jobs, CLI) are either
non-GET or carry no HTTP context at all.
`checkFileAccess()` now checks the HTTP method. Only GET requests are blocked, which covers
all real download vectors (WebDAV, public shares, ZIP folder downloads, preview
thumbnails) while leaving server-side reads untouched.

### DIFF
--- a/lib/Storage/SecureViewWrapper.php
+++ b/lib/Storage/SecureViewWrapper.php
@@ -15,6 +15,7 @@ use OCA\Richdocuments\Service\SecureViewService;
 use OCP\Files\ForbiddenException;
 use OCP\Files\IRootFolder;
 use OCP\Files\Storage\IStorage;
+use OCP\IRequest;
 use OCP\IUserSession;
 use OCP\Server;
 
@@ -24,6 +25,7 @@ class SecureViewWrapper extends Wrapper {
 	private IRootFolder $rootFolder;
 	private IUserSession $userSession;
 	private SecureViewService $secureViewService;
+	private IRequest $request;
 
 	private string $mountPoint;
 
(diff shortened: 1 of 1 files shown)


elzody (approved):
This seems like a reasonable way to resolve this. I don't think the failing test is related; I will restart the workflow and it will likely pass. I'll also make a note of it to check for flakiness there in the future.

PR #5598, opened 2026-04-26
feat: add Office overview page with recent / shared / templates

karlitschek:
Adds a new top-level "Office" entry to the app menu that opens a Vue SPA at /apps/richdocuments/overview, providing a single landing page for the user's office work.

Features:
- Home with three sections: My recent docs, Shared with me, Templates
- Dedicated views for each, with date grouping (Today / Yesterday / Earlier this week / Earlier this month / Older) and sticky headers
- Type filter pills (Documents / Spreadsheets / Presentations / PDFs)
- List and grid view toggle, persisted in localStorage
- Hover quick-preview popover (mounted at <body> via singleton)
- Active-editor badges with pulsing live indicator (WOPI tokens)
- Pinning / favourites integrated with Nextcloud's per-user tags so pins also appear in the Files app's Favorites view
- Type-coloured thumbnails with frame on grid cards
- Friendly empty-state illustrations with calls to action
- Confetti + showSuccess toast on document creation
- Smooth fade/slide route transitions, all motion respects prefers-reduced-motion

Backend:
- OverviewService runs an indexed user-folder SearchQuery (mime IN + 60-day mtime window), partitions by ownership for recent vs shared, and batch-loads active editors and favourites per page
- OverviewController renders the SPA shell as RENDER_AS_USER
- OverviewApiController exposes paginated OCS endpoints, plus create-from-template and favourite-toggle, with strict input validation and OCS-bypass CSRF handling

Frontend:
- Vue 2 + vue-router 3 SPA mounted at #content (Files-app pattern)
- Reuses Nextcloud's design system primitives (NcContent, NcAppNavigation, NcAvatar, NcDateTime, NcDialog, NcEmptyContent, NcLoadingIcon, NcButton, NcTextField)
- Optimistic UI for pin toggling

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/css/overview.scss
+++ b/css/overview.scss
@@ -1,0 +1,9 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+// The Vue app mounts at #content (replacing it) — see overview.js. All
+// per-component styles live in their respective .vue files; this file is
+// kept as a stable hook for any future page-wide tweaks.
+
--- a/lib/Controller/OverviewApiController.php
+++ b/lib/Controller/OverviewApiController.php
@@ -1,0 +1,196 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Richdocuments\Controller;
+
+use InvalidArgumentException;
+use OCA\Richdocuments\Service\OverviewService;
+use OCP\AppFramework\Http\Attribute\NoAdminRequired;
+use OCP\AppFramework\Http\DataResponse;
+use OCP\AppFramework\OCS\OCSBadRequestException;
+use OCP\AppFramework\OCS\OCSException;
+use OCP\AppFramework\OCS\OCSForbiddenException;
+use OCP\AppFramework\OCS\OCSNotFoundException;
+use OCP\AppFramework\OCSController;
+use OCP\Files\NotFoundException;
+use OCP\Files\NotPermittedException;
+use OCP\IRequest;
+use Psr\Log\LoggerInterface;
+
+/**
+ * OCS endpoints powering the Overview page.
+ *
+ * All endpoints require an authenticated user (no public access). The
+ * Nextcloud OCS framework adds CSRF protection, brute-force throttling,
+ * and rate limiting. Inputs are coerced to safe types here and validated
+ * deeper in OverviewService.
+ */
+class OverviewApiController extends OCSController {
+	private const MAX_QUERY_LENGTH = 200;
+	private const MAX_FILENAME_LENGTH = 200;
+	private const MAX_FOLDER_PATH_LENGTH = 4000;
+
+	public function __construct(
+		string $appName,
+		IRequest $request,
+		private OverviewService $service,
+		private ?string $userId,
+		private LoggerInterface $logger,
+	) {
+		parent::__construct($appName, $request);
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @throws OCSForbiddenException when called by a guest
+	 */
+	#[NoAdminRequired]
+	public function recent(?string $q = null, ?string $type = null, int $offset = 0, int $limit = 0): DataResponse {
+		$this->requireUser();
+		[$q, $type, $offset, $limit] = $this->normalizeListParams($q, $type, $offset, $limit);
+		return new DataResponse($this->service->getRecent($this->userId, $q, $type, $offset, $limit));
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @throws OCSForbiddenException when called by a guest
+	 */
+	#[NoAdminRequired]
+	public function shared(?string $q = null, ?string $type = null, int $offset = 0, int $limit = 0): DataResponse {
+		$this->requireUser();
+		[$q, $type, $offset, $limit] = $this->normalizeListParams($q, $type, $offset, $limit);
+		return new DataResponse($this->service->getShared($this->userId, $q, $type, $offset, $limit));
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @throws OCSForbiddenException when called by a guest
+	 */
+	#[NoAdminRequired]
+	public function templates(?string $q = null, ?string $type = null, int $offset = 0, int $limit = 0): DataResponse {
+		$this->requireUser();
+		[$q, $type, $offset, $limit] = $this->normalizeListParams($q, $type, $offset, $limit);
+		return new DataResponse($this->service->getTemplates($this->userId, $q, $type, $offset, $limit));
+	}
+
+	/**
+	 * Toggle "favourite" status for a file the current user can see.
+	 * Pins / unpins the file in the overview list.
+	 *
+	 * @NoAdminRequired
+	 *
+	 * @throws OCSBadRequestException invalid arguments
+	 * @throws OCSNotFoundException   file not visible to the user
+	 * @throws OCSForbiddenException  guest call
+	 */
+	#[NoAdminRequired]
+	public function favourite(int $fileid = 0, bool $favorite = false): DataResponse {
+		$this->requireUser();
+		if ($fileid <= 0) {
+			throw new OCSBadRequestException('fileid is required');
+		}
+		try {
+			$this->service->setFavourite($this->userId, $fileid, $favorite);
+		} catch (NotFoundException) {
+			throw new OCSNotFoundException();
+		} catch (\Throwable $e) {
+			$this->logger->error('Overview: failed to toggle favourite', [
+				'exception' => $e,
+				'app' => 'richdocuments',
+			]);
+			throw new OCSException('Failed to toggle favourite');
+		}
+		return new DataResponse(['fileid' => $fileid, 'favorite' => $favorite]);
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @throws OCSBadRequestException invalid arguments
+	 * @throws OCSNotFoundException   template or folder missing
+	 * @throws OCSForbiddenException  no write permission on folder
+	 * @throws OCSException           unexpected server error
+	 */
+	#[NoAdminRequired]
+	public function createFromTemplate(int $templateFileId = 0, string $filename = '', string $folderPath = '/'): DataResponse {
+		$this->requireUser();
+
+		if ($templateFileId <= 0) {
+			throw new OCSBadRequestException('templateFileId is required');
+		}
+		$filename = trim($filename);
+		if ($filename === '' || mb_strlen($filename) > self::MAX_FILENAME_LENGTH) {
+			throw new OCSBadRequestException('filename is required and must be <= ' . self::MAX_FILENAME_LENGTH . ' characters');
+		}
+		if (mb_strlen($folderPath) > self::MAX_FOLDER_PATH_LENGTH) {
+			throw new OCSBadRequestException('folderPath is too long');
+		}
+
+		try {
+			$result = $this->service->createFromTemplate($this->userId, $templateFileId, $filename, $folderPath);
+			return new DataResponse($result);
+		} catch (NotFoundException) {
+			throw new OCSNotFoundException('Template or target folder not found');
+		} catch (NotPermittedException) {
+			throw new OCSForbiddenException('No permission to create file in target folder');
+		} catch (InvalidArgumentException $e) {
+			throw new OCSBadRequestException($e->getMessage());
+		} catch (\Throwable $e) {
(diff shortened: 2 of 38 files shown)


karlitschek:
@jancborchardt

css/overview.scss
elzody, line 1: As far as I can tell, this file is completely unneeded and was only added "just in case." It should be removed.

lib/Controller/OverviewApiController.php
```diff
@@ -0,0 +1,196 @@
...
+class OverviewApiController extends OCSController {
+	private const MAX_QUERY_LENGTH = 200;
+	private const MAX_FILENAME_LENGTH = 200;
+	private const MAX_FOLDER_PATH_LENGTH = 4000;
+
+	public function __construct(
+		string $appName,
+		IRequest $request,
+		private OverviewService $service,
+		private ?string $userId,
```
elzody, line 42: This should not be nullable. All of the methods in this controller require a user to be logged-in, so the user ID should never not exist. It also looks visually confusing since it is placed in between two dependency-injected services.

lib/Controller/OverviewApiController.php
```diff
@@ -0,0 +1,196 @@
...
+		$offset = max(0, $offset);
+		$limit = $limit <= 0 ? OverviewService::PAGE_SIZE : min(max(1, $limit), OverviewService::PAGE_SIZE);
+
+		return [$q, $type, $offset, $limit];
+	}
+
+	/**
+	 * @throws OCSForbiddenException
+	 */
+	private function requireUser(): void {
```
elzody, line 191: There are no public endpoints defined in this controller, therefore it is not needed to verify there is a user logged-in at the beginning of each controller method. This method and all the calls to it should be removed.

lib/Controller/OverviewApiController.php
```diff
@@ -0,0 +1,196 @@
...
+		private ?string $userId,
+		private LoggerInterface $logger,
+	) {
+		parent::__construct($appName, $request);
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @throws OCSForbiddenException when called by a guest
```
elzody, line 51: None of the controller methods can be called by a guest, as they are not given the `#[PublicPage]` attribute. This doc comment on each of the methods can be removed.

lib/Controller/OverviewApiController.php
```diff
@@ -0,0 +1,196 @@
...
+		if ($type !== null) {
+			$type = trim($type);
+			if ($type === '' || $type === 'all') {
+				$type = null;
+			} elseif (!array_key_exists($type, OverviewService::TYPE_GROUPS)) {
+				$type = null;
+			}
+		}
+
+		// Clamp pagination. Server caps at PAGE_SIZE; clients can ask for less but not more.
```
elzody, line 181: Pagination clamping should not be in the controller, but rather the service.

lib/Controller/OverviewApiController.php
```diff
@@ -0,0 +1,196 @@
...
+				'app' => 'richdocuments',
+			]);
+			throw new OCSException('Failed to create document from template');
+		}
+	}
+
+	/**
+	 * @return array{0: ?string, 1: ?string, 2: int, 3: int}
+	 */
+	private function normalizeListParams(?string $q, ?string $type, int $offset, int $limit): array {
```
elzody, line 158: There are constants from `OverviewService` being used here in the controller. Those things should be taken care of in the service itself, and the controller should not know the business logic.

lib/Controller/OverviewApiController.php
```diff
@@ -0,0 +1,196 @@
...
+		// Normalize search string.
+		if ($q !== null) {
+			$q = trim($q);
+			if ($q === '') {
+				$q = null;
+			} elseif (mb_strlen($q) > self::MAX_QUERY_LENGTH) {
```
elzody, line 164: Silently trimming a search query like this seems like bad practice. If someone passes a search query, it will be silently truncated and they might receive an unexpected response. This probably needs to at least be moved to the service, and handled with an explicit error that the controller can catch and return an appropriate response.

lib/Controller/OverviewApiController.php
```diff
@@ -0,0 +1,196 @@
...
+				$q = mb_substr($q, 0, self::MAX_QUERY_LENGTH);
+			}
+		}
+
+		// Validate the type filter strictly: only known group keys (or "all"
```
elzody, line 169: Validation should probably happen in the service. The controller shouldn't override the service's type validation.

lib/Controller/OverviewApiController.php
```diff
@@ -0,0 +1,196 @@
...
+
+	/**
+	 * Toggle "favourite" status for a file the current user can see.
+	 * Pins / unpins the file in the overview list.
+	 *
+	 * @NoAdminRequired
+	 *
+	 * @throws OCSBadRequestException invalid arguments
+	 * @throws OCSNotFoundException   file not visible to the user
+	 * @throws OCSForbiddenException  guest call
```
elzody, line 92: This doc comment doesn't make sense. No guest can reach this endpoint.

lib/Service/OverviewService.php
```diff
@@ -0,0 +1,639 @@
...
+		$extension = $this->extensionFromMime($template->getMimeType()) ?? pathinfo($template->getName(), PATHINFO_EXTENSION);
+		$basename = $this->sanitizeFilename($filename);
+		if ($basename === '') {
+			throw new \InvalidArgumentException('Empty filename');
+		}
+		$fullName = $extension !== '' ? $basename . '.' . $extension : $basename;
+		$fullName = $folder->getNonExistingName($fullName);
+
+		$newFile = $folder->newFile($fullName, $template->fopen('rb'));
+
```
elzody, line 217: Would need to do something like this here so that the template source is kept for the newly created file.
```suggestion
if ($this->templateManager->isSupportedTemplateSource($template->getExtension())) {
    $this->templateManager->setTemplateSource($newFile->getId(), $templateId);
}
```

lib/Controller/OverviewController.php
```diff
@@ -0,0 +1,45 @@
...
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	public function index(): TemplateResponse {
+		$this->initialState->provideInitialState('overview', [
+			'pageSize' => \OCA\Richdocuments\Service\OverviewService::PAGE_SIZE,
+			'homeSectionLimit' => \OCA\Richdocuments\Service\OverviewService::HOME_SECTION_LIMIT,
+			'windowDays' => \OCA\Richdocuments\Service\OverviewService::RECENT_WINDOW_DAYS,
+		]);
+
+		return new TemplateResponse(Application::APPNAME, 'overview', [], TemplateResponse::RENDER_AS_USER);
```
elzody, line 43: The app name is already available, so bringing in the `Application::APPNAME` constant is a bit redundant.
```suggestion
		return new TemplateResponse($this->appName, 'overview', [], TemplateResponse::RENDER_AS_USER);
```

lib/Controller/OverviewController.php
```diff
@@ -0,0 +1,45 @@
...
+		IRequest $request,
+		private IInitialState $initialState,
+	) {
+		parent::__construct($appName, $request);
+	}
+
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	public function index(): TemplateResponse {
+		$this->initialState->provideInitialState('overview', [
```
elzody, line 37: It *might* make sense to not pass these as initial state, but via the API if the frontend really needs them. Initial state makes sense for configuration values and such, but this I'm not sure.

lib/Service/OverviewService.php
```diff
@@ -0,0 +1,639 @@
...
+		],
+	];
+
+	/**
+	 * MIME types the overview surfaces (docx/xlsx/pptx/odt/ods/odp/pdf).
+	 * Built from TYPE_GROUPS so the two stay in sync by construction.
+	 *
+	 * @var list<string>
+	 */
+	public const ALLOWED_MIMETYPES = [
```
elzody, line 83: For now, having a bunch of duplicate MIME types in this class is fine, but in the future I want to do a refactoring so that they are not duplicated across the code base. Just adding this note here so it's on the record.

lib/Service/OverviewService.php
```diff
@@ -0,0 +1,639 @@
...
+	}
+
+	/**
+	 * Templates the user has defined in their personal templates folder.
+	 *
+	 * @return array{items: array<int, array<string, mixed>>, nextOffset: ?int}
+	 */
+	public function getTemplates(string $userId, ?string $query, ?string $type, int $offset, int $limit): array {
+		$this->templateManager->setUserId($userId);
+		$templates = $this->templateManager->getUser();
```
elzody, line 132: Instead of re-implementing the template MIME filtering below, you should specify the type when calling `TemplateManager::getUser()`.

lib/Service/OverviewService.php
```diff
@@ -0,0 +1,639 @@
...
+		return ['items' => $items, 'nextOffset' => $nextOffset];
+	}
+
+	/**
+	 * Resolve a template node by id. Used by the create-from-template
+	 * endpoint to validate that a template exists before copying.
+	 *
+	 * @throws NotFoundException
+	 */
+	public function getTemplateNode(string $userId, int $templateId): File {
```
elzody, line 172: This seems like an unnecessary wrapper. This method is only called from one place, so could just be inlined.

lib/Service/OverviewService.php
```diff
@@ -0,0 +1,639 @@
...
+			]),
+		];
+	}
+
+	/**
+	 * Shared core: search the user's folder, partition by ownership.
+	 *
+	 * @return array{items: array<int, array<string, mixed>>, nextOffset: ?int}
+	 */
+	private function getOwnedOrShared(string $userId, ?string $query, ?string $type, int $offset, int $limit, bool $ownedOnly): array {
```
elzody, line 233: Ideally you wouldn't use private methods to do raw database queries in the service. Better would be to use public APIs. Instead of doing the queries here, maybe it makes sense to do that in the WopiMapper and add the necessary methods?

elzody (changes requested):
General feedback, as going through the code and picking out everything wrong proved to be too much work. That said, there are still some specific comments I made when looking through the code.

- Navigating away from the "Home" tab and then back again breaks the navigation (/overview works but not /overview/)
- Opening a document from the "My recent documents" section of the "Home" page breaks the URL
    - When you try to close the document, you are left at a blank screen unless you manually change the URL
- The "... is editing this document" is not totally accurate (says I am editing but have not been for several minutes)
- Shouldn't you be able to simply make a new document instead of being forced to select a template?
- With the above point in mind, it does not even detect any of my templates so I couldn't create a document in the first place
- css/overview.scss is not used for anything
- There is a lot of redundant or dead code
- It does not seem like Nextcloud Vue components are being used everywhere, affecting accessibility and overall look and feel as a Nextcloud application
- It has the green dot like someone is editing the file even though the file has been closed for several minutes
- Favoriting a document only pins it on the "My recent documents" page and not in "Home"
- The overall design is not totally consistent with the mockup
- Why is a confetti visual needed? Seems like a waste
- Some JS files could probably be incorporated into components directly
- The hover preview seems a bit much, it's large and kind of distracting
- There are no tests

marcoambrosini (commented):
A few points from design:

1. The pattern of having file types selector in the content is opposite to what both google and microsoft have. Once a document type is selected, in both platforms those are the only files types are displayed in the app content. Recents, shared, etc., are also in the content. That's why it was initially decided to have file types selectors in the navigation. 
I don't dislike the top navigation with tabs proposed here, but if we do take this direction, we should remove the left navigation entirely, and only rely on these tabs. We can integrate recents and shared with me in the content as they were planned [here](https://github.com/nextcloud-gmbh/office-team/issues/76).
2. The “Create new” as a navigation link is a new pattern for us, I think we should stick to having this in the content, or at least as a button at the top of the navigation, depending on what we decide to do with point 1. Also I cannot find a way to actually create a new document once routed to the “Create new” page.
3. Some of the components used are not from our components library, these would need to be updated to make use of the ones we have.
4. Not design related but flagging anyway: for the display of the list of documents, it was agreed for technical reasons that the same components used in the Files app for would be re-used here. In this pr, new custom ones are used. @juliusknorr and @emberfiend ping for when you're back on this last point as you would know the detailed reasoning. 

cc @jancborchardt for further input

jancborchardt:
@karlitschek for the record, @marcoambrosini and I talked at the time of his review and yes, the idea is that any changes needed are made by him directly.

marcoambrosini:
@karlitschek the review was mostly for the record and to take notes, but I also wanted to make you aware of where this might go since you spent time making design decisions that I didn't want to override without raising the concerns first.

elzody:
I can probably already start making code changes that are unrelated to the visual design of the feature, then. We can coordinate the design improvements since it's a big PR. Might eventually be worth splitting this PR up into different pieces, then they can be merged into this branch or something.

cartierbresson:
@elzody will this make it for the 34 release? We're trying to figure out whether to include it in the release presentation or not. Thank you!

juliusknorr:
It was moved to https://github.com/nextcloud/office, closing here

Issue #2170, opened 2022-04-26
Implement token refresh or renewal to avoid expiration


juliushaertl:
- [ ] Check if there is anything within the WOPI protocol to extend the token TTL or renew the token
- [ ] Option A: Extend the token TTL on checkFileInfo calls or other WOPI operations
- [ ] Option B: On COOL
  - Show a message on expired token and let the user reload manually
    - A post message could be emitted and then richdocuments could generate a new WOPI token
  - We could also extend the token through an API call on the richdocuments side
  - If tokens should be renewed, COOL needs a way to replace the existing token while keeping the session active

juliushaertl:
Another related case this morning: 

1. Have a document open
2. Put your laptop to sleep at the end of a workday
3. Wake up the laptop next morning
4. See a document that fails to connect

- The WOPI token was expired, but Collabora is not aware

We probably should think about a post message from Collabora that allows nextcloud to create a new token in this case and pass it along or just reload collabora.

https://github.com/nextcloud/richdocuments/assets/3404133/0199cc61-407b-436d-8af5-ff3dc1d79154

juliushaertl:
cc @pedropintosilva Probably something to briefly talk about next week :)

PR #5601, opened 2026-04-28
fix(wopi): stable guest UserId and correct IsAnonymousUser for named guests

Arildm-no:
## Problem

Two bugs in `WopiController::checkFileInfo()` prevent cursor presence from working when a named guest edits a document via a public link alongside a logged-in user.

### Bug 1 — Guest `UserId` is randomly generated on every WOPI call

```php
// Before (broken)
$guestUserId = 'Guest-' . \OCP\Server::get(\OCP\Security\ISecureRandom::class)->generate(8);
```

The WOPI protocol requires `UserId` to be stable for the duration of a session. Collabora calls `CheckFileInfo` multiple times (e.g. on token refresh), and each call returned a *different* random ID for the same guest. Collabora treats each new ID as a new user, breaking cursor tracking and view presence entirely.

### Bug 2 — `IsAnonymousUser` is unconditionally `true` for all public link users

```php
// Before (broken)
if ($isPublic) {
    $response['IsAnonymousUser'] = true; // set even when guest has a display name
}
```

Collabora interprets `IsAnonymousUser = true` as a privacy signal and hides that user's cursor from other editors. This is correct for a truly anonymous viewer (no name entered), but wrong for a guest who has explicitly entered a display name in the Nextcloud sharing prompt.

## Fix

1. Derive `UserId` deterministically from the WOPI token using SHA-256, so the same guest token always maps to the same identity within a session.
2. Only set `IsAnonymousUser = true` when the guest has *no* display name. Named guests get `false`, enabling cursor visibility.

```php
// After
$guestUserId = 'Guest-' . substr(hash('sha256', $wopi->getToken()), 0, 8);

$response['IsAnonymousUser'] = empty($wopi->getGuestDisplayname());
```

## Steps to reproduce

1. Share a document with a public edit link
2. Open the document as a logged-in admin
3. Open the public link in another browser / incognito window, enter a name when prompted
4. Both users type — the logged-in user cannot see the guest's cursor or writing indicator

## Expected behaviour

Named guests have a visible, coloured cursor indicator, just like any other named collaborator.

## Tested on

- Nextcloud 33
- richdocuments 10.1.2
- Collabora Online CODE 25.04.9.4

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -485,7 +485,7 @@ class WopiController extends Controller {
 			}
 
 			$isPublic = empty($wopi->getEditorUid());
-			$guestUserId = 'Guest-' . substr(hash('sha256', $wopi->getToken()), 0, 8);
+			$guestUserId = 'Guest-' . substr(hash('sha256', $wopi->getToken()), 0, 16);
 			$userId = !$isPublic ? $wopi->getEditorUid() : $guestUserId;
 
 			$userConfig = $this->settingsService->generateSettingsConfig($type, $userId);
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -161,7 +161,7 @@ class WopiController extends Controller {
 			return new JSONResponse([], Http::STATUS_FORBIDDEN);
 		}
 
-		$guestUserId = 'Guest-' . \OCP\Server::get(\OCP\Security\ISecureRandom::class)->generate(8);
+		$guestUserId = 'Guest-' . substr(hash('sha256', $wopi->getToken()), 0, 16);
 		$user = $this->userManager->get($wopi->getEditorUid());
 		$userDisplayName = $user !== null && !$isPublic ? $user->getDisplayName() : $wopi->getGuestDisplayname();
 		$isSmartPickerEnabled = (bool)$wopi->getCanwrite() && !$isPublic && !$wopi->getDirect();
@@ -293,7 +293,7 @@ class WopiController extends Controller {
 
 		if ($isPublic) {
 			$response['UserExtraInfo']['is_guest'] = true; // DEPRECATED
-			$response['IsAnonymousUser'] = true;
+			$response['IsAnonymousUser'] = empty($wopi->getGuestDisplayname());
 		} else {
 			$response['IsAnonymousUser'] = false;
 		}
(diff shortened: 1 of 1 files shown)


lib/Controller/WopiController.php
```diff
@@ -133,7 +133,7 @@ public function checkFileInfo(
 		}
 
 		$isPublic = empty($wopi->getEditorUid());
-		$guestUserId = 'Guest-' . \OCP\Server::get(\OCP\Security\ISecureRandom::class)->generate(8);
+		$guestUserId = 'Guest-' . substr(hash('sha256', $wopi->getToken()), 0, 8);
```
joshtrichards:
```suggestion
		$guestUserId = 'Guest-' . substr(hash('sha256', $wopi->getToken()), 0, 16);
```

(Birthday Paradox mitigation).

lib/Controller/WopiController.php
```diff
@@ -434,7 +434,7 @@ public function getSettings(
 			}
 
 			$isPublic = empty($wopi->getEditorUid());
-			$guestUserId = 'Guest-' . \OCP\Server::get(\OCP\Security\ISecureRandom::class)->generate(8);
+			$guestUserId = 'Guest-' . substr(hash('sha256', $wopi->getToken()), 0, 8);
```
joshtrichards: Same.

joshtrichards (changes requested):
Good catch! Makes sense. 

I'm surprised I can't find any open bug reports on this topic.

Anyhow, from my end, as noted, I'd just like to see the output bumped up from the hash.

P.S. #2170 impacts this too (I assume). But that's not guest specific - and, more importantly, this is still a good improvement. And, presumably guests will likely also benefit from whatever eventually targets #2170 anyhow (by virtue of this change).

Arildm-no:
Thanks for the thorough review, @joshtrichards! Both requested changes have been applied:

- **Hash length bumped to 16 chars** in both `checkFileInfo()` and `getSettings()` — good call on the Birthday Paradox mitigation, 64-bit is clearly the right default here.
- **Rebased onto current `main`** — resolved the conflict with the `$isVersion` public-link guard that was added upstream; both changes coexist cleanly.

Arildm-no:
Added `Signed-off-by` to both commits to address the DCO requirement. Hash length was already bumped to 16 chars in the previous update. Should be good for review now.

elzody:
@Arildm-no I can't seem to reproduce this, would you be able to share more info on how you did that? On the latest versions it works fine.

PR #5632, opened 2026-05-03
feat(document): improve proxy status polling resilience

joshtrichards:
* Resolves: #  (various, possibly)
* Target version: main

### Summary

* Replace the unbounded 100ms polling loop for `proxy.php` status checks with limited exponential retries. 
* Split server-starting retries from transport failures so we can show a more specific fallback error when the built-in CODE server cannot be reached at all. 

This avoids indefinite waiting and improves startup robustness / user feedback for built-in server deployments.

### TODO

- [ ] Adjust tests

### Notes

Possible future follow-up: 
* unify/extract out shared proxy-status logic
  - we have two proxy status implementations: one in `src/document.js` and one in `src/services/collabora.js` (and `Office.vue` indirectly which uses the latter basically).

### Checklist

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -59,7 +59,7 @@ const setProxyLoadingMessage = (message) => {
 const showProxyFallbackError = () => {
 	setProxyLoadingMessage(t(
 		'richdocuments',
-		'Error: The Collabora Online Built-in server is taking too long to start. Please try again or set up a standalone server.'
+		'Error: The Collabora Online Built-in server is taking too long to start. Please try again or set up a standalone server.',
 	))
 	checkingProxyStatus = false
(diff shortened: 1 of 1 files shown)


elzody:
@joshtrichards Is there anything else left to finish up here?

joshtrichards:
Tests at the very least.

When I was working on this last, I'd simultaneously started trying to get a CODE integration test added in the CI chain (in a separate draft PR). 

I think I got distracted on that. ;-)

There may be some other things in this PR itself still... I'll have to look over my notes. I have a window that's been sitting open on my desktop specifically to revisit/wrap this... So there was probably something else.

 I'll try to check this coming week!

PR #5635, opened 2026-05-05
fix(richdocuments): gate conversion with SecureViewService check

chrip:
Server-side conversion bypassed the Secure View / watermark restriction that the viewer enforces, allowing a user with view-only secure access to download a clean copy via the conversion API. Reuse SecureViewService (same logic the viewer uses) to deny conversion for files that should be secured. Handle the documented NotFoundException so a cache miss surfaces as a clear, translated error instead of a 500.

* Resolves: # 
* Target version: main

### TODO

- [ x] If the file is secured, the "Save as..." is not possible because this could bypass the security. We could hide the menu entry in this case. Would need changes in `apps/files/src/actions/convertAction.ts`

### Checklist

- [x ] Sign-off message is added to all commits

### DIFF
--- a/lib/Conversion/ConversionProvider.php
+++ b/lib/Conversion/ConversionProvider.php
@@ -152,6 +152,7 @@ class ConversionProvider implements IConversionProvider {
 				$secured = $this->secureViewService->shouldSecure(
 					$file->getInternalPath(),
 					$file->getStorage(),
+					false,
 				);
 			} catch (NotFoundException $e) {
 				$this->logger->warning('Could not determine Secure View status for conversion target', ['exception' => $e]);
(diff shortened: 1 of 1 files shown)


lib/Conversion/ConversionProvider.php
```diff
@@ -144,6 +147,23 @@ public function convertFile(File $file, string $targetMimeType): mixed {
 			));
 		}
 
+		if ($this->secureViewService->isEnabled()) {
+			try {
+				$secured = $this->secureViewService->shouldSecure(
+					$file->getInternalPath(),
+					$file->getStorage(),
```
elzody, line 154: Your change succeeds in blocking the conversion, but the error message given in the log is not the same. Looks like passing `false` to `shouldSecure()` allows it to take the correct branch, else you get a `ForbiddenException`:


```suggestion
					$file->getStorage(),
					false,
```

juliusknorr (approved):

PR #5644, opened 2026-05-08
feat: office overview page

elzody:
### Summary
Adds a new "Office" app visible in the app menu that provides an overview of documents, spreadsheets, presentations, and diagrams. You can open them for editing, search for a specific file, or create a new one either from a template for a blank file.

#### Sidebar
The Files sidebar is not available since the Files app is not mounted. The sidebar relies on this, as it needs access to the Pinia store the Files app itself provides. This means that the following sidebar features are not available:

- Sharing
- File versions
- Comments

An alternative is that we can register our own sidebar tabs, but we'd need to re-implement a lot of the desired logic.

#### File Grid
Because this feature resides in Nextcloud Office (richdocuments) and uses Vue 2 without Pinia, we are unable to re-use the existing file grid components defined in the Files app. It would also likely be sloppy even if we could use them, as it would mean we're using components from another app. A new component was needed, which is `FileCard.vue`.

### Checklist


juliusknorr:
Thanks, was incorporated in https://github.com/nextcloud/office so I'll close this here

PR #5656, opened 2026-05-13
feat(direct-editing): integrate with OCP\DirectEditing API

juliusknorr:
Plug richdocuments into the server's shared Direct Editing API so mobile apps (and any other clients) can open Collabora documents through the standard /ocs/v2.php/apps/files/api/v1/directEditing/... endpoints, alongside other editors registered via OCP\DirectEditing.

This allows clients and richdocuments to at some point drop custom implementation and there is no need to maintain two implementations for the same use case.

- Add OfficeDirectEditor (IEditor) and four ACreateFromTemplate creators (document, spreadsheet, presentation, drawing) backed by the existing TemplateManager and FileCreatedFromTemplateListener.
- Extract the user/file render branch of DirectViewController::show into DirectEditingViewService so both the new IEditor::open path and the legacy /direct/{token} webview share one implementation (federation redirect, template-aware WOPI token, document template response).
- Route the legacy OCS endpoints (/api/v1/document and /api/v1/templates/new) through OCP\DirectEditing\IManager and mark them deprecated; response shape unchanged.
- Keep the share/federation flows (/api/v1/share, /api/v1/direct/share/initiator and the /direct/{token} webview) untouched -- they have no equivalent in OCP\DirectEditing.
- Factor Capabilities mimetype dedup into helpers so the editor and the capability advert read from one source of truth.


juliusknorr:
- [x] Emit `loaded` in addition to `documentLoaded` in richdocuments
- [x] Limit exposing the new API for newer client versions of the iOS and android client (anything 34 or above)

lib/Controller/OCSController.php
```diff
@@ -74,12 +85,22 @@ public function createDirect($fileId) {
...
+			// folder; Manager::open() requires a string, so fall back to the
+			// filename which is enough for the manager to resolve by fileId.
+			$path = $userFolder->getRelativePath($node->getPath()) ?? $node->getName();
+
+			// Register directly so the legacy endpoint keeps working for
+			// older mobile clients where RegisterDirectEditorListener gates
+			// out the editor from OCP\DirectEditing discovery.
+			$this->directEditingManager->registerDirectEditor($this->officeDirectEditor);
+			/** @psalm-suppress UndefinedInterfaceMethod IManager does not expose open() but the concrete Manager does, same pattern as files-app DirectEditingController */
+			$token = $this->directEditingManager->open($path, Application::APPNAME, $node->getId());
```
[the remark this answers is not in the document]
juliusknorr: This is fine

juliusknorr:
@tobiasKaminsky This would be ready for a test round from my perspective. I am unsure which versions the clients expose, for now 34 is used, but let me know if there is an internal one that is rather part of the user agent

https://github.com/nextcloud/richdocuments/pull/5656/changes#diff-818be48534441167c68b704fcea9e4af804c6edf817eff35045d670cc6ce6de5R27

lib/Service/DirectEditingViewService.php
```diff
@@ -0,0 +1,106 @@
...
+class DirectEditingViewService {
+	use DocumentTrait;
+
+	public function __construct(
+		private IRootFolder $rootFolder,
+		private TokenManager $tokenManager,
+		private TemplateManager $templateManager,
+		private FederationService $federationService,
+		private InitialStateService $initialState,
+		private Capabilities $capabilities,
```
elzody, line 41: `$initialState` and `$capabilities` don't seem to be used anywhere.

lib/Service/DirectEditingViewService.php
```diff
@@ -0,0 +1,106 @@
...
+		private IConfig $config,
+		private AppConfig $appConfig,
+		private IEventDispatcher $eventDispatcher,
+		private LoggerInterface $logger,
```
elzody, line 45: `$eventDispatcher` and `$logger` don't seem to be used anywhere.

elzody (approved):
Looks good other than the unused code. Gave it some testing locally and the API seems to work fine.

PR #5668, opened 2026-05-16
fix: mention snap install command for CODE

### LINKED ISSUE #1179: Suggested richdocumentscode install command doesn't work in all situations
**Describe the bug**
richdocumentscode can in many cases not be installed via the web interface due to timeouts. As a result, richdocuments (rightly) suggests installing using `occ` if it doesn't work. See https://github.com/nextcloud/richdocuments/blob/94dab7f3eac3a376645b0234cd3d22c21dd30031/src/components/AdminSettings.vue#L91 . However, that command assumes that one has installed Nextcloud manually instead of using e.g. the snap, where the comand doesn't work. It would be trivial to check the `$SNAP` and `$SNAP_NAME` environment variables and print the proper command in that situation (`sudo nextcloud.occ app:install richdocumentscode` specifically), but I'm afraid I don't quite have the vue chops to pull it off.

SSDWGG:
## Summary
- keep the existing manual occ install command for CODE
- add the corresponding Nextcloud Snap command for admins using Snap installations

Fixes #1179

## Tests
- npx eslint src/components/AdminSettings.vue

### DIFF
--- a/lib/Settings/Admin.php
+++ b/lib/Settings/Admin.php
@@ -56,6 +56,7 @@ class Admin implements ISettings {
 					'web_server' => strtolower($_SERVER['SERVER_SOFTWARE']),
 					'os_family' => PHP_VERSION_ID >= 70200 ? PHP_OS_FAMILY : PHP_OS,
 					'platform' => php_uname('m'),
+					'is_snap' => $this->isSnapInstallation(),
 					'fonts' => $this->fontService->getFontFileNames(),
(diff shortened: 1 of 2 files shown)


juliusknorr:
Thanks for your contribution. We already have a lot of text in the settings page where I do not think we should add even more installation methods to that. I would guess there is a way to detect that Nextcloud runs in a snap? That way we could just show the correct occ command. What do you think about that?

Issue #5545, opened 2026-04-04
Nextcloud Office "Save as" browse option


nozem79:
When saving a file in Nextcloud Office user has to type the location there it needs to be saved, that is not handy/intiutive. 

Hope to get a browse option the the next version :-)

joshtrichards:
See #5455

Issue #5455, opened 2026-03-03
Improve save as dialog


juliusknorr:
Currently our save as dialog is rather simple and technical. It allows users to put files to a location by using an absolute path within Nextcloud. 

<img width="645" height="276" alt="Image" src="https://github.com/user-attachments/assets/f5f30e5f-c4ec-410a-b180-e8e7fafce6d2" />

For a better usability we could rather use the file picker from https://github.com/nextcloud-libraries/nextcloud-dialogs, however that currently does not allow you to cover the save as use case where you would provide a filename in it as well when you pick a folder location. We probably should extend the library as this may be useful also for the files app (e.g. when copying or moving a file to a new location where you can directly specify a new name) or other apps

Reference could be the usual operating system save as dialogs:

<img width="992" height="631" alt="Image" src="https://github.com/user-attachments/assets/928aee5d-6f00-471c-88d5-f85b593bd13f" />

Issue #3891, opened 2024-08-07
Server-Side Validation for “Save As” Actions in Mobile Client Web View


alperozturk96:
When a user tries to create a spreadsheet or document and presses “Save As” from the mobile client, which uses a web view, the server should perform the validation for file name

**Demo**

https://github.com/user-attachments/assets/616ad9b5-a8c0-4366-8bc6-65e9da003639

PR #5689, opened 2026-05-21
fix(SaveAs): improve filename validation, fix cant-empty-field bug, prefill path where available

ghost:
### Summary

improve filename validation: some basic sanity checks on file names, pretty permissive so shouldn't annoy anyone
fix cant-empty-field bug: switched unedited flag to null so that emptying the field doesn't cause it to be regenerated
prefill path where available: allows anonymous-browser [direct editing endpoint](https://github.com/nextcloud/richdocuments/blob/main/docs/mobile_editor.md) sessions to see the full file path

### Checklist

### DIFF
--- a/src/components/Modal/SaveAs.vue
+++ b/src/components/Modal/SaveAs.vue
@@ -19,7 +19,7 @@
 					{{ t('richdocuments', 'Cancel') }}
 				</NcButton>
 				<NcButton type="primary"
-					:disabled="isChecking"
+					:disabled="isChecking || !isValidName"
 					@click="close">
 					{{ isChecking ? t('richdocuments', 'Checking…') : t('richdocuments', 'Save') }}
 				</NcButton>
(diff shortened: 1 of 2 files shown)


joshtrichards:
Related:
-  may or may not close out: #3891
- #5545 
- #5455

joshtrichards (commented):
Cypress tests will need to be adjusted to accomodate / cover.

src/components/Modal/SaveAs.vue
```diff
@@ -97,6 +97,20 @@ export default {
 				this.selectedPath = value
 			},
 		},
+		isValidName() {
+			const value = this.newFileName.trim()
```
juliusknorr, line 101: Should also validate against https://github.com/nextcloud-libraries/nextcloud-files/blob/main/lib/utils/filename-validation.ts#L120 but fine to get this in as it fixes the original bug

juliusknorr (approved):

PR #5690, opened 2026-05-22
Office overview design pass

marcoambrosini:
Design pass on the Office overview page:

- Document-type themed gradient background on the template section
- Scrollable template row with arrow navigation
- Restyled template and gallery file cards (borders, shadows, mime icon)
- Search moved into the app navigation
Screenshot 2026-05-22 at 20 11 44

### DIFF
--- a/src/views/OfficeOverview.vue
+++ b/src/views/OfficeOverview.vue
@@ -111,6 +111,10 @@
 										:size="48" />
 								</template>
 
+								<template v-if="activeCreator" #icon>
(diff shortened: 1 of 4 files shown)


src/views/OfficeOverview.vue
```diff
@@ -459,15 +472,22 @@ export default {
 
 .office-overview__files-title {
 	margin: 0;
-	font-size: var(--default-font-size);
+	font-size: 24px;
```
moodyjmz, line 475: should we have a var for this? so it is explicit what it is?

moodyjmz (changes requested):

PR #5692, opened 2026-05-22
chore: Adjust naming to Nextcloud Office (Collabora)

juliusknorr:
Make sure we are clear about which integration app this is


joshtrichards (approved):

joshtrichards:
Cypress failure is related to open PDF change it appears. Though unclear offhand why it would have failed just due to the string change.

juliusknorr:
Cypress failures are unrelated

elzody (approved):

PR #5699, opened 2026-05-27
fix: only bump file mtime when document transitions from modified to saved

juliusknorr:
* Resolves: # 
* Target version: main

### Summary

Opening a file (e.g. a presentation) and closing it without making any edits was bumping the modified date in the Files listing to "seconds ago". A refresh reset it to the real mtime, since the file on disk was never touched.

Cause: the `Doc_ModifiedStatus` post-message handler updated the UI's mtime whenever `args.Modified !== this.modified`, which fires on both `false → true` (doc becomes dirty) and `true → false` (saved). For presentations, Collabora emits `Modified: true` shortly after open, triggering a spurious mtime bump.

Fix: only call `FilesAppIntegration.updateFileInfo` on the `true → false` transition, matching the original intent of the handler ("Update file metadata after save", f227260fd).

### Checklist

### DIFF
--- a/src/view/Office.vue
+++ b/src/view/Office.vue
@@ -533,7 +533,7 @@ export default {
 				this.buttonClicked(args)
 				break
 			case 'Doc_ModifiedStatus':
-				if (args.Modified !== this.modified && !this.openingLocally) {
+				if (this.modified && !args.Modified && !this.openingLocally) {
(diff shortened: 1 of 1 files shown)


juliusknorr:
Actually this is not needed, Collabora seems to sent lots of messages on unchanged odp files

PR #5705, opened 2026-05-29
 fix(frontend): retry built-in CODE cold starts and show startup status in Office viewer

joshtrichards:
* Resolves: #  (likely various bug and trouble reports; improves UX with Built-in CODE)
* Target version: main

## Summary

Improve the Nextcloud Office frontend experience when using the built-in CODE server during cold start.

Previously, opening a document against the built-in CODE proxy could fail too quickly during startup:
- `checkProxyStatus()` only retried a few times
- retries happened almost immediately
- users could remain stuck on the generic loading overlay and often had to retry manually

This change makes the frontend treat built-in CODE startup as a normal temporary state:
- poll the proxy status endpoint for longer during cold start
- retry at a controlled interval instead of using a tiny fixed retry count
- keep the Office loading overlay visible while waiting
- show a more accurate loading message when the built-in CODE server is starting

## User impact

When the built-in CODE server is cold-starting, users should now see:
- a more accurate message: **"Starting the built-in CODE server …"**
- automatic waiting/retry behavior
- fewer cases where reopening the document manually is required

### `src/services/collabora.js`
- replace the global retry-counter approach in `checkProxyStatus()`
- switch to a time-based polling loop
- keep polling while proxy status is `starting`, `stopped` (i.e. not yet started), or `restarting`
- tolerate transient request/network failures throughout the polling window
- fail only after a defined overall timeout
- preserve explicit failure when the proxy reports `error`

### `src/view/Office.vue`
- show a dedicated loading message while waiting for the built-in CODE server to become ready
- clear that loading message once proxy readiness succeeds
- keep the existing failure path if startup ultimately times out or configuration is invalid
- fix inconsistent indentation in the `Insert_Button` block

## Notes

- This PR only updates the frontend behavior in `richdocuments`
- backend/proxy changes are handled separately in `richdocumentscode`
  - I'm working on some backend changes (in `richdocumentscode`) in CollaboraOnline/richdocumentscode#343 that should further improve the UX by eliminating blocking on cold starts and also make the status API a bit more useful, but those changes are not a requirement/dependency for this PR.
- the polling behavior is only relevant when the configured WOPI URL points to `proxy.php`

## Testing

Suggested manual verification:
1. Use a setup with the built-in CODE server enabled
2. Ensure the CODE process is not already running
3. Open a document
4. Confirm the UI shows the startup message while the server warms up
5. Confirm the document opens automatically once the proxy reports `OK`
6. Confirm true startup failures still surface an error state

### Checklist

### DIFF
--- a/src/services/collabora.js
+++ b/src/services/collabora.js
@@ -10,6 +10,12 @@ export const LOADING_ERROR = {
 	PROXY_FAILED: 2,
 }
 
+const PROXY_STARTING_STATES = new Set(['starting', 'stopped', 'restarting'])
+const PROXY_POLL_INTERVAL_MS = 1500
+const PROXY_POLL_TIMEOUT_MS = 30000
+
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
+
 export const isCollaboraConfigured = () => {
 	const collaboraCapabilities = getCapabilities()?.collabora
 	return isBuiltinCodeServerUsed() || collaboraCapabilities.length !== 0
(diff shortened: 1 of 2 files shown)


elzody (approved):

elzody:
The Cypress failures are related to an issue with the files PDF viewer, not this PR.

Issue #5742, opened 2026-06-09
Migrate static mimetype list to dynamic discovery via WOPI


juliusknorr:
Right now the list of supported mimetypes is maintained statically in the app, which means it can drift from what the connected Collabora/document server actually supports. Whenever upstream adds or drops a format, we have to update the hardcoded list by hand.

The WOPI /hosting/discovery endpoint already advertises the supported actions and their associated mimetypes/extensions, so we should source the list from there instead of hardcoding it.

This will also benefit connecting other office suites via WOPI that may have a different subset of supported mimetypes.

PR #5752, opened 2026-06-10
feat: Add application/epub+zip as supported mime type

juliusknorr:
* Resolves: #
* Target version: main

### Summary

Adds `application/epub+zip` to the optional mime types (`mimetypesNoDefaultOpen`).

EPUB is a read-only/import-only format in Collabora and cannot be saved back, so it is exposed as an optional mime type rather than a default-open one — consistent with how `application/pdf` is handled.

### Checklist


juliusknorr:
Ideally this is being replaced with #5742 but this is a small backportable approach

elzody (approved):

PR #5755, opened 2026-06-11
fix(RemoteService): remove redundant stream cleanup in convertFileTo()

### LINKED ISSUE #5743: TypeError in RemoteService.php:72 during preview generation despite successful thumbnail creation
## Describe the bug

When generating document thumbnails through Nextcloud Office, the thumbnail is created and displayed successfully, but Nextcloud logs an exception:

```text
fclose(): supplied resource is not a valid stream resource
in file '/var/www/nextcloud/apps/richdocuments/lib/Service/RemoteService.php' line 72
```

The issue occurs during preview generation (`/core/preview`) and does not affect opening, editing, saving, or displaying thumbnails.

## To Reproduce

1. Install Nextcloud 33.0.5.1.
2. Install Nextcloud Office 10.2.0.
3. Connect to Collabora Online 26.04.1.4.1 running in Docker behind Nginx.
4. Upload a DOCX document.
5. Open the Files app and allow a thumbnail preview to be generated.
6. Check `nextcloud.log`.

## Expected behavior

Thumbnail previews should be generated without exceptions being logged.

## Actual behavior

Thumbnail previews are generated and displayed correctly, but the following exception is logged:

```text
fclose(): supplied resource is not a valid stream resource
in file '/var/www/nextcloud/apps/richdocuments/lib/Service/RemoteService.php' line 72
```

## Screenshots

Not applicable.

## Client details:

* OS: Windows 11 x64
* Browser: Firefox
* Version: 151.0
* Device: Desktop

## Server details

**Operating system:**
Ubuntu 24.04 LTS

**Web server:**
Nginx

**Database:**
MariaDB 10.11.14

**PHP version:**
8.4.21

**Nextcloud version:**
33.0.5.1

**Version of the richdocuments app**
10.2.0

**Version of Collabora Online**
26.04.1.4.1 (Docker)

**Configuration of the richdocuments app**

```json
{
    "disable_certificate_verification": "",
    "doc_format": "ooxml",
    "enabled": "yes",
    "installed_version": "10.2.0",
    "public_wopi_url": "https://collabora.mydomain.hu",
    "timeout": "30",
    "types": "filesystem,prevent_group_restriction",
    "wopi_allowlist": "127.0.0.1,<internal IP>,<external IP>",
    "wopi_url": "https://collabora.mydomain.hu"
}
```

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)

```text
reqId: 2skP25SndZJDxuu5yz65
method: GET
url: /core/preview?fileId=590401&x=250&y=250

fclose(): supplied resource is not a valid stream resource
in file '/var/www/nextcloud/apps/richdocuments/lib/Service/RemoteService.php' line 72
```

Relevant stack trace:

```text
PreviewController
 -> PreviewManager
 -> Generator
 -> OCA\Richdocuments\Preview\Office->getThumbnail()
 -> OCA\Richdocuments\Service\RemoteService->convertFileTo()
 -> fclose()
```

Relevant code:

```php
public function convertFileTo(File $file, string $format, int $timeout = RemoteOptionsService::REMOTE_TIMEOUT_DEFAULT) {
    $fileName = $file->getStorage()->getLocalFile($file->getInternalPath());
    $stream = fopen($fileName, 'rb');

    if ($stream === false) {
        throw new Exception('Failed to open stream');
    }

    try {
        return $this->convertTo($file->getName(), $stream, $format, [], $timeout);
    } finally {
        fclose($stream);
    }
}
```

Additional observations:

* DOCX files open correctly.
* Editing works correctly.
* Saving works correctly.
* Thumbnail previews are generated and displayed correctly.
* The exception appears to occur after successful preview generation.
* No user-visible malfunction has been observed.

#### Browser log

No browser-side errors observed.

</details>

nicfab:
## Summary

Fixes #5743

`convertFileTo()` opens a raw stream and passes it to `convertTo()`, which hands it to the Guzzle-based HTTP client as the multipart request body. Guzzle wraps it in a PSR-7 stream and closes the underlying resource once the request has been sent. By the time the `finally` block runs, the resource is no longer valid, so `fclose()` logs a warning on every conversion — one per generated preview/thumbnail.

Since Guzzle takes ownership of the stream and `convertTo()` already handles request errors, this PR drops the redundant try/finally block entirely and returns the conversion result directly, as discussed in the review below.

## Testing

The removal of the explicit cleanup has been discussed in the review; the original symptom and its diagnosis were validated independently by multiple users in #5743 on different stacks (Nextcloud 32/33, richdocuments 9.1/10.2, MariaDB/PostgreSQL, Ubuntu/Debian, Collabora 25.x/26.x): conversions and previews keep working as expected and the warning is no longer logged.

### DIFF
--- a/lib/Service/RemoteService.php
+++ b/lib/Service/RemoteService.php
@@ -68,11 +68,7 @@ class RemoteService {
 			throw new Exception('Failed to open stream');
 		}
 
-		try {
-			return $this->convertTo($file->getName(), $stream, $format, [], $timeout);
-		} finally {
-			fclose($stream);
-		}
+		return $this->convertTo($file->getName(), $stream, $format, [], $timeout);
 	}
 
 	/**


elzody:
Thanks a lot for creating this pull request! Your contribution means a lot. :)

That said, the psalm failure brings up a good point:
```
Error: lib/Service/RemoteService.php:74:8: RedundantCondition: Type resource for $stream is always resource (see https://psalm.dev/122)
```

Even though `is_resource()` is a seemingly logical way to check if the stream is still there, from psalm's perspective it will always be a stream and there is no need to check the type. This got me thinking about a larger issue, though. I am not so sure we even need the try/finally block in the code to begin with. You are right that Guzzle closes the stream itself, but `convertTo()` already catches any errors thrown when sending the request. Even if an error is thrown, the stream should still get closed and there is not a need to check if the stream still exists (it will already have been closed).

TL;DR: I am wondering if we can just remove the try/finally altogether and just do
```php
return $this->convertTo($file->getName(), $stream, $format, [], $timeout);
```

What do you think @nicfab and @juliusknorr?

nicfab:
> Thanks a lot for creating this pull request! Your contribution means a lot. :)
> 
> That said, the psalm failure brings up a good point:
> 
> ```
> Error: lib/Service/RemoteService.php:74:8: RedundantCondition: Type resource for $stream is always resource (see https://psalm.dev/122)
> ```
> 
> Even though `is_resource()` is a seemingly logical way to check if the stream is still there, from psalm's perspective it will always be a stream and there is no need to check the type. This got me thinking about a larger issue, though. I am not so sure we even need the try/finally block in the code to begin with. You are right that Guzzle closes the stream itself, but `convertTo()` already catches any errors thrown when sending the request. Even if an error is thrown, the stream should still get closed and there is not a need to check if the stream still exists (it will already have been closed).
> 
> TL;DR: I am wondering if we can just remove the try/finally altogether and just do
> 
> ```
> return $this->convertTo($file->getName(), $stream, $format, [], $timeout);
> ```
> 
> What do you think @nicfab and @juliusknorr?

Thanks a lot for the review and the kind words! You're right on both counts: psalm's complaint is legitimate given the type narrowing (the `false` case is already excluded above), and since Guzzle takes ownership of the stream and closes it once the request has been sent — with `convertTo()` already handling request errors — the try/finally is indeed redundant. The only theoretical leftover would be a stream not being explicitly closed if something throws before the request body is consumed, but PHP releases the resource as soon as it goes out of scope, so it's a non-issue in practice.

I've updated the PR accordingly: dropped the try/finally and the method now returns the conversion result directly.

elzody:
@nicfab The failures now are not related to your changes -- PHP 8.2 support was dropped so it caused a lot of failures. Luckily, it has been fixed with https://github.com/nextcloud/richdocuments/pull/5759, so you should just need to rebase your changes on top of the latest changes from the `main` branch, and they will pass.

nicfab:
> @nicfab The failures now are not related to your changes -- PHP 8.2 support was dropped so it caused a lot of failures. Luckily, it has been fixed with #5759, so you should just need to rebase your changes on top of the latest changes from the `main` branch, and they will pass.

Thanks for the heads-up! Rebased on top of the latest `main`.

juliusknorr (approved):

nicfab:
Thanks for the review and merge prep, @juliusknorr! The only red check is `static-psalm-analysis`, but it's failing at the "Check for vulnerable PHP dependencies" step (roave/security-advisories vs the locked phpseclib 3.0.52) — unrelated to this change. Happy to rebase again if a dependency bump lands, just let me know.

Cris70:
Is it possible to complete the backport to stable32?
Thank you in advance

xrammit:
@juliusknorr since this issue seems to affect many installations/people do you mind releasing patched versions soon?

Issue #5770, opened 2026-06-16
WOPI PutFile fails with 500 (uncaught TypeError from files_versions) leading to save failures and conflict hints


juliusknorr:
## Summary

On a production system (Nextcloud server 34, COOLWSD 25.04, richdocuments) saves from Collabora intermittently fail. Collabora reports a conflict on the file and then fails to save. Investigation shows two distinct but related symptoms:

1. **Save failure (HTTP 500):** A `TypeError` thrown inside the `files_versions` write hook during `putContent()` escapes `WopiController::putFile()` uncaught and returns a bare `500 {"message":"Error"}` to Collabora.
2. **Conflict hint:** Collabora's own timestamp-based conflict detection raises `documentconflict` — partly as a downstream effect of the failed/retried saves, partly via the WOPI timestamp comparison.

> **Update — the 500 is a known nextcloud/server bug, already fixed on master:** [nextcloud/server#60842](https://github.com/nextcloud/server/pull/60842) *"fix(files_versions): guard null path in event listeners"* (milestone NC 35, merged 2026-06-11) guards the five callers of `getPathForNode()` from passing `null` downstream — it explicitly addresses this `basicOperation(): Argument #2 ($path) must be of type string, null given` in public-link/groupfolder scenarios. Backports are in flight, including **stable34 (#61189)**, stable33 (#61188), stable32 (#61187), stable31 (#61329). The production version `34.0.0.11` predates the stable34 backport, which is why it still crashes. (An earlier, related but different null fix was [#51609](https://github.com/nextcloud/server/pull/51609), NC 32.) The richdocuments-side hardening below is still worthwhile so a hook `\Error` no longer surfaces as an uncaught 500.

## Logs

Nextcloud (`richdocuments`), on `POST …/wopi/files/12049115_…/contents`:

```
Uncaught error: OC\Files\View::basicOperation(): Argument #2 ($path) must be of type string, null given,
called in …/View.php on line 525 in file …/View.php line 1204
```

Previous (root) exception:

```
TypeError: OC\Files\View::basicOperation(): Argument #2 ($path) must be of type string, null given
#0 …/View.php(525): OC\Files\View->basicOperation()
#1 …/Filesystem.php(496): OC\Files\View->file_exists()
#2 …/apps/files_versions/lib/Storage.php(165): OC\Files\Filesystem::file_exists()
[...]

PR #5771, opened 2026-06-16
fix(wopi): harden file saving against failures wrongly seen as conflicts

juliusknorr:
## Summary

Hardens the WOPI save paths so that a *failed* save can no longer surface to the Collabora client as a spurious document conflict, and so that errors raised inside filesystem write hooks no longer escape as uncaught errors.

Background and full investigation: #5770.

On a production system (NC 34) saves intermittently failed with `500 Internal Server Error` and the client reported a conflict. The root cause of the 500 is a `TypeError` thrown inside the `files_versions` write hook (`getPathForNode()` returns `null` → `Storage::store(null)` → `View::basicOperation(…, null)`). That is a `nextcloud/server` bug already fixed on master in nextcloud/server#60842 (backport to stable34: #61189). Because a `TypeError` is an `\Error` and not an `\Exception`, our `catch (\Exception)` did not catch it, so it escaped as an opaque 500 — which the client can then reconcile into a false conflict.

## Changes

- **Catch `\Throwable`** (not just `\Exception`) in `putFile()` and `postFile()`, so errors raised inside filesystem hooks produce a controlled response and a meaningful log entry instead of an uncaught error.
- **mtime no-op guard:** snapshot the file's mtime before writing and log a distinct error if it changed despite a failed save. An advanced mtime after a failure indicates a partial write, which is exactly what makes the client assume the document changed externally.
- **`LockedException` → `409 Conflict`** instead of `500`, so the client retries the save instead of treating a transient lock as a server error.

This does not (and should not) suppress conflicts caused by a genuinely concurrent successful save — it only ensures a *failed* save never *looks* like an external change.

## Follow-up (not in this PR)

- Revisit `Helper::toISO8601()` second-only precision (always `.000000`) to reduce timestamp-rounding false positives — needs validation against the Collabora client and is tracked separately.

## Testing

- `php -l` clean. Manual reasoning against the reported production traces; no automated test added for the hook-error path as it requires the versions listener fault injection.

## AI tool use disclosure

This change was prepared with the assistance of an AI coding agent (Claude Code). The investigation, code changes, and this description were AI-assisted and reviewed by the committer before submission. Commits carry an `Assisted-by:` trailer.

https://claude.ai/code/session_0148r9TcHSywDqJVHqyMggub

---
_Generated by [Claude Code](https://claude.ai/code/session_0148r9TcHSywDqJVHqyMggub)_

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -651,6 +651,12 @@ class WopiController extends Controller {
 				return new JSONResponse(['COOLStatusCode' => self::COOL_STATUS_DOC_CHANGED], Http::STATUS_CONFLICT);
 			}
 
+			// Remember the current mtime so we can detect whether a failed save
+			// left the file modified (a partial write). Reporting an advanced mtime
+			// afterwards would make the WOPI client assume the document changed
+			// externally and raise a spurious conflict.
+			$mtimeBefore = $file->getMTime();
+
 			$content = fopen('php://input', 'rb');
 
 			$freespace = (int)$file->getStorage()->free_space($file->getInternalPath());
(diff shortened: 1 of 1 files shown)


lib/Controller/WopiController.php
```diff
@@ -821,7 +839,10 @@ public function postFile(
 		} catch (NotFoundException $e) {
 			$this->logger->warning($e->getMessage(), ['exception' => $e]);
 			return new JSONResponse([], Http::STATUS_NOT_FOUND);
-		} catch (\Exception $e) {
+		} catch (\Throwable $e) {
```
CarlSchwan, line 842: not a fan of catching a Throwable, the fix in server is almost fully backported

juliusknorr: Well, we still log it and I'd want some logging in place. I can rethrow, but having a defined HTTP status returned seems good no?

PR #5834, opened 2026-07-07
fix(previews): properly read from lazy cache

### LINKED ISSUE #5600: Documentation to disable preview generation wrong
In the documentation, in order to disable preview generation, the config value is set as lazy:
https://github.com/nextcloud/richdocuments/blob/dbaf059de992c7756f841741e5034c9ecd489b78/docs/app_settings.md?plain=1#L40-L43

When the configuration value is actually loaded, only non-lazy values are loaded, since [getAppValueBool()](https://github.com/nextcloud/server/blob/d3534595da939722c69953f66d2c1000156e4bea/lib/public/IAppConfig.php#L240) defaults to non-lazy if not explicitly set.
https://github.com/nextcloud/richdocuments/blob/dbaf059de992c7756f841741e5034c9ecd489b78/lib/AppConfig.php#L247-L249

elzody:
* Target version: main

### Summary
According to the documentation in [app_settings.md](https://github.com/nextcloud/richdocuments/blob/main/docs/app_settings.md#previews), the correct way to disable document preview generation is via the following command:

```
occ config:app:set richdocuments preview_generation --type boolean --lazy --value false
```

However, I noticed that this does not actually disable preview generation in all cases, as the app config does not actually check for a value in the lazy cache:

https://github.com/nextcloud/richdocuments/blob/68cdca19da53a4f53a159bec32ea7fccf8d87a17/lib/AppConfig.php#L253-L255

According to the documentation for [`OCP\AppFramework\Services\IAppConfig::getAppValueBool()`](https://nextcloud-server.netlify.app/classes/ocp-appframework-services-iappconfig#method_getAppValueBool), if a config value is lazy then the `$lazy` argument must be set to `true`. If it is not set to true then it would only be loaded from the lazy cache if something else reached for the lazy cache earlier in the request, which loads all lazy values into the cache. Otherwise, `$lazy` defaults to `false` so if nothing else in the request gets a value from the lazy cache, then the preview generation is not actually disabled.

This change ensures that the preview generation config value is fetched from the lazy cache explicitly, so that it always has the expected behavior and is not dependent on something else hitting it earlier in the request.

### Checklist

### DIFF
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -251,7 +251,7 @@ class AppConfig {
 	}
 
 	public function isPreviewGenerationEnabled(): bool {
-		return $this->appConfig->getAppValueBool('preview_generation', true);
+		return $this->appConfig->getAppValueBool('preview_generation', true, lazy: true);
 	}
 
 	/**


AndyScherzinger (approved):
not very eager, eh ;)

Issue #5832, opened 2026-07-07
WOPI timestamp comparison uses stale modification time causing false DOC_CHANGED conflicts


demirkaric:
**Describe the bug**

During collaborative editing of an ODS document with multiple users, Collabora repeatedly displays:

> Document has been changed in storage. What would you like to do with your unsaved changes?

Although the document is only being edited through Collabora and nobody modifies it outside the active WOPI session.

After investigating the source code, I traced the conflict to the following code in richdocuments:

```php
if (!empty($wopiHeaderTime) && $wopiHeaderTime !== Helper::toISO8601($file->getMTime() ?? 0)) {
    return new JSONResponse(
        ['COOLStatusCode' => self::COOL_STATUS_DOC_CHANGED],
        Http::STATUS_CONFLICT
    );
}
```

The comparison fails because the modification time used during the comparison is older than the actual modification time of the file on disk.

Executing:

```bash
occ files:scan --path="<path>"
```

immediately updates `oc_filecache.mtime`, and after that the WOPI conflicts disappear.

During debugging I compared the filesystem timestamp with the metadata stored in `oc_filecache`.

Filesystem:

```
Modify: 2026-06-05 06:43:04 UTC
```

`oc_filecache`:

```
mtime: 2026-06-05 06:38:15 UTC
```

Running `occ files:scan` immediately synchronized `oc_filecache.mtime` with the filesystem timestamp and the conflicts stopped occurring.

This appears to indicate that stale metadata is being used during the WOPI timestamp comparison.

The issue has also been observed on previous Nextcloud versions before upgrading to 33.0.5.1.

---

**To Reproduce**

1. Install Nextcloud and Collabora Online on separate servers.
2. Use local filesystem storage.
3. Share a large ODS document between several users.
4. Let multiple users edit the document simultaneously for some time.
5. Eventually one or more users receive:

   > Document has been changed in storage.

6. Nextcloud responds with `HTTP 409 Conflict` and `COOLStatusCode=DOC_CHANGED`.

---

**Expected behavior**

The WOPI timestamp comparison should not fail when the file has only been modified through the current collaborative editing session.

The metadata used for the timestamp comparison should always reflect the current state of the stored file.

---

**Screenshots**

-

---

**Client details:**

- OS: Windows 11
- Browser: Google Chrome
- Browser Version: 138.0.7204.101 (64-bit)
- Device: Desktop

---

## Server details

**Operating system:**

Ubuntu Server 24.04 LTS

[...]

Issue #5794, opened 2026-06-23
Send reliable timestamp for when "Document changed in storage" conflicts  happen


pedropintosilva:
**Describe the bug**

It seems - in some cases - users are seeing the **"Document changed in storage"** conflict dialog even though no one else touched the file. This could be due to how richdocuments produces/compares the timestamp as it seems a cached value is been passed.

LastModifiedTime is already sent but its value - in some cases - is unreliable due to stale cached `mtime` after putContent. Probably best to ask storage layer again for fresh metadata instead of trusting that cached value. This might not fix the issue but it will fix for that situation and make further chasing easier to do.

Related: https://gerrit.collaboraoffice.com/c/online/+/4918 from @mmeeks

mmeeks:
Hmm? not sure what the 'LastModifiedTime' thing means above - unless that's a Nextcloud / richdocuments detail but I'd like to see something like: (not compiled / tested) included

[0001-Add-a-stamp-field-with-the-actual-mtime-to-the-json-.patch](https://github.com/user-attachments/files/29254998/0001-Add-a-stamp-field-with-the-actual-mtime-to-the-json-.patch)

juliusknorr:
This was what Pedro and I thought was required to be added in the conflict case. It would be in sync with the naming used in other places, but we can pick your patch if `stamp` is what should be used from the Collabora Online side.

mmeeks:
Ah - the name doesn't really matter; we will show whatever is in that JSON next to the COOLStatusCode to the user if they ask for that so do use whatever is most consistent :-)

mmeeks:
Ah - but, of course - while we are here - if you'd like to go wild and lets us know who it was that did the last document upload and the size of that as well we can standardize that and present that more attractively in the dialog:

"JimBob up-loaded a new version <ten seconds ago> of size <1023 bytes>" or whatever :-) it all helps the user I expect.

pedropintosilva:
@elzody as discussed today, I'm pasting here our findings while having
* A collaborative editing session
* On a odp document
* Every user was editing in the web browser
* At least 2 users had the NC Sync client app running with that file synchronized
   * No Sync client error, file was also not opened locally, and there's just one 'sync' action in the detailed backlog (which seems to be right after the call/editing sessions) .


<img width="902" height="369" alt="Image" src="https://github.com/user-attachments/assets/53fd58d3-98ca-4993-8861-0c05cd99b91f" />

[...]

PR #5870, opened 2026-07-14
fix(wopi): use fresh storage mtime for conflict detection and responses

elzody:
**Resolves:**
- https://github.com/nextcloud/richdocuments/issues/5832
- https://github.com/nextcloud/richdocuments/issues/5794

**Target version:** `main`

### Summary
Apparently, in some cases when multiple people are editing a document, especially in longer editing sessions, the `mtime` of the file in the `oc_filecache` table can become out of sync (behind) what the `mtime` is on the file storage, leading to a conflict. Re-scanning files on an instance via `occ files:scan` resolves the conflict as the file cache's `mtime` is updated to match the storage's.

I have not been able to reproduce the issue, but hopefully reading the `mtime` from the storage itself helps instead of relying on a potentially outdated one. Additionally, it includes the `LastModifiedTime` in the conflict dialog.

### Checklist

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -641,14 +641,18 @@ class WopiController extends Controller {
 		try {
 			$file = $this->getFileForWopiToken($wopi);
 			$wopiHeaderTime = $this->request->getHeader('X-COOL-WOPI-Timestamp');
+			$storageMtime = $file->getStorage()->filemtime($file->getInternalPath());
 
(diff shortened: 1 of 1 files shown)


demirkaric:
Hope the fix will be released soon. We are experiencing a lot of issues in production because of this bug.

Aveyron-RetD:
Did the server release https://github.com/nextcloud/server/pull/62204 will fix this or will make things worse

PR #5873, opened 2026-07-17
feat: Add text/markdown as supported mime type

vmiklos:
This is related to
<https://github.com/CollaboraOnline/online/issues/16018>.

### Summary

Once both https://gerrit.collaboraoffice.com/c/online/+/7180 is submitted and this PR is merged, users can edit markdown files in COOL, which was inconsistent, since convert-to worked, but not interactive editing.

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -66,6 +66,7 @@ class Capabilities implements ICapability {
 		'application/pdf',
 		'application/epub+zip',
 		'text/plain',
+		'text/markdown',
 		'text/spreadsheet',
 	];
 
--- a/src/file-actions.js
+++ b/src/file-actions.js
@@ -51,3 +51,46 @@ const openPdf = {
 }
 
 registerFileAction(openPdf)
+
+const openMarkdown = {
+	id: 'office-open-markdown',
+
+	iconSvgInline: () => {
+		// Make sure the icon is the correct color
+		return appIcon.replaceAll(/#(fff|0{6})/g, 'currentColor')
+	},
+
+	displayName: () => {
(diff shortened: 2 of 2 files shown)


vmiklos:
Hi @elzody @juliusknorr , could you please review this when you have time? Thanks.

We recently noticed that while the COOL convert-to API supported markdown and the postmessage API also supports markdown copy&paste, actual open, edit, save of .md files is not working with COOL. This is to fix it. Use-case is something like opening a .md snippet in COOL, then you give some polished look like adding an image in the header, a page number in the footer interactively, then export to PDF or something like that.

elzody:
I had the same thought as Caolán in [this comment](https://gerrit.collaboraoffice.com/c/online/+/7180/comments/78cbc351_cebef68d). If we add `text/markdown` support it might ideally go into the `MIMETYPES_OPTIONAL` array - that way we can keep the Text app as the default but still allow the user to edit Markdown files with Collabora if they wish. We do this already with PDFs (see [src/file-actions.js](https://github.com/nextcloud/richdocuments/blob/main/src/file-actions.js)).

vmiklos:
> If we add `text/markdown` support it might ideally go into the `MIMETYPES_OPTIONAL` array - that way we can keep the Text app as the default but still allow the user to edit Markdown files with Collabora if they wish.

Thanks for the review, makes sense. I updated the PR to do it the PDF way: the MIME type goes to the MIMETYPES_OPTIONAL list and then a new file action in src/file-actions.js. Please take a look.

elzody (approved):
I gave it a quick spin and it works nicely. Thanks for the contribution 💜️

PR #5888, opened 2026-07-20
fix(ci): fix collabora aliasgroup

elzody:
* Target version: main

### Summary
It seems something with the latest Collabora image broke CI. It is now not responding with discovery XML from `/hosting/discovery`. This seems to be related to an upstream change that shipped with Collabora 26.04 that changes the alias groups setting, so the workflows need to change to use the correct alias group.

https://github.com/CollaboraOnline/online/commit/69a7b0efa82e5a098939a96896fdac7a8cb01efb

### Checklist


elzody:
This PR seems irrelevant and the underlying problem will be fixed with https://gerrit.collaboraoffice.com/c/online/+/7354 I think.

PR #5891, opened 2026-07-21
fix(settings): resize the settings iframe to its reported height

timar:
* Resolves: # 
* Target version: main

### Summary

The embedded settings page posts an Iframe_Height message so the parent can grow the iframe to fit its content. The handler read event.data as an object, but the message arrives as a JSON string, so data.MessageId was always undefined and the resize never ran, leaving a short frame with an inner scrollbar. Parse the string before inspecting it.

Also set style.height instead of the deprecated height attribute, which ignores a "<n>px" value, verify the sender origin, and add a min-height fallback for the moment before the first height message arrives.

### Checklist

### DIFF
--- a/src/components/CoolFrame.vue
+++ b/src/components/CoolFrame.vue
@@ -98,9 +98,26 @@ export default {
 				if (this.iframeUrl && event.origin !== new URL(this.iframeUrl).origin) {
 					return
 				}
-				const data = event.data
-				if (data.MessageId === 'Iframe_Height') {
-					document.getElementById(this.iframeName).height = data.Values.ContentHeight
+
+				// The embedded settings page posts a JSON string, so event.data is a
+				// string here, not an object. Parse it before inspecting MessageId.
+				let data = event.data
+				if (typeof data === 'string') {
+					try {
+						data = JSON.parse(data)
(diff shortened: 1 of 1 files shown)


src/components/CoolFrame.vue
```diff
@@ -93,13 +93,29 @@ export default {
...
 	methods: {
 		handlePostMessage(event) {
-			try {
-				const data = event.data
-				if (data.MessageId === 'Iframe_Height') {
-					document.getElementById(this.iframeName).height = data.Values.ContentHeight
+			// Only trust messages from the Collabora origin that serves the iframe.
+			if (event.origin !== new URL(this.iframeUrl).origin) {
+				return
+			}
```
elzody: We are adding post message origin verification in https://github.com/nextcloud/richdocuments/pull/5874, but for now this is fine to stay to get this fix in. I can handle any merge conflicts or adjustments that need taken care of later once the other PR is merged. ^^

Just adding this here for reference.

elzody (approved):

elzody:
The Cypress tests should pass again after a rebase onto the latest `main` branch. Could you take care of that?

PR #5934, opened 2026-07-27
fix(ci): stabilize cypress tests

### LINKED ISSUE #5930: TypeError: openLocally is not a function on close of federated document (blank page loop)
Beschreibung:
When opening a federated document (shared via Federated Cloud Sharing between two Nextcloud instances) with Nextcloud Office/Collabora and then closing it, the browser is left with a blank page instead of returning to the Files view. The bug is reproducible in both directions (Server A opens a file federated from Server B, and vice versa).

Schritte zum Reproduzieren:

Two Nextcloud instances (A and B) with Federation trust established (federation_use_trusted_domains enabled in richdocuments on both)
Share a folder from A to a user on B via Federated Cloud Sharing
On B, open an office document from the federated share with Nextcloud Office
Edit and click "Close"
Result: blank page instead of returning to Files view

Konsolen-Fehler:

Error during post message handler Object TypeError: Te.openLocally is not a function
    at Object.onClose (document.js:634:17)
    at document.js:439:21
    at postMessage.tsx:87:17
    at Array.forEach (<anonymous>)
    at e.handlePostMessage (postMessage.tsx:81:34)
    at postMessage.tsx:18:19

Zusätzlicher Hinweis: After the error, the URL on the accessing instance remains unchanged (including openfile=true), so reloading the page re-opens the same file in Collabora again — creating a loop. Workaround: manually navigate to the plain /apps/files/ URL.

Umgebung:

Nextcloud-Version: 34.0.2.1 (both servers)
Richdocuments-Version: 11.1.0 (both servers)
Collabora Online (coolwsd): 11.1.0 (both servers)
federation_use_trusted_domains: yes (both servers)
2 independent Nextcloud instances, each with its own coolwsd, native package installation (Debian)

elzody:
### Summary
This pull request attempts to stabilize failing Cypress tests and addresses the following concerns:

- There was a Vue.js mixin being called but `document.js` is not a Vue component so it did not exist
- Add a direct editing guard to `PostMessages.sendPostMessage('parent', 'close')` to prevent an infinite loop in `document.js`
- Grant `SYS_ADMIN` and `SYS_CHROOT` capabilities to the Collabora container so that forkit processes can call `coolmount`, preventing PHP workers from being stalled
- The "Save as" test in `direct.spec.js` was made more robust and uses a more resilient selector
- The post message target origin test in `direct.spec.js` was made less flakey by confirming Collabora has finished initializing before setting the post message spy

*🤖 This PR was made with the help of Claude but the PR description is my own.*

### Checklist

### DIFF
--- a/.github/workflows/cypress-e2e.yml
+++ b/.github/workflows/cypress-e2e.yml
@@ -47,6 +47,7 @@ jobs:
           aliasgroup1: 'http://172.17.0.1'
         ports:
           - "9980:9980"
+        options: --cap-add SYS_ADMIN --cap-add SYS_CHROOT
 
     steps:
       - name: Checkout server
--- a/cypress/e2e/direct.spec.js
+++ b/cypress/e2e/direct.spec.js
@@ -220,13 +220,18 @@ describe('Direct editing (legacy)', function() {
 				.then((token) => {
 					cy.nextcloudTestingAppConfigSet('richdocuments', 'uiDefaults-UIMode', 'classic')
 					cy.logout()
-					cy.visit(token)
+					cy.visit(token, {
+						onBeforeLoad(win) {
+							cy.spy(win, 'postMessage').as('postMessage')
+						},
+					})
 					cy.waitForCollabora(false)
+					cy.waitForPostMessage('App_LoadingStatus', { Status: 'Document_Loaded' })
 					cy.get('[data-cy="coolframe"]').then($iframe => {
 						const collaboraOrigin = $iframe[0].contentWindow.location.origin
 						cy.spy($iframe[0].contentWindow, 'postMessage').as('postMessage')
-						cy.dispatchMessageFromOrigin(collaboraOrigin, { MessageId: 'App_LoadingStatus', Values: { Status: 'Document_Loaded' } })
-						cy.waitForPostMessage('Host_PostmessageReady', undefined, { targetOrigin: collaboraOrigin })
+						cy.dispatchMessageFromOrigin(collaboraOrigin, { MessageId: 'Action_Paste', Values: {} })
+						cy.waitForPostMessage('Action_Paste', undefined, { targetOrigin: collaboraOrigin })
 					})
 				})
 		})
@@ -250,7 +255,7 @@ describe('Direct editing (legacy)', function() {
 	it('Save as', function() {
 		createDirectEditingLink(randUser, this.fileId)
 			.then((token) => {
-				cy.nextcloudTestingAppConfigSet('richdocuments', 'uiDefaults-UIMode', 'tabbed')
+				cy.nextcloudTestingAppConfigSet('richdocuments', 'uiDefaults-UIMode', 'notebookbar')
 				cy.logout()
 				cy.visit(token, {
 					onBeforeLoad(win) {
@@ -265,9 +270,14 @@ describe('Direct editing (legacy)', function() {
 						.should('be.visible')
 
 					cy.get('button[aria-label="File"]').click()
-					cy.get('button[aria-label="Save As"]').click()
+					cy.get('button[aria-label="Save As"]')
+						.should('be.visible', { timeout: 10_000 })
+						.click()
 
-					cy.get('#saveas-entries #saveas-entry-1').click()
+					cy.get('#saveas-entries > div', { timeout: 30_000 })
+						.contains('Rich Text (.rtf)')
+						.should('be.visible')
+						.click()
 				})
 
 				cy.get('.saveas-dialog')
--- a/src/document.js
+++ b/src/document.js
@@ -635,9 +635,9 @@ const documentsMain = {
 
 		$('footer,nav').show()
 		documentsMain.UI.hideEditor()
-		documentsMain.openLocally()
-
-		PostMessages.sendPostMessage('parent', 'close')
+		if (!isDirectEditing()) {
+			PostMessages.sendPostMessage('parent', 'close')
+		}
 	},
 
 	onCloseViewer() {


chrip (dismissed):
## Summary

Three separate fixes bundled together: a real production bug fix in `src/document.js`,
a Save-As selector hardening in `direct.spec.js`, and a capability grant for the Collabora
service container in CI.

I verified each against the actual code and against the CI logs of this PR's head
(run 30320028239, `a13ffc62`) compared to main. The results are mixed:

- **The `src/document.js` diagnosis is correct and the fix is a genuinely good catch** — I traced
  the infinite loop end to end (details below). Nice work; this is a real user-facing bug,
  not just a test problem.
- **The Save-As selector fix works** — that was one of the two flaky failures on main, and the
  test now passes in this PR's run.
- **The targetOrigin test fix does not work** — it is still red on this PR's head, in exactly the
  same test, and swapping the MessageId cannot address the actual cause.
- **The CI capability grant does not achieve its stated goal** — I diffed the Collabora logs
  with and without it; the mount still fails, just one step later, and it produces *more*
  error lines than before. It also grants `SYS_ADMIN` to an unpinned third-party image.

Also: the PR description's second bullet describes the opposite of what the code does.

## Verification

| Claim / Item | Reality | Status |
|--------------|---------|--------|
| `openLocally` is a Vue-mixin method that does not exist on `documentsMain` | Confirmed. Only defined in `src/mixins/openLocal.js:80`. `documentsMain` has no such member, so `documentsMain.openLocally()` was always a `TypeError`. | ✓ |
| Removing it "prevents an infinite loop" | Confirmed, and the mechanism is real: `document.js:436-440` maps incoming `close`/`UI_Close` → `onClose()`; `onClose()` posts `'close'` to `parent`; in direct editing `window.parent === window`, so the page messages *itself*; `allowedOrigins` includes `window.location.origin` (`document.js:584-587`) so it is accepted; `parsePostMessage('close')` yields `msgId = 'close'` → `onClose()` again → unbounded loop. The `TypeError` on the preceding `openLocally()` line was swallowed by the `try/catch` in `postMessage.tsx:handlePostMessage`, which is why the loop had been masked until now. | ✓ |
| `isDirectEditing` is available in `document.js` | Confirmed, imported at `src/document.js:18`; guard style matches the existing one at `src/document.js:403`. | ✓ |
| "Save as" test was actually flaky on main | Confirmed. Main run 30117271093 and 30048559850 both fail with `Expected to find element: '#saveas-entries #saveas-entry-1', but never found it`. The content-based selector fixes this — `✓ Save as (9600ms)` in this PR's run. | ✓ |
| `uiDefaults-UIMode` value `notebookbar` is correct | Confirmed as the app default (`lib/Service/InitialStateService.php:164`, `src/helpers/coolParameters.js:14`). | ✓ |
| "Disables the Collabora mount jail in CI" | ❌ The diff does the opposite — it *grants* `SYS_ADMIN`/`SYS_CHROOT` so the jail's mount namespace can work. The commit message (`grant SYS_ADMIN and SYS_CHROOT…`) is accurate; the PR description is not. | ❌ |
| "It reduces errors in the logs" | ❌ Measured, not guessed. Green main run (job 89592681975, no cap-add): 3 lines, failing at `enterMountingNS, CLONE_NEWUSER unshare failed (EPERM)` `JailUtil.cpp:86`. This PR's head (job 90153789084, with cap-add): 4 lines, now failing later at `enterMountingNS, root mount failed (EACCES)` `JailUtil.cpp:104` plus a new `/usr/bin/coolmount: mount failed to bind … Permission denied`. Net: one *more* error line. | ❌ |
| The mount noise was causing test failures / slow teardown | ❌ It is present verbatim in a fully **green** main run (30126982354). coolwsd falls back and works. This was never blocking. | ❌ |
| "stabilize cypress tests" — CI is now stable | ❌ Still red at head: `Direct editing (legacy) > PostMessage origin security > sends messages with the Collabora targetOrigin` — `Timed out retrying after 20000ms`, `cypress/support/commands.js:303`. | ❌ |

## Issues & Suggestions

### 🔴 Blocking

- **The targetOrigin test is still failing — the fix does not work.**
  `cypress/e2e/direct.spec.js:225-229`. The failure is at `cypress/support/commands.js:303`
  (`expect(messagesMatchingId.length).to.be.greaterThan(0)`), i.e. **no** `postMessage` call
  carrying `"MessageId":"Action_Paste"` was ever recorded on the spy. So the problem is not the
  choice of message ID, and swapping `App_LoadingStatus`/`Host_PostmessageReady` for
  `Action_Paste` cannot fix it.

  For what it's worth, the message *should* be forwarded: `Action_Paste` is in the `ViewerToLool`
  list (`src/document.js:350-357`), `isViewerMode` is false, so `document.js:407-409` hits
  `PostMessages.sendPostMessage('loolframe', data)`, and `targetOrigins.loolframe` is set to the
  urlsrc origin (`document.js:588-591`) — the `targetOrigin` assertion would have passed *if the
  call had been seen*. Two candidates worth instrumenting:

  1. **Stale spy target.** `cy.spy($iframe[0].contentWindow, 'postMessage')` patches an own
     property on the iframe's `WindowProxy`, but the app resolves
     `document.getElementById('loleafletframe').contentWindow` *lazily at send time*
     (`document.js:29`). Any internal navigation inside `cool.html` after `waitForCollabora()`
     replaces the underlying `Window` and drops the patch. This is my main suspicion, and it
     would explain both the old and new versions failing.
  2. **Incoming message rejected.** If `$iframe[0].contentWindow.location.origin` does not equal
     `new URL(urlsrc).origin`, `postMessage.tsx:35-39` rejects the synthetic event outright and
     nothing is forwarded. Note CI sets `wopi_url=http://localhost:9980` (workflow line 155) while
     `CYPRESS_collaboraUrl=http://172.17.0.1:9980` (line 19), and commit `7b25afad`
     ("fix(ci): properly get collabora origin") suggests this has bitten before.

  Cheap diagnostic: mirror the sibling test and also assert on `@consoleWarn` — if
  `'PostMessageService: rejected message from unexpected origin'` fires, it is (2); if not, it is (1).
  For (1), spy on the send side (wrap `PostMessages.sendPostMessage` / the top window) rather than
  on the Collabora frame's `contentWindow`.

  Until this is resolved the PR does not deliver on its title, so it should not merge.

- **Drop the `--cap-add SYS_ADMIN --cap-add SYS_CHROOT` change.**
  `.github/workflows/cypress-e2e.yml:50`. Per the log comparison in the table above it does not
  fix the mount (EPERM at `JailUtil.cpp:86` simply becomes EACCES at `JailUtil.cpp:104`, plus an
  extra `coolmount` error), it adds noise rather than removing it, and the noise it targets is
  present in green runs and blocks nothing.

  On top of that it is a CI-hardening regression: `SYS_ADMIN` is close to root, and the image it
  is granted to is `juliushaertl/nc-code-nightly:latest` — a third-party personal namespace, on a
  floating tag, already carrying a `# zizmor: ignore[unpinned-images]` waiver. Granting near-root
  capabilities to an unpinned nightly image that runs alongside a repo checkout is not a trade I
  would make for log cosmetics. Getting the mount to actually succeed would additionally require
  relaxing seccomp/AppArmor, which is worse still.

  If the log noise is genuinely worth removing, do what the PR *description* says and turn the
  jail off through coolwsd configuration in the existing `extra_params` (alongside
  `--o:ssl.enable=false --o:home_mode.enable=true`) rather than through container capabilities —
  please confirm the exact option name against the CODE version in use before relying on it.

- **PR description contradicts the code.** Bullet 2 says the jail is *disabled*; the change
  *grants capabilities so the jail can work*. Please fix the description (or the code) so they
  agree — this is the kind of mismatch that survives into the changelog.

### ⚠️ Major

- **`isDirectEditing()` is a proxy for the wrong condition.** `src/document.js:638`. The actual
  precondition for the self-message loop is `window.parent === window`, not "direct editing".
  The same legacy `documents` template is rendered top-level *without* a `direct` param by
  `DocumentController` — e.g. `publicPage()` (`lib/Controller/DocumentController.php:214-228`,
  route `/public` in `appinfo/routes.php:17`) sets no `'direct'` key, so `Config.get('direct')` is
  falsy while the page is still top-level. On that path `document.js:403-405` re-posts the raw
  `'close'` payload to `parent` (itself) for the same reason, so the loop class is still reachable.
  Guarding on `window.parent !== window` (or refusing to post to self inside
  `PostMessageService.sendPostMessage`) fixes the bug rather than one instance of it.

- **`.should('be.visible', { timeout: 10_000 })` is a no-op.** `cypress/e2e/direct.spec.js:269`.
  `should()` has no options-object overload — the signatures are `(chainers)`,
  `(chainers, value)`, `(chainers, method, value)` and `(callbackFn)`. The object is passed as an
  assertion argument to a property assertion and silently ignored, so the retry budget is still
  the default. Timeouts belong on the command:
  `cy.get('button[aria-label="Save As"]', { timeout: 10000 }).should('be.visible').click()`.
  (Also worth matching the repo's existing `{ timeout: 30000 }` style — no numeric separators.)

- **The identical broken call is still present at `src/document.js:472.**
  `case 'Get_Views_Resp'` calls `documentsMain.openLocally()` — same non-existent mixin method,
  same latent `TypeError`. It is currently unreachable only because `documentsMain.openingLocally`
  (read at `:467`) is never assigned anywhere in `document.js`, i.e. it is dead code. Note
  `unlockFile` at `:469` *does* exist on `documentsMain` (`:539`), so the block reads as if it
  works. Since this PR is already removing one instance, please either remove this block too or
  add a `TODO`/issue link — leaving one of two identical bugs is the confusing outcome.

- **Scope: a production bug fix is buried in a CI/test PR.** The `src/document.js` change fixes a
  real user-facing hang (unbounded postMessage loop on close in direct editing) and deserves its
  own PR, its own title, and changelog/backport visibility — not `fix(ci): stabilize cypress
  tests`. Nextcloud's one-concern-per-PR guidance applies. Splitting it also lets the good part
  merge while the targetOrigin question is still open. Worth checking whether it needs a
  `stable32`/`stable31` backport.

### ℹ️ Minor / 💡 Suggestions

- ℹ️ **`tabbed` → `notebookbar` is now inconsistent.** `cypress/e2e/open.spec.js:61` still uses
  `'tabbed'` for the same app config key. `open.spec.js` passes in CI, so `tabbed` evidently
  produces a notebookbar UI too — which means either this change is cosmetic, or `open.spec.js`
  needs the same treatment. Pick one and make both specs agree.
- ℹ️ **Commit message typo:** `fix(document): stablize cypress tests` → "stabilize". Also consider
  a `fix(direct-editing)`-style scope, since the commit changes runtime behaviour rather than a test.
- ℹ️ **No linked issue.** If the direct-editing close hang was reported, linking it would help;
  otherwise the split-out PR is a good place to describe the loop.
- 💡 **The content-based Save-As selector is the right call** — index-based `#saveas-entry-1` was
  exactly what broke on main. It does couple to the English label `Rich Text (.rtf)`, but the test
  already depends on `aria-label="File"` / `"Save As"`, so that is consistent.
- 💡 Consider recording the loop in a regression test. `cy.closeDirectDocument()` in the Save-As
  test already exercises the guarded path implicitly, which is good; the `!isDirectEditing()`
  branch (where `'close'` *is* now posted to the parent for the first time, since the swallowed
  `TypeError` used to prevent it) has no coverage at all and is the riskier half of the change.

## Code quality & conventions

- **Commits:** Conventional Commits followed (`fix(document)`, `fix(test)`, `fix(ci)`); one typo
  ("stablize"). Three distinct concerns, cleanly separated per commit — good, and it makes the
  split-out suggested above easy.
- **DCO:** ✓ `Signed-off-by:` present on all three commits; the DCO check is green.
- **AI disclosure:** ✓ Exemplary — the description discloses Claude assistance and notes the
  description itself is the author's, and every commit carries
  `Assisted-by: ClaudeCode:claude-sonnet-4-6`.
- **Tests:** This is largely a test PR. The one production change has no direct assertion, though
  the existing `closeDirectDocument()` path covers the direct-editing side implicitly.
- **Licensing:** No new files, so no SPDX headers needed. REUSE check green.
- **CI:** Everything green except Cypress (`runner nightly-1`, `cypress-summary`) — failing on the
  test this PR set out to fix.

## Verdict

**Request changes** — the `src/document.js` fix is correct and well diagnosed (I verified the full
loop), and the Save-As selector demonstrably fixes one of main's two flaky failures. But the
targetOrigin test is still red at head for a reason the change does not touch, and the CI
capability grant provably does not do what the description claims while handing `SYS_ADMIN` to an
unpinned third-party image — that one should just be dropped. Splitting the production fix into its
own PR would let the solid parts land now.

Assisted-by: ClaudeCode:claude-opus-5

rikled (approved):

PR #5971, opened 2026-08-10
fix(templates): avoid extra file write

elzody:
Avoids an extra file write when no template fields are provided. Not only does this avoid an extra write to the file, but it consequently also prevents the files_antivirus app from doing a second, sequential scan on the file; this should reduce the latency between creating a file and seeing it.

### Checklist

### DIFF
--- a/lib/Listener/FileCreatedFromTemplateListener.php
+++ b/lib/Listener/FileCreatedFromTemplateListener.php
@@ -55,7 +55,7 @@ class FileCreatedFromTemplateListener implements IEventListener {
 			return;
 		}
 
-		if ($this->capabilitiesService->hasFormFilling()) {
+		if ($this->capabilitiesService->hasFormFilling() && !empty($event->getTemplateFields())) {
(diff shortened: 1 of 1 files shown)


rikled (approved):

moodyjmz (approved):
Verified the mechanism: this restores the pre-existing setTemplateSource fallback for the empty-fields case, matching the intent of the earlier commit that gated it behind hasFormFilling(). Chased down a possible PDF-flatten regression (PdfService::fillFields unconditionally flattens) — confirmed unreachable, since no PDF template creator is registered anywhere in richdocuments, so the template file can never be a PDF here.

Two minor non-blocking notes, not relevant to this PR's correctness:
- extractFields() gates on Capabilities::MIMETYPES, which excludes MIMETYPES_MSOFFICE — so template field-filling is already inert for default (ooxml) installs, independent of this change. Worth its own issue.
- No test coverage for this listener, before or after this change.

PR #5972, opened 2026-08-12
fix: hanging test in direct.spec.js

elzody:
Fixes hanging tests in `direct.spec.js` because the sessions were not being cleared properly, and fixes a psalm error regarding a change to OCP.

### Checklist

### DIFF
--- a/cypress/e2e/direct.spec.js
+++ b/cypress/e2e/direct.spec.js
@@ -288,7 +288,7 @@ describe('Direct editing (legacy)', function() {
 
 				cy.get('.saveas-dialog button.button-vue--vue-primary').click()
 
-				cy.waitForPostMessage('Action_Save_Resp', { success: true })
+				cy.waitForPostMessage('Action_Save_Resp', { success: true, fileName: 'document.rtf' })
 				cy.closeDirectDocument()
 			})
 	})
--- a/cypress/e2e/direct.spec.js
+++ b/cypress/e2e/direct.spec.js
@@ -288,9 +288,7 @@ describe('Direct editing (legacy)', function() {
 
 				cy.get('.saveas-dialog button.button-vue--vue-primary').click()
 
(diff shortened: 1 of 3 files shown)


moodyjmz:
**Save-As test regains its hang-fix but drops an assertion it doesn't need to drop.**

`cy.get('@loleafletframe').within(() => { cy.verifyOpen('document.rtf') })` was verifying that Collabora actually reopened the file *as* `document.rtf` after Save As. Swapping to `cy.waitForPostMessage('Action_Save_Resp', { success: true })` fixes the hang (the `@loleafletframe` alias is stale after the Save-As reload, so `.within()` was retrying against a detached node until timeout) — but now the test passes on *any* successful save, not specifically a save-as-rtf.

The fix is already established in this repo: `cypress/e2e/integration.spec.js:96` runs the same Save-As-to-rtf flow and asserts on the `fileName` field in the same postMessage payload:

```js
cy.waitForPostMessage('Action_Save_Resp', { success: true, fileName: exportFilename })
```

Suggest mirroring that here instead of dropping the check entirely:

```diff
-				cy.get('@loleafletframe').within(() => {
-					cy.verifyOpen('document.rtf')
-				})
+				cy.waitForPostMessage('Action_Save_Resp', { success: true, fileName: 'document.rtf' })
```

`checkExpectedValues` in `cypress/support/commands.js` does per-key equality against `message.Values`, so this is a drop-in — no new plumbing needed. Keeps the hang fixed, restores the coverage.

(Everything else in the PR — the migration `changeColumn` → `modifyColumn`, and the `logout` command switching from a stale `cy.session` cache to `clearAllSavedSessions()`/`clearCookies()` — looks correct as-is, no changes requested there.)

moodyjmz (approved):
Fix applied exactly as suggested (fileName assertion restored). All three changes verified against actual repo/runtime behavior — LGTM.

PR #5973, opened 2026-08-12
fix: pass language to Collabora file conversions

### LINKED ISSUE #5165: Language Parameter Not Provided To Collabora Conversion API
Please update me when this is fixed, as it annoyed users in my cloud env  - with me being unable to fix it without studying all the code and APIs. 

copy from issue:  https://github.com/CollaboraOnline/online/issues/12773
```
I think its a bug or smth like that.
Where does collabora pull the language and locale for nextcloud folder view -> right click -> save as pdf from? 
Seemingly, no language or locale setting works!
Your forum is so utterly broken, one is unable to even post images or links properly.


how to test?

1. open test.xlsx
3. go to insert & click on header/footer
5. put the datetime top right (right area); custom format
7. you see the format as your set language in nextcloud! keep that in mind
9. now close the document, click right click on it in nextcloud folder view and click on save as → export as pdf
11. Check the exported pdf’s date (click on it for preview). It shows the wrong date format

https://forum.collaboraonline.com/t/asap-language-settings-calc-documents-xslx-save-as-pdf-locale-setting/3990

Version is v24.04.10.1.1 collabora/code
```

answer by maintainers (?)
```
Collabora Online's [Conversion API](https://sdk.collaboraonline.com/docs/conversion_api.html) can take language parameter. But Nextcloud's richdocuments app does not send it.
This works as expected.

curl -k -F "data=@datum.xlsx" -F "lang=de-DE" http://localhost:9980/cool/convert-to/pdf > datum.pdf
So this is a richdocuments bug after all.
cc @elzody
``` 
by @timar  see https://github.com/CollaboraOnline/online/issues/12773#issuecomment-3291759719

xhon-pelushi:
## Summary
- Pass the current user locale to Collabora file conversions as a BCP 47 `lang` parameter.
- Send conversion options as separate multipart fields (not merged into the file part).
- Add focused PHPUnit coverage for language + multipart option payload.

Fixes #5165

## Test plan
- [ ] Save-as PDF (or other conversion) with a non-English locale and confirm locale-sensitive formatting uses the right language.
- [ ] `composer run test:unit` for the new conversion tests when PHP tooling is available.

### DIFF
--- a/lib/Conversion/ConversionProvider.php
+++ b/lib/Conversion/ConversionProvider.php
@@ -165,7 +165,18 @@ class ConversionProvider implements IConversionProvider {
 			}
 		}
 
-		return $this->remoteService->convertFileTo($file, $targetFileExtension);
+		return $this->remoteService->convertFileTo(
+			$file,
+			$targetFileExtension,
+			conversionOptions: ['lang' => $this->getConversionLanguage()]
+		);
+	}
+
+	private function getConversionLanguage(): string {
+		$locale = $this->l10n->getLocaleCode();
+		$language = $locale !== '' ? $locale : $this->l10n->getLanguageCode();
+
+		return str_replace('_', '-', $language);
 	}
 
 	private function getMimeProvidersFor(array $inputMimeTypes, string $outputMimeType): array {
--- a/lib/Service/RemoteService.php
+++ b/lib/Service/RemoteService.php
@@ -58,9 +58,15 @@ class RemoteService {
 	}
 
 	/**
+	 * @param array<string, string> $conversionOptions
 	 * @return resource|string
 	 */
-	public function convertFileTo(File $file, string $format, int $timeout = RemoteOptionsService::REMOTE_TIMEOUT_DEFAULT) {
+	public function convertFileTo(
+		File $file,
+		string $format,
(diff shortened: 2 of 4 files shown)


elzody:
Thank you for working on this @xhon-pelushi. That you want to work on the richdocuments app means a lot. <3

I do have some feedback that I think is worth considering:

In `src/helpers/index.js` there is a function called `languageToBCP47()` that translates the language setting given by Nextcloud into a BCP 47 tag that Collabora can correctly handle. Given that, to fix the problem this PR tries to solve, this logic would likely need to be duplicated in PHP since the `/convert-to` endpoint's `lang` field also takes the same BCP 47 language tags. I think it would make more sense to move the logic for that to the PHP side and then serve it to the front-end as an initial state, so the logic only lives in one place.

That said, I have not totally tested this PR yet, but given that it works as intended I am happy to merge it in the meantime and we can create a separate issue for refactoring that bit. But if you want to already take the opportunity to take care of that, that's also fine.

Will give it a proper test soon.

elzody (approved):
I got around to testing it and it seems to fix the problem. As said before, it would be best to merge the language setting to BCP 47 tag translation logic so that it lives on the back-end. I have created an issue for that at https://github.com/nextcloud/richdocuments/issues/6034 so that we can get that in as a follow-up.

PR #5974, opened 2026-08-13
Hide extra fonts instructions when built-in CODE selected

### LINKED ISSUE #3804: The extra fonts "Make sure to set this URL" instructions should not be shown when built-in CODE is selected
This setting is already active in the Built-in CODE app. At best it's confusing. At worst, it's misleading when one is troubleshooting font issues and may send operators down a rabbit hole that is unnecessary.

If someone is interested in implementing this, change will need to happen in here:

https://github.com/nextcloud/richdocuments/blob/f020d3904ccd3e00a534bbe9054df263c953dc53/src/components/AdminSettings.vue

xhon-pelushi:
Fixes #3804

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -295,10 +295,10 @@
 			<SettingsFontList :fonts="settings.fonts"
 				:label="t('richdocuments', 'Available fonts')"
 				@deleted="onFontDeleted" />
-			<em>
+			<em v-if="showFontConfigHint">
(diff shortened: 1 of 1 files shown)


elzody (approved):

elzody:
The change seems sound. It is possible some of the tests will fail, but to resolve that it is likely just a matter of rebasing these changes on top of the latest main branch; there were some fixes recently to stabilize the tests.

PR #6014, opened 2026-08-30
[stable32] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency


elzody (approved):
An update to `nextcloud/ocp` would not be the cause of the error, I think it's something nightly build related.

PR #6024, opened 2026-09-03
fix(templates): include ooxml types in form filling

### LINKED ISSUE #5991: Template field filling is silently inert for MS Office / OOXML templates
## TL;DR
`TemplateFieldService::extractFields()` only proceeds for mimetypes in `Capabilities::MIMETYPES` — which holds ODF/legacy formats, not the OOXML ones. Since richdocuments defaults to `doc_format=ooxml`, the field-extraction/fill-in-dialog feature never fires for the default document/spreadsheet/presentation templates. No error, no log — it just silently returns no fields every time.

<details>
<summary>Details</summary>

**Where:** `lib/Service/TemplateFieldService.php`, `extractFields()`:

```php
if (!in_array($file->getMimetype(), Capabilities::MIMETYPES)) {
    return [];
}
```

**The gap:** `Capabilities::MIMETYPES` (`lib/Capabilities.php:17-38`) contains only ODF and a handful of legacy/alternate formats (`.odt`, `.ods`, `.odp`, `.odg`, RTF, etc.). All OOXML/MS mimetypes (`.docx`, `.xlsx`, `.pptx`, `.doc`, `.xls`, `.ppt`, and their macro-enabled variants) live in the separate `Capabilities::MIMETYPES_MSOFFICE` constant (`lib/Capabilities.php:40-62`), which `extractFields()` never checks.

**Why it matters:** `RegisterTemplateFileCreatorListener.php` defaults to `doc_format=ooxml` (`getAppValue(Application::APPNAME, 'doc_format', 'ooxml')`), meaning the document/spreadsheet/presentation template types register with `.docx`/`.xlsx`/`.pptx` mimetypes out of the box. For any install on that default:

1. User picks a custom template with fields defined.
2. Frontend calls `listTemplateFields` → `TemplateManager::getTemplateFields()` → `TemplateFieldService::extractFields()`.
3. `extractFields()` hits the mimetype guard, returns `[]`, unconditionally, regardless of whether the template actually has fields.
4. Frontend's `fields.length > 0` check (`apps/files/src/views/TemplatePicker.vue` in nextcloud/server) is always false, so the "Fill template fields" dialog never appears.

The end result: the template form-filling feature (fields extraction + fill-in dialog) is effectively dead for any install using the default OOXML document format — the only way it currently works is if an admin has switched `doc_format` to `odf`.

**Suggested fix:** have `extractFields()` check `array_merge(Capabilities::MIMETYPES, Capabilities::MIMETYPES_MSOFFICE)`, matching the guard already used in `FileCreatedFromTemplateListener::handle()`.

**Found while reviewing** #5971, which is unrelated and does not need to block on this — this is a pre-existing, separate bug.

</details>

elzody:
* Resolves: https://github.com/nextcloud/richdocuments/issues/5991
* Target version: main

### Summary
MS office MIME types should be considered in addition to OpenDocument
types when creating a document from a template. This allows support for
form filling on the MS office document types

Assisted-by: ClaudeCode:claude-sonnet-5

### Checklist

### DIFF
--- a/lib/Service/TemplateFieldService.php
+++ b/lib/Service/TemplateFieldService.php
@@ -66,7 +66,7 @@ class TemplateFieldService {
 				return $fields;
 			}
 
-			if (!in_array($file->getMimetype(), Capabilities::MIMETYPES)) {
+			if (!in_array($file->getMimetype(), array_merge(Capabilities::MIMETYPES, Capabilities::MIMETYPES_MSOFFICE))) {
 				return [];
 			}
 
--- a/tests/lib/Service/TemplateFieldServiceTest.php
+++ b/tests/lib/Service/TemplateFieldServiceTest.php
@@ -1,0 +1,103 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace Tests\Richdocuments;
+
+use OCA\Richdocuments\AppConfig;
+use OCA\Richdocuments\Service\CapabilitiesService;
+use OCA\Richdocuments\Service\PdfService;
+use OCA\Richdocuments\Service\RemoteService;
+use OCA\Richdocuments\Service\TemplateFieldService;
+use OCP\Files\File;
+use OCP\Files\IRootFolder;
+use OCP\Files\Storage\IStorage;
+use OCP\ICache;
+use OCP\ICacheFactory;
+use OCP\ITempManager;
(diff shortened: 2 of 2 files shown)


moodyjmz (approved):
LGTM — this matches the `array_merge(MIMETYPES, MIMETYPES_MSOFFICE)` guard already used in `FileCreatedFromTemplateListener.php:39`, and the new test actually exercises the OOXML case that was broken.

Minor, non-blocking: this union is now duplicated in two places, and `AppConfig::getMimeTypes()` already computes almost the same thing — could be a shared helper at some point.

elzody:
The failures are not related to this. The tests pass when run against stable Collabora, but nightly fails due to an unrelated error. I've already raised it with Collabora.

PR #6055, opened 2026-09-08
feat(wopi): report the file's SHA-256 in CheckFileInfo

timar:
After an upload that gets no answer, the editor has to work out whether what is in storage is what it just sent. Without a hash to compare it falls back on the byte count, which cannot tell our own document from another editor's of the same length: the wrong answer either loses the other writer's work or puts a needless conflict in front of the user.

Report the file's SHA-256 in the CheckFileInfo SHA256 field so there is something better to compare. The field is only present when a hash is on record; CheckFileInfo is on the document-open path and must not read a whole file to compute one.

For it to be on record when it matters, hash the body of PutFile as it streams past and store the result, the way the DAV endpoint stores the checksums a sync client sends. Wrapping the stream costs no extra read, since the bytes pass through us in any case. The scanner clears the checksum whenever a file changes, so what we hand out is never stale: it either describes the current contents or is absent, and setting it after putContent() keeps it that way.

Files written by other means, or before this change, simply carry no SHA-256 and the editor falls back as it does today.

### Checklist

### DIFF
--- a/lib/Controller/WopiController.php
+++ b/lib/Controller/WopiController.php
@@ -7,6 +7,7 @@
 
 namespace OCA\Richdocuments\Controller;
 
+use OC\Files\Stream\HashWrapper;
 use OCA\Files_Versions\Versions\IVersionManager;
 use OCA\Richdocuments\AppConfig;
 use OCA\Richdocuments\AppInfo\Application;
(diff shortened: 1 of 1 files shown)


timar:
ping @elzody @juliusknorr 
Are you considering to take this feature? The COOL side has been merged already.

PR #6069, opened 2026-09-11
fix(viewer): use the header height token for the document offset

Darshan-upadhyay1110:
- Nextcloud 35 reduces --header-height from 50px to 44px.
- The fixed -50px offset now cuts off the top of the Collabora Online toolbar.
- Use calc(var(--header-height) * -1) so the offset follows the theme.

Before:
image

After:

image


elzody:
Closing this as we use the backport bot: https://github.com/nextcloud/richdocuments/pull/6082

PR #2301, opened 2022-07-06
Exclude MSOffice mime types association if OnlyOffice app is installed

proxyconcept:
Signed-off-by: Grégory Marigot <gmarigot@teicee.com>

* Resolves: #2281 
* Target version: master 

### Summary

The Nextcloud Office app configure the association with all possible office documents mime types.
But if we also have the OnlyOffice app, it's not possible to open any document with it...

Collabora and OnlyOffice are both great and it's a common case to want to install both: Collabora for OpenOffice documents and OnlyOffice for MSOffice documents.

This patch is a workaround specific for this case (and because it's not possible to wait for nextcloud/server#8105) : the Nextcloud Office app check if the OnlyOffice app is not installed before adding all the MSOffice mime types.

### Checklist

- [x] Code is properly formatted
- [x] Sign-off message is added to all commits
- [ ] Documentation (manuals or wiki) has been updated or is not required

Issue #928, opened 2020-04-26
option not to open OOXML and MS-Office files (docx, ..., doc, ...)


kolAflash:
Please provide an option to disable opening OOXML and MS-Office files (docx, ..., doc, ...) by default.
This is needed, if another Nextcloud app like OnlyOffice is installed to open those files.

Currently, when installing LibreOfficeOnline and OnlyOffice, the behaviour is very confusing.
E.g. docx files open with OnlyOffice when logged into Nextcloud. But when an anonymous user clicks a shared link to a docx file, it's being opened with LibreOfficeOnline (this is not for link-shared folders, but only single link-shared files).

juliushaertl:
cc @jancborchardt 

I remember we had some discussion about running both office suites together but I do not remember if there was any outcome on this.

jancborchardt:
It seems @kolAflash's core issue is that the behavior is different for logged-in people vs guests. Of course this should be consistent.

Since with 19 our default is Collabora Online, that should also stay the default if both are installed.

PR #6087, opened 2026-09-16
feat(settings): option to keep opening MS Office files when another office app is enabled

### LINKED ISSUE #3456: richdocuments effectively disable Collabora if onlyoffice plugin is installed

acbelle59:
### Summary

Since #2301, when the `onlyoffice` (or `officeonline`) app is enabled, richdocuments stops handling Microsoft Office formats (docx, xlsx, pptx…) and only keeps OpenDocument formats, so that each app handles its own family of formats. That is a sensible default, but some admins want the opposite: keep opening every format with Collabora by default, and let users pick the other app from the file actions menu ("Open in ONLYOFFICE"), per document — see #3456 (and the earlier #928).

This adds an **opt-in** admin setting, `msoffice_with_other_office_apps` (default off, so nothing changes for existing installations). When enabled, `hasOtherOOXMLApps()` returns `false` and the MS Office mimetypes stay in the default list. It is exposed exactly like `disable_certificate_verification`: settings controller, admin initial state, capabilities, a checkbox in the admin settings (Advanced settings, under "Use OOXML by default"), and a paragraph in `docs/app_settings.md`.

### How it was tested

Nextcloud 34.0.3, richdocuments 11.1.1, onlyoffice 10.1.2 (with the OnlyOffice connector's default formats released so that it does not claim the simple click):

- **option enabled**: a simple click opens Collabora; the file actions menu offers "Open in ONLYOFFICE"; the same document was edited with both editors (and from an iPhone with Collabora), all changes saved;
- **option disabled**: current behaviour unchanged — ONLYOFFICE takes the click, Collabora is not offered.

Resolves #3456.


elzody:
At first glance, perhaps instead of introducing a new configuration option, we could just default to adding an option in the files action menu to "Open with Nextcloud Office (Collabora)" when another office app that handles OOXML types is enabled. That way it would just be there for those who want to use it.

